PR #1, opened 2018-11-30
Use offsets rather than counts, np.clip

nsmith-:
Not tested, need to setup some tests


lgray:
Lemme know when you've tested it and I'll merge.

We can work on tests with CI once we've got everything more or less in place.

lgray:
Going to merge this now so I can build up the continuous integration.

PR #6, opened 2018-12-02
tools.analysis_objects reorganization and better tests

lgray:
```JaggedDecoratedFourVector``` and ```JaggedWithLorentz``` are no longer provided in this package.

There is now ```JaggedCandidateArray``` and ```JaggedTLorentzVectorArray``` which are at first more descriptive names. ```JaggedTLorentzVectorArray``` is exactly the same as ```JaggedWithLorentz```.

```JaggedCandidateArray``` supplies the same functionality as the original ```JaggedDecoratedFourVector``` but is now created through a mixin to JaggedArray rather than a daughter class. This forced a few changes which should make the code more robust to misuse and easier to debug.

@nsmith-


lgray:
@jpivarski - if you have time I would appreciate comments on what I have done wrong with the ```JaggedCandidateArray```

jpivarski:
@lgray By "wrong," you mean as an open-ended request for comments or because something specific isn't satisfactory? It looks pretty good to me. You probably want to reduce your dependence on `fromcounts`, since that forces a commutative sum to make offsets, and the offsets of arrays with the same structure won't be the same arrays in memory. If you use `fromoffsets` (after a single conversion from counts to offsets), the jagged structure of all compatible arrays will be represented by a single set of offsets, which means less sanity checking to see if they are compatible.

Also, I don't understand why some of these methods need to be overloaded, like `cross` and `pairs`. I'll stare at it more, though.

I think I'll have some time this afternoon to look at all the performance issues you've been sending my way.

lgray:
@jpivarski Yes, more or less an open ended request for comments, but particularly pertaining to clarity of analysis code written with these classes and good style/structure (especially with respect to interfaces with jagged as you've talked about so far).
Thanks for explaining the underlying difference for fromoffsets vs. fromcounts, I'll make this change soon. 

Thanks a lot for looking at this performance bottleneck we spotted!

PR #12, opened 2018-12-09
Implement JECs, also refactor evaluators to handle more cases

lgray:
Work in progress.

- New extractor function for JECs
   - Evaluator is on the way

- Factored lookup_tools.evaluator into the evaluator itself and many 'lookup' subclasses
   - Adapted extractor and extractor functions to handle passing around the new extra info

- Some small cleaning here and there (like only reading first lines of files, and such)


lgray:
@nsmith- add here if you want to

Issue #23, opened 2018-12-15
Implement JEC systematics


lgray:

lgray:
This was partially addressed in #24 but some remaining things are needed, specifically working out the analysis level specifics.

lgray:
Finished addressing in #30

Issue #17, opened 2018-12-13
Make JER evaluator / smearer to complement JEC


lgray:
Name says it all.

lgray:
This was partially addressed in #24 but some remaining things are needed.
This depends on #20 so that we can implement the deterministic smearing for well-matched jets.
The final transform from looked-up SF values to updated jet energies is also needed.

lgray:
Addressed in #30

PR #24, opened 2018-12-19
Additions from implementing the rest of DAZSLE Hbb analysis

lgray:
* jet energy resolutions (eta/pt)
* jet energy scale uncertainties
* jet energy resolution scale factors

Todo -> make factorized jet corrector, make jet smearer

To think about -> how to return systematics variations??? array[-1],[0],[1] is sort of intuitive maybe?

root converter routine also now skips a few more TGraph types... We should eventually figure out how to use them.


lgray:
Addresses #17 and #23

nsmith-:
Ah I'll need to add matplotlib to the CI tools as well.
Any reason not to merge? I can then pull into my branch

lgray:
Go ahead and squash+merge.

PR #25, opened 2018-12-20
Update root_converters.py

fnavarro94:


lgray:
@fnavarro94 This is already handled in #24

PR #26, opened 2018-12-20
Histograms and plotting

nsmith-:
This should in principle obviate histbook.  Consider the `Hist` definition and filling implementation somewhat stable.


nsmith-:
So I got it to pass in python 3.6 but not 3.5?? Clearly the usage of dictionaries needs to be removed.

tests/test_hist_tools.py
```diff
@@ -0,0 +1,84 @@
...
+    baby_bison_h = np.random.normal(loc=.5, scale=0.1, size=20)
+    baby_bison_w = np.random.normal(loc=200, scale=10, size=20)
+    baby_bison_cutefactor = 2.5*np.ones_like(baby_bison_w)
+    h_mascots_2.fill(animal="bison", vocalization="baa", height=baby_bison_h, mass=baby_bison_w, weight=baby_bison_cutefactor)
+    h_mascots_2.fill(animal="fox", vocalization="none", height=1., mass=30.)
+
+    h_mascots = h_mascots_1 + h_mascots_2
+    assert h_mascots.project("vocalization", values="h*").project("height").project("mass").project("animal").values()[()] == 1040.
+
+    subphylums = {
```
lgray: The plural of Phylum is Phyla

lgray:
@nsmith- when you've got your gallery notebook setup lemme know. I want to poke at it.

nsmith-:
So currently slices on dense axes can only be integrated out, any other operation will de synchronize the axis definition with the ndarray.  I'll fix it, but for now we have at least a working demo of most of the things we need.

nsmith-:
Ok I'm happy with the functionality here.  I think the core class is stable enough to be put to work.
If no complaints, I'll merge

lgray:
Nope no complaints. I played around with the notebooks, this is a very good direction.
I'll definitely be switching once I get the ggHbb more or less implemented.
Currently working on the muon control region, jet systematics after that.

PR #29, opened 2018-12-20
Match objects within dR or dPt

areinsvo:
First attempt at matching particles. See https://github.com/CoffeaTeam/fnal-column-analysis-tools/issues/18


lgray:
Here's profiler output from 1.3M events from a boosted W sample. Matching gen muons.

```
4.785 <module>  <ipython-input-12-ee2f2ae909d9>:38
├─ 3.180 argmatch  build/bdist.macosx-10.6-x86_64/egg/fnal_column_analysis_tools/analysis_objects/JaggedCandidateArray.py:245
│  ├─ 1.106 argmin  awkward/array/jagged.py:1220
│  │  └─ 1.102 argmin  awkward/array/jagged.py:1220
│  │     └─ 1.101 _argminmax  awkward/array/jagged.py:1234
│  │        └─ 1.101 _argminmax_general  awkward/array/jagged.py:1270
│  │           └─ 0.349 argmin  numpy/core/fromnumeric.py:1040
│  │              └─ 0.206 _wrapfunc  numpy/core/fromnumeric.py:49
│  ├─ 1.019 delta_r  uproot_methods/classes/TLorentzVector.py:259
│  │  └─ 0.977 delta_r2  uproot_methods/classes/TLorentzVector.py:109
│  │     ├─ 0.403 func  numpy/lib/mixins.py:22
│  │     │  └─ 0.400 __array_ufunc__  awkward/array/jagged.py:844
│  │     │     ├─ 0.159 __array_ufunc__  awkward/array/jagged.py:844
│  │     │     │  └─ 0.052 _tojagged  awkward/array/jagged.py:784
│  │     │     ├─ 0.091 flatten  awkward/array/jagged.py:1095
│  │     │     │  └─ 0.060 __getitem__  awkward/array/jagged.py:501
│  │     │     ├─ 0.076 _tojagged  awkward/array/jagged.py:784
│  │     │     └─ 0.055 fromcounts  awkward/array/jagged.py:180
│  │     ├─ 0.383 delta_phi  uproot_methods/classes/TLorentzVector.py:106
│  │     │  ├─ 0.328 func  numpy/lib/mixins.py:22
│  │     │  │  └─ 0.321 __array_ufunc__  awkward/array/jagged.py:844
│  │     │  │     ├─ 0.124 __array_ufunc__  awkward/array/jagged.py:844
│  │     │  │     │  └─ 0.049 _tojagged  awkward/array/jagged.py:784
│  │     │  │     ├─ 0.073 flatten  awkward/array/jagged.py:1095
│  │     │  │     ├─ 0.053 _tojagged  awkward/array/jagged.py:784
│  │     │  │     └─ 0.052 fromcounts  awkward/array/jagged.py:180
│  │     │  └─ 0.051 memofunction  uproot_methods/base.py:84
│  │     └─ 0.057 memofunction  uproot_methods/base.py:84
│  ├─ 0.562 __getitem__  awkward/array/jagged.py:501
│  │  ├─ 0.501 __getitem__  awkward/array/jagged.py:501
│  │  │  ├─ 0.302 _tojagged  awkward/array/jagged.py:784
│  │  │  │  └─ 0.288 offsets2parents  awkward/array/jagged.py:54
│  │  │  └─ 0.133 parents  awkward/array/jagged.py:426
│  │  │     └─ 0.121 offsets2parents  awkward/array/jagged.py:54
│  │  └─ 0.061 isstringslice  awkward/util.py:178
│  └─ 0.411 cross  awkward/array/jagged.py:1061
│     ├─ 0.196 _argcross  awkward/array/jagged.py:1026
│     │  └─ 0.148 offsets2parents  awkward/array/jagged.py:54
│     └─ 0.183 _broadcast  awkward/array/jagged.py:772
│        └─ 0.181 parents  awkward/array/jagged.py:426
│           └─ 0.181 offsets2parents  awkward/array/jagged.py:54
├─ 0.856 __getitem__  awkward/array/jagged.py:501
│  ├─ 0.405 parents  awkward/array/jagged.py:426
│  │  └─ 0.404 offsets2parents  awkward/array/jagged.py:54
│  │     └─ 0.085 cumsum  numpy/core/fromnumeric.py:2092
│  │        └─ 0.085 _wrapfunc  numpy/core/fromnumeric.py:49
│  └─ 0.186 __getitem__  awkward/array/table.py:525
│     └─ 0.186 _newslice  awkward/array/table.py:410
├─ 0.218 func  numpy/lib/mixins.py:22
│  └─ 0.218 __array_ufunc__  awkward/array/jagged.py:844
└─ 0.071 __array_ufunc__  awkward/array/jagged.py:844
```
so argmatch is running at about 2.5 us per event. This is by far the heaviest operation needed, even compared to histogram lookups and jet energy corrections.

I was running the profiler only around the extraction of the good status==1 muons and then the matching of the gen particles to the reco muons. JaggedArray argmin and associated functions taking quite some time, as well as deltar2. Suspect there may be some room for improvement.

@jpivarski Looking at _argminmax_general() there's a bunch of time spent in there. I see some allocations inside of it and such (numpy.arange, numpy.empty). I realize at some point you need them to be there, but perhaps there a way to optimize? deltar2 and other places look pretty reasonable, so don't think there is much to gain there aside from using our own pre-cached values of eta/phi.

lgray:
Yes, indeed, JaggedArray._argminmax_general is the biggest time user here. I tried rewriting the matching using the pre-computed eta and phi in the jagged candidate array and it's by far the biggest time user even with two cross()es instead of one!. It may be worth a look to optimize it.

lgray:
offsets2parents also looks like it's taking a fair amount of time, but the only thing requiring a large amount of time there is memory allocations.

jpivarski:
The `minmax_general` is a killer because it's a Python for loop for a case that can't be Numpyified. It just couldn't be forced into `numpy.maximize.reduceat`. It could be handled trivially with a compiled reimplementation of key functions, something that I'm planning for this upcoming summer's GSoC. (Remember how I said that this would be for performance in certain cases, but may even be needed for generalization? That's the project I'm talking about.)

If you absolutely can't get around this, then perhaps we need to start that project early, at least to make a replacement for `minmax_general`.

lgray:
@jpivarski This is something we can simply beat with more cores for now. I don't think it is any variety of show stopper (unless we find we are doing a lot of gen matching), but it seemed reasonable to point it out to you in case there was a work around now or planned.

lgray:
@areinsvo Aside from the lower level speed issues, which will get fixed elsewhere, this looks quite nice! Thanks!

I have one request:
- please add to argmatch() the option to match with a delta-pT threshold as well. This will already cover most cases, and we can add more cases as we encounter them.

areinsvo:
In the process of adding one more test, I realized that the output of argmatch can't be used directly as fancy indexing like I expected because the output is still doubly jagged. Using flatten() doesn't seem to work on the innermost dimension like we expect it too. @jpivarski is this a feature or a bug? If it's a feature, is there a way to flatten it the way we need?

Using the dummy arrays in the tests and matching particles to themselves:
jca_gen.argmatch(jca_gen,0.3,0.1) = [[[0] [1] [2]] [] [[0] [1]]]
jca_gen.argmatch(jca_gen,0.3,0.1).flatten() = [[0] [1] [2] [0] [1]] instead of [[0 1 2] [] [0 1]]  like we need

jpivarski:
Reducers work on the innermost dimension, but flatten is not a reducer. It has an `axis` parameter, which is by default `0` (outermost dimension). You probably want `flatten(axis=1)`.

(Jagged reducers can't take an `axis` parameter as they can in Numpy because subarray sizes wouldn't line up as they'd need to. However, flattening doesn't suffer from this problem, so we can use the Numpy idiom. This is essential complexity, not accidental.)

lgray:
@jpivarski Since it's the innermost ("most" jagged??) dimension of the jagged array shouldn't it be, in general, axis=-1?

jpivarski:
I wasn't speaking in general— I thought your case was two deep.

The `flatten` axis parameter doesn't have the negative shortcut because there isn't a well-defined "total depth." The structure could contain Tables that have different levels of jaggedness. The axis parameter can't pass through Tables, so it wouldn't apply anyway, but counting from the last applicable level would lead to gotchas— unexpected behavior.

lgray:
@jpivarski OK - fun. Yes, in this case (and most reasonable cases) the depth will only ever be two. Would be nice to devise some way to indicate the last level of jaggedness in an array somehow, not really needed though. Suspect it may become obvious in a C++ implementation.

lgray:
@areinsvo I'm taking some slight artistic liberties with your code at the moment, getting the output to format correct, etc. I also added one thing, if there is no gen match in a argmatch() call I return -1.

This way you can do gen-matching fancy indexing without asking for the masking array too (since you can generate it on the fly).

Ok with you?

PR #33, opened 2019-01-03
Fix density plot for irregular bins

nsmith-:
Not sure what I was trying to do the first time.  Can someone cross-check?


lgray:
What exactly are you trying to verify here? That a density plot with irregular bins has area one given the normalization you've put in here? The old one was indeed a bit wonky.

nsmith-:
Right, although I think we might also want to have an option for normalization = integral (i.e. bin-width-normalized)

PR #30, opened 2018-12-27
More artifacts that are needed to make the gghbb analysis work and other cosmetic things

lgray:
- tanh -> sinh for eta calculation
- effective area lookup from text file
- histograms you keep are now defined in finalize
- faster construction from pt/eta/phi/mass if available
- deal with TMath:: functions in JEC formula definitions
- jec, jer, jer-sf, jec-uncertainties operating OK
- factorized jet corrector and JME style tools
- JetTransformer class to apply corrections easily

To Investigate Further:
- Improve evaluation speed of jetmet corrections

PR #34, opened 2019-01-04
setup.py needs to match test envirnment

nsmith-:
..at least.
`scipy` and `matplotlib` are also dependencies, I think we should add them


lgray:
Go ahead and add scipy and matplotlib. We start to use more of scipy in #30 and in hbb.

PR #36, opened 2019-01-05
Add a stub to make histograms compatible with striped client

nsmith-:
A server side change is also necessary.
For people using the docker micro server container, here's a patch:
https://gist.github.com/nsmith-/db35bca2994ba39f6286b56ff616d858


nsmith-:
What the heck, I didn't even touch that file and can't reproduce build failure locally.

lgray:
Nope

lgray:
This problem is caused by the issue reported here in uproot: 
https://github.com/scikit-hep/uproot-methods/issues/31

@jpivarski

nsmith-:
Will tackle the uproot-methods change separately.

PR #38, opened 2019-01-07
Update to new return format for hist.numpy()

lgray:
as in the title.

It actually repacks the new format into the old format.


lgray:
Waiting for a new release of uproot methods, and then this test will pass.

nsmith-:
Why did you merge it then? uproot is currently pinned to the working release

lgray:
Ah, didn't notice. I'll revert.

PR #41, opened 2019-01-09
Allow unicode to be used for histogram stuff in python2

nsmith-:
This lets you open histograms pickled in python3 with python2.


nsmith-:
@lgray can we do a `0.1.1` for this?  I want to send around a demo thing

lgray:
Yeah, fine by me. There are other gghbb follow up issues coming too.

PR #45, opened 2019-01-20
Uproot adaptor and other goodies

nsmith-:
The `UprootJob` class provides in-notebook job scheduling for uproot-based jobs with an identical interface to striped.
The `HistCollectorCallback` provides a callback utility to use with striped and the standalone histogramming tools.

Sneak peak:
screen shot 2019-01-19 at 9 27 33 pm


fnal_column_analysis_tools/lookup_tools/lumi_tools.py
```diff
@@ -0,0 +1,78 @@
...
+class LumiData(object):
+    """
+        Class to hold and parse per-lumiSection integrated lumi values
+        as returned by brilcalc, e.g. with a command such as:
+        $ brilcalc lumi -c /cvmfs/cms.cern.ch/SITECONF/local/JobConfig/site-local-config.xml \
+                -b "STABLE BEAMS" --normtag=/cvmfs/cms-bril.cern.ch/cms-lumi-pog/Normtags/normtag_PHYSICS.json \
+                -u /pb --byls --output-style csv -i Cert_294927-306462_13TeV_PromptReco_Collisions17_JSON.txt > lumi2017.csv
+    """
+    def __init__(self, lumi_csv):
+        self._lumidata = np.loadtxt("lumi2017.csv", delimiter=',', usecols=(0,1,6,7), converters={
```
lgray: use lumi_csv here instead of hardcoded string...

fnal_column_analysis_tools/lookup_tools/__init__.py
```diff
@@ -1,2 +1,3 @@
-from fnal_column_analysis_tools.lookup_tools.extractor import extractor as extractor
-from fnal_column_analysis_tools.lookup_tools.evaluator import evaluator as evaluator
+from .extractor import extractor as extractor
+from .evaluator import evaluator as evaluator
+from .lumi_tools import LumiData, LumiMask, LumiList
```
lgray: Can you move lumi_tools one level up to make it equal with jetmet_tools? Lookup tools ends up being more or less experiment independent, so I'd like to keep it clean so we can extract it into its own package later.

Also, none of the tools you add are evaluated through an evaluator or extracted using a converter.

nsmith-: Sounds good, done

PR #46, opened 2019-01-23
Fix bug in plot1d when overlay is None

nsmith-:
oops guess I stopped testing that one


lgray:
Add a test :-)

nsmith-:
how to test plots though

lgray:
pickle.dumps should be constant over time? Use that to make comparison of plot objects?

nsmith-:
https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test
Later though

lgray:
ah, cute

PR #50, opened 2019-02-01
Update fnal-column-analysis-tools to opportunistically use awkward.numba

lgray:
@nsmith- any better way to do this?


fnal_column_analysis_tools/util.py
```diff
@@ -0,0 +1,6 @@
+try:
+    import awkward.numba as awkward
+except :
```
nsmith-, line 3: whitespace

nsmith-:
This is how I imagined to do it

PR #64, opened 2019-04-03
Adding Saiyan Stuff

mcremone:


fnal_column_analysis_tools/arrays/Builder.py
```diff
@@ -0,0 +1,130 @@
+import awkward
```
nsmith-: `from ..util import awkward`
This allows a global switch for the different awkward versions (`awkward.numba`, `awkward.cuda`, etc.)

fnal_column_analysis_tools/arrays/Builder.py
```diff
@@ -0,0 +1,130 @@
...
+import uproot, uproot_methods
```
nsmith-: `uproot` isn't used, right?

tests/test_arrays.py
```diff
@@ -0,0 +1,124 @@
+import uproot, uproot_methods
+import numpy as np
+import fnal_column_analysis_tools
+from fnal_column_analysis_tools.arrays import Builder
+
+file = uproot.open("nano_5.root")
```
nsmith-, line 6: Add this file to `tests/samples` if it isn't too big.  Also, I think the whole test should be wrapped in a descriptive function name.

nsmith- (changes requested):

lgray:
@mcremone Please implement the changes Nick has requested.

tests/test_arrays.py
nsmith-: Gotta change the path here too.  Also, 40mb is a bit big, possible to make it a bit smaller, say <1 mb?

mcremone: Path is consistent no? I added the root file in tests.

nsmith-:
Superseded by #67

PR #65, opened 2019-04-04
added numba versions of lumi tools

jpata:
Speed up the luminosity tool operations using numba. Relies on the numba [dictionary](https://numba.pydata.org/numba-doc/dev/reference/pysupported.html#dict) type for hash maps, which landed in numba 0.43

## LumiData
```python
runs = arrs[b"run"]
lumis = arrs[b"luminosityBlock"]

runslumis = np.zeros((len(runs), 2), dtype=np.uint32)
runslumis[:, 0] = runs[:]
runslumis[:, 1] = lumis[:]
```

```python
ld = LumiDataOrig("../data/lumi2017.csv")
%timeit ld.get_lumi(runslumis)
```
1.56 s ± 6.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
ld2 = LumiData("../data/lumi2017.csv")
%timeit ld2.get_lumi(runslumis)
```
10.3 ms ± 48 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

```python
ld.get_lumi(runslumis), ld2.get_lumi(runslumis)
```
(116.541978639, 116.54197863899992)

## LumiMask

```python
m = LumiMaskOrig("../data/Cert_294927-306462_13TeV_EOY2017ReReco_Collisions17_JSON.txt")
m2 = LumiMask("../data/Cert_294927-306462_13TeV_EOY2017ReReco_Collisions17_JSON.txt")
```

```python
mask = m(runs, lumis)
mask2 = m2(runs, lumis)
np.all(mask==mask2)

```
True

```python
%timeit m(runs, lumis)
```
940 ms ± 3.17 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
%timeit m2(runs, lumis)
```
212 ms ± 3.52 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


lgray:
@jpata This looks nice, and keeps the rest of the array-like syntax we're used to at least on the user facing side. So long as that's preserved I'm happy, we can do anything behind the scenes so long as the data received are still columnar and that columnar structure is exposed to the compiler (as it is when doing SoA / offsets operations).

nsmith-:
Thanks for the tests!  Too bad windows numba is handicapped.  One pain point we've seen is that recent numba (`>0.37`) relies on LLVM 7, which can be challenging in some deployments.

jpata:
Had to introduce a few more things for windows (explicit casting, disable parallel loops), now tests pass. Let me know if there is anything else you would like to see!

PR #66, opened 2019-04-11
Add support for split JEC uncertainties sources

mverzett:
There are no unit tests added, please let me know if/where to put them.


nsmith-:
Will extending https://github.com/CoffeaTeam/fnal-column-analysis-tools/blob/master/tests/test_lookup_tools.py#L81 work for this case?  If not, feel free to add more tests.

lgray:
Was about to ask the same thing. Please add a test to make sure all cases work.

mverzett:
OK, done. Thanks for pointing me the location

nsmith-:
So you'll probably want to pull master into your branch to deal with #69 but the python 2.7 failure might be legitimate.  @lgray its something in the numpy text converter, maybe you are familiar

lgray:
@mverzett I'll take a look today.

mverzett:
@lgray for the record: I tried running locally in a python 2.7 env and it did not crash. I create some temporary files to split the sources, could they be the cause of the crash?

nsmith-:
@mverzett can you try pulling master in, and pushing to the branch?  Maybe the test fail is temporary (other than the unrelated numba thing)

lgray:
@mverzett Hey, so it fails for me as well in python3 on my mac.

Investigating.

lgray:
@mverzett Indeed, this isn't something that's compatible across systems.

I would suggest you instead chop up the files into strings in memory and then parse those using the StringIO file-like wrapper for strings.

lgray:
@mverzett Any movement here?

lgray:
please rebase this to master!

mverzett:
Hi, sorry for the long silence, got stuck in other projects/workshops. I’ll try to complete it by the end of the week.

lgray:
@mverzett No worries, just keeping the PR warm.

lgray:
@mverzett the pytest stuff passed but you've got a bunch of flake8 compliance issues.
Just do `python setup.py flake8` to get your own copy of the list.

lgray:
@nsmith- Can you have a quick look at this one, any comments?

nsmith-:
I don't have any issue

PR #68, opened 2019-04-12
Accumulator tests and updates

nsmith-:
This is a handy way to keep track of counts, weights, etc. that would be annoying to make histograms for.


nsmith-:
Appveyor build fail was a fluke it seems.

PR #74, opened 2019-04-25
Fork pyxrootd

nsmith-:
This resolves a bug in the `processor.run_uproot_job` tool which causes things to completely hang if using `futures_executor` with xrootd files.


nsmith-:
I found a better solution, going into uproot https://github.com/scikit-hep/uproot/pull/270

PR #78, opened 2019-05-14
Parsl executor and corresponding tests with simple config

lgray:
This implements the basic parsl workflow.
Executor test makes sure that we get out expected counts of muons and dimuons is test root data.
Could do more?

This PR also includes some cleanup for spark and some udf templates.


fnal_column_analysis_tools/processor/dataframe.py
```diff
@@ -13,7 +13,7 @@ class LazyDataFrame(MutableMapping):
     """
     def __init__(self, tree, stride=None, index=None, preload_items=None):
         self._tree = tree
-        self._branchargs = {'awkwardlib': awkward}
+        self._branchargs = {'awkwardlib': awkward, 'flatten': True}
```
lgray: @nsmith- FYI this is needed to make it so that the same analysis code works everywhere. I have to manually flatten jagged arrays in the spark version, and it's less confusing when you don't have to do .content on everything going into a JaggedCandidateArray.

Also makes it seemless with striped (in the future).

nsmith-: Hm ok, you'll have to update `binder/muonspectrum_v3.ipynb` as well, then.

lgray: Shall do, we can revisit this choice when we free ourselves of pandas.

lgray:
@nsmith- Can you give this a quick look?

fnal_column_analysis_tools/processor/dataframe.py
nsmith-: This does completely disable usage of array branches where no `nObject` branch exists.  Could it be a configurable thing?  Recall that in striped, there was a magic member in each `ArrayObject` for the counts array.

lgray: Making it configurable is no problem.
It'll also not really be needed when we no longer need to go through pandas and can just awkward.fromarrow() things.

lgray: But I think this does force us to conform to a certain standard when we want something to scale out.... I think it's OK to eat that for now?

PR #82, opened 2019-05-26
Update spark_executor.py

njutxw:
update


lgray:
While I personally have nothing against comments in Chinese, these additions are not actually updates to this code, merely comments. Furthermore these comments seem to be related to basic features of python and not really about anything concerning what the code actually does, or expressing general confusion.

Furthermore this fails flake8. Unless there are substantial additions to the code you wish to make, I will close this PR without merging in 2 days.

PR #83, opened 2019-05-26
Update spark_executor.py

njutxw:
some annotations


lgray:
This appears to be a copy of #82. Closing.

PR #86, opened 2019-05-29
user can specify max chunks to process with lazy chunking

aminnj:
Based on https://github.com/CoffeaTeam/fnal-column-analysis-tools/issues/85

Idea of the PR is to make small-scale tests more rapid:

* Add `maxchunks` parameter to `run_uproot_job`, which limits processing of each dataset to that many chunks (if not `None`).
* Add lazy version (`_get_chunking_lazy`) of `_get_chunking`.
* Make `_get_chunking` use multi-threaded `uproot.numentries` with futures

If the user specifies `maxchunks`, the lazy chunking function is used. Otherwise, the blocking (but multi-threaded with 10 workers) one is used. If there are less than 5 files, it just uses a single thread to not get bitten by the overhead. I didn't test that cutoff too rigorously.

Some numbers:
for ~280 NanoAOD files in hadoop accessed via fuse mount, multithreading gives the per-file event counts in ~10-15 seconds. Otherwise, it's ~30-60seconds.

### DIFF
--- a/fnal_column_analysis_tools/processor/executor.py
+++ b/fnal_column_analysis_tools/processor/executor.py
@@ -74,11 +74,12 @@ def _work_function(item):
 @lru_cache(maxsize=128)
 def _get_chunking(filelist, treename, chunksize):
     items = []
-    executor = None if len(filelist)<5 else concurrent.futures.ThreadPoolExecutor(10)
-    for fn,nentries in uproot.numentries(filelist, treename, total=False, executor=executor).items():
+    executor = None if len(filelist) < 5 else concurrent.futures.ThreadPoolExecutor(10)
+    for fn, nentries in uproot.numentries(filelist, treename, total=False, executor=executor).items():
         for index in range(nentries // chunksize + 1):
             items.append((fn, chunksize, index))
     return items
+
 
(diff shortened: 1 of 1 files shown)


lgray:
Nice addition! 

For running on resource constrained batch systems it might be a nice idea to extend this functionality to ensure staying within RSS bounds? i.e. actually use lazyarray in awkward?

I'll make an issue here, something to think about.

fnal_column_analysis_tools/processor/executor.py
```diff
@@ -74,14 +74,21 @@ def _work_function(item):
 @lru_cache(maxsize=128)
 def _get_chunking(filelist, treename, chunksize):
     items = []
-    for fn in filelist:
-        nentries = uproot.numentries(fn, treename)
+    executor = None if len(filelist) < 5 else concurrent.futures.ThreadPoolExecutor(10)
```
nsmith-, line 77: I was going to ask if this hardcoded 10 thread option could be configured via module global or something.

aminnj: By "module global" is there already some kind of configuration file somewhere that you could point me to?

PR #93, opened 2019-05-31
Travis cleaning

lgray:
- still run awkward numba tests on the right versions of it
- make it so that we can deploy automatically (I hope)


lgray:
Problems in fcat.array and I need to enforce the right versions of uproot / uproot-methods for awkward 0.8.15.

Issue #96, opened 2019-06-07
iterative_executor crashes


aminnj:
I think because of https://github.com/CoffeaTeam/coffea/blob/1b0b8fac4ce0d774e4c8785a2c43188b95c12667/coffea/processor/executor.py#L116 
and
https://github.com/CoffeaTeam/coffea/blob/1b0b8fac4ce0d774e4c8785a2c43188b95c12667/coffea/processor/executor.py#L130,
https://github.com/CoffeaTeam/coffea/blob/1b0b8fac4ce0d774e4c8785a2c43188b95c12667/coffea/processor/executor.py#L30 will fail. Adding `workers=1` to the argument list of `iterative_executor` fixes it for me

nsmith-:
ack yes, anyway `workers` in the context of futures executor can and should be different from the number of parallel threads used to preprocess the files (and compute the chunking)

lgray:
I'll supply a PR and some tests in 1 or 2 hours, unless you wanna do it @nsmith-.

nsmith-:
Sounds like you're on it, thanks!

lgray:
@aminnj Fixed in #99 and v0.5.1 of coffea.

PR #99, opened 2019-06-07
Fix crash in iterative executor from workers

lgray:
Addresses #96.

- have separate "pre_workers" for preprocessing
- add tests

@nsmith-


nsmith-:
There must be a better pattern than using `**kwargs` as a gutter for extra params, but otherwise OK.

nsmith-:
Also can we make codecov shut up?

lgray:
Yeah I'll get it to be quiet.

lgray:
@aminnj This will be fixed in v0.5.1, or already in master.

PR #108, opened 2019-06-12
Try using xenial with java8

lgray:
As title, see if we can keep away from testing in ubuntu 14.xx...


lgray:
Nope, guess not, save this one for later for or anyone who feels like getting java8 to work on recent ubuntu.

PR #109, opened 2019-06-12
Make value_accumulator class behave like defaultdict

nsmith-:
i.e. use a default factory, and don't call it 'accumulator' because
that's really confusing in some of the other source files.


nsmith-:
Oh, and also keep track of xrootd data servers in uproot jobs.

lgray:
@nsmith- parsl and spark executors too.

PR #120, opened 2019-06-23
custom treename per filelist entry

lukasheinrich:
this allows a specification of a filelist like this

```
fileset = {
    'ttbar': {
        'treename': 'ttbar_mc16a_Nom',
        'files': glob.glob('../exported/ttbar_mc16a/*.root')
    },
    'CC1N1_higgsino_300_mc16a': {
        'treename': 'C1N1_higgsino_300_mc16a_Nom',
        'files': glob.glob('../exported/C1N1_higgsino_300_mc16a/*.root')
    },
}

output = processor.run_uproot_job(
    fileset,
    processor_instance=MyProcessor(),
    executor=processor.futures_executor,
    executor_args={'workers': 10, 'function_args': {'flatten': True}},
)
```

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -109,7 +109,7 @@ def _get_chunking_lazy(filelist, treename, chunksize):
             yield (fn, chunksize, index)
 
 
-def run_uproot_job(fileset, treename, processor_instance, executor, executor_args={}, chunksize=500000, maxchunks=None):
+def run_uproot_job(fileset, processor_instance, executor, executor_args={}, chunksize=500000, maxchunks=None):
     '''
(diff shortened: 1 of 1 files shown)


lgray:
@lukasheinrich Right now this patch removes the option of specifying a single tree name for all datasets (typical CMS setup). I think what should happen is that the original case should be the default and if the common tree name isn't passed in and there's a treename specified in the call to `run_uproot_job`.

This is what's causing the tests to fail.

lgray:
@lukasheinrich any movement on this? There's a little bit of homework for you. :-)

lgray:
This is fully fixed with backward compatibility in #128

PR #121, opened 2019-06-24
Column accumulator

nsmith-:
An accumulator that holds a numpy array that grows along its first
dimension as instances are added together.

e.g.
```python
import numpy as np
from coffea.processor import column_accumulator

a = column_accumulator(np.arange(6).reshape(2,3))
b = column_accumulator(np.arange(12).reshape(4,3))
a += b
a.value
```
produces:
```
array([[ 0,  1,  2],
       [ 3,  4,  5],
       [ 0,  1,  2],
       [ 3,  4,  5],
       [ 6,  7,  8],
       [ 9, 10, 11]])
```

### DIFF
--- a/binder/accumulators.ipynb
+++ b/binder/accumulators.ipynb
@@ -85,9 +85,9 @@
     {
      "data": {
       "text/plain": [
-       "{'cutflow': defaultdict_accumulator(int, {'pt>200': 3}),\n",
-       " 'pt': <Hist (pt) instance at 0x12033a940>,\n",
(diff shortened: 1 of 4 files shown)


nsmith-:
I guess `(a, *b)` is python 3.7+ actually

nsmith-:
Should this be defined using factory style to be consistent with the others?
To some extent the empty numpy array serves as a factory, e.g. `np.zeros(shape=(0,))`

PR #126, opened 2019-06-27
Fix bug in hist Interval hash

nsmith-:


lgray:
How critical is this?

nsmith-:
Allows plot1d to overlay dense binned axes.  Not super urgent but nice to have

PR #134, opened 2019-07-05
Trying to make travis happy...

lgray:
First try using --no-cache-dir.


lgray:
The solution was pinning pyarrow to `0.13.0`, there appears to be an incompatibility between pyarrow `0.14.0` and pip-installed pyspark.

lgray:
@matthewfeickert for your reference

PR #139, opened 2019-07-12
Fix and testing for Clopper Pearson interval

### LINKED ISSUE #138: coffea.hist.plot.clopper_pearson_interval gives wrong result for numerator==denominator
**Describe the bug**
If numerator == denominator, clopper_pearson_interval always returns [1,1] as the corresponding interval.

**To Reproduce**
```python
from coffea.hist.plot import clopper_pearson_interval
clopper_pearson_interval(10,10,0.95)
# Result: array([1., 1.])

````
**Expected behavior**
I expected to get array([0.69150289, 1.])

**Proposed fix**
This is trivially fixed by changing [this line](https://github.com/CoffeaTeam/coffea/blob/2ae59e8c75fa8e5af9a245e27bd7aa363a4f116b/coffea/hist/plot.py#L66) from
```python
interval[:, num == denom] = 1.
```
to 
```python
interval[1, num == denom] = 1.
```

If you agree, I'd be happy to make a PR.

AndreasAlbert:
Fixes #138

### DIFF
--- a/coffea/hist/plot.py
+++ b/coffea/hist/plot.py
@@ -63,7 +63,7 @@ def clopper_pearson_interval(num, denom, coverage=_coverage1sd):
     hi = scipy.stats.beta.ppf((1 + coverage) / 2, num + 1, denom - num)
     interval = np.array([lo, hi])
     interval[:, num == 0.] = 0.
-    interval[:, num == denom] = 1.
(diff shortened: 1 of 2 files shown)


lgray:
Thanks a lot! I'll mint a new version momentarily.

PR #142, opened 2019-07-16
Use laurelin to read root files in spark tests

lgray:
As in title.

See the tests for examples.
@nsmith- 

@perilousapricot - I'll send you a little test script so you can parameterize the file type and do speed tests. On my mac laurelin is about a factor of 3 slower than parquet for opening and counting the files.


PerilousApricot:
:+1: A factor of 3 isn't great, but considering there's zero optimization, I'm not too disappointed. I'm writing up a plan of action to get caching done rite https://github.com/PerilousApricot/spark-ttree/milestone/2

PR #146, opened 2019-07-22
Fix for JEC dataera that contains an underscore

jpata:
Fix crash on parsing a JEC name `Fall17_17Nov2017_V6_MC_UncertaintySources_AK4PFchs_AbsoluteFlavMap` where the dataera contains an underscore. I tested that evaluation seems to work fine now, also added a test for the parsing.


lgray:
The new jec file that you added, can you gzip it just to keep the repo size down?

PR #153, opened 2019-07-28
Improve hist interface

### LINKED ISSUE #90: Different interfaces for hist.rebin() and hist.group()
```
Signature: hfcat.rebin(old_axis, new_axis)
```
vs.
```
Signature: hfcat.group(new_axis, old_axes, mapping, overflow='none')
```

Since these are conceptually similar operations, ideally the order of arguments would be the same (old_axis, new_axis, ...).

nsmith-:
Fixes #90 

Tackles some outstanding issues with the `Hist` object interface

### DIFF
--- a/coffea/hist/hist_tools.py
+++ b/coffea/hist/hist_tools.py
@@ -847,6 +847,22 @@ class Hist(AccumulatorABC):
     def profile(self, axis_name):
         raise NotImplementedError("Profiling along an axis")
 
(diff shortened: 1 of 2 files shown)


lgray:
can you rebase this real quick so the tests run properly?

lgray:
Ah, I suppose I can do that...

lgray:
Fixes #90

PR #157, opened 2019-08-06
adds six as a install requirement

lukasheinrich:
`coffea` seems to import the python 2/3 compatibility library `six` in a few places, but it's not listed as a requirement in `setup.py`. This PR adds this.


lukasheinrich:
(though it's a bit weird, since `setup.py` itself requires `six` so it won't install without it anyways.

PR #158, opened 2019-08-23
Propagate renaming of 'project' function to plotting.

AndreasAlbert:
In c88a861, `project` was renamed to `integrate`, but a few spots were left unchanged. I suggest to also change them because currently I receive `DeprecationWarnings` from non-custom code inside the coffea package, which is not desirable.


nsmith-:
ahh thanks

PR #163, opened 2019-09-19
Fix unexpected keyword

### LINKED ISSUE #162: Properly handle case where `data_flow=None` is passed to `run_parsl_job`
When `data_flow=None` is passed to [`run_parsl_job`](https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L193) with multiple executor_args, the following exception is raised:
```
~/.local/lib/python3.6/site-packages/coffea/processor/executor.py in run_parsl_job(fileset, treename, processor_instance, executor, data_flow, executor_args, chunksize)
    248     killParsl = False
    249     if data_flow is None:
--> 250         data_flow = _parsl_initialize(**executor_args)
    251         killParsl = True
    252     else:

TypeError: _parsl_initialize() got an unexpected keyword argument 'timeout'
```

annawoodard:
The `_parsl_initialize` function is only expecting one argument. Fixes #162.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -247,7 +247,7 @@ def run_parsl_job(fileset, treename, processor_instance, executor, data_flow=Non
     # when we're done
     killParsl = False
     if data_flow is None:
-        data_flow = _parsl_initialize(executor_args['config'])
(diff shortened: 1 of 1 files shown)


lgray:
Thanks for the catch!

PR #165, opened 2019-09-19
Fix hangs in Parsl executor

### LINKED ISSUE #164: Fix file-opening hangs with Parsl
When running `run_parsl_jobs` (as called in [this](https://github.com/annawoodard/coffeandbacon/blob/update-parsl/analysis/baconbits-parsl.ipynb) notebook) at Notre Dame, I encountered file-opening hangs (~several hangs / ~several hundred tasks). The implemented [timeout](https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/parsl/detail.py#L61) is not working as expected in my case. I determined that after the attempt is made to open the file, in a handful cases, the process hangs and the timeout has no effect. Checking the process with `ps` shows it still alive, but it never proceeds past the file open call. In general I don't think a python-space timeout is enough, if the hang is happening in something xrootd-related under-the-hood somewhere. In this case I think we need something a bit more heavy-handed-- a signal interruption to the process. This will kill the worker, so I've implemented a Parsl-side worker watchdog (see [here](https://github.com/Parsl/parsl/pull/1294)).

For reference, the Parsl config I am using is included below.

```python
import os
from parsl.providers import CondorProvider
from parsl.config import Config
from parsl.executors import HighThroughputExecutor
from parsl.addresses import address_by_hostname

proxy = '/tmp/x509up_u{}'.format(os.getuid())

worker_init = '''

source /cvmfs/sft.cern.ch/lcg/views/LCG_95apython3/x86_64-centos7-gcc7-opt/setup.sh
export PATH=~/.local/bin:$PATH
export PYTHONPATH=~/.local/lib/python3.6/site-packages:$PYTHONPATH

export X509_USER_PROXY=`pwd`/{}
'''.format(os.path.basename(proxy))

config = Config(
    executors=[
        HighThroughputExecutor(
            address=address_by_hostname(),
            cores_per_worker=1,
            worker_debug=True,
            provider=CondorProvider(
                cores_per_slot=8,
                init_blocks=80,
                max_blocks=80,
                worker_init=worker_init,
                transfer_input_files=[proxy],
            ),
        )
    ],
    retries=5,
    strategy=None
)
```

annawoodard:
Major changes:
1) The timeout decorator is now applied to all Parsl apps
   (previously it was not being used). This switches the timeout mechanism
   from a threading-exception-based timeout to a more heavy-handed
   signal-based timeout which interrupts the process.
2) A few bugs in the timeout decorator are fixed: always pass it a
   an int, and disable the alarm after the task has finished.
3) The number of allowed retries is now controlled by the `retries` keyword arg
   to the Parsl app, rather than hard-coded in the app definition.

Fixes #164.

### DIFF
--- a/coffea/processor/parsl/detail.py
+++ b/coffea/processor/parsl/detail.py
@@ -13,6 +13,7 @@ from parsl.config import Config
 from parsl.executors import HighThroughputExecutor
 
 from ..executor import futures_handler
+from .timeout import timeout
 
 try:
     from collections.abc import Sequence
@@ -45,29 +46,16 @@ def _parsl_stop(dfk):
     parsl.clear()
 
 
+@timeout
 @python_app
 def derive_chunks(filename, treename, chunksize, ds, timeout=10):
     import uproot
     from collections.abc import Sequence
-    from concurrent.futures import ThreadPoolExecutor, TimeoutError
 
     uproot.XRootDSource.defaults["parallel"] = False
 
-    afile = None
-    for i in range(5):
-        with ThreadPoolExecutor(max_workers=1) as executor:
-            future = executor.submit(uproot.open, filename)
-            try:
-                afile = future.result(timeout=timeout)
-            except TimeoutError:
-                afile = None
-            else:
-                break
-
-    if afile is None:
-        raise Exception('unable to open: %s' % filename)
-
     afile = uproot.open(filename)
+
     tree = None
     if isinstance(treename, str):
         tree = afile[treename]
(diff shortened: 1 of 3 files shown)


lgray:
Excellent, thank you!

nsmith-:
Is this by chance related to https://github.com/scikit-hep/uproot/pull/270 ?
It would be good to understand why we get timeouts.

lgray:
@nsmith- no, this happens with that env variable set. This is somewhere deeper in xrootd or at least in the interaction between parsl and xrootd.

annawoodard:
cc @klannon

PR #167, opened 2019-09-20
Simplify parsl dfk

annawoodard:
This PR removes explicitly passing around the Parsl DFK. It's not a critical change, I just think it simplifies things a bit.


lgray:
Looks like you need to update the tests as well!

annawoodard:
Thanks @lgray! The tests caught a real bug, fixed in 9d3357e. Things should be good to go now.

PR #171, opened 2019-09-25
More docstrings

nsmith-:
As a PR because some code changes were made to lumi tools to make numba kernel functions private


lgray:
Thanks!

PR #175, opened 2019-10-10
Use tqdm.autonotebook and require tqdm version

paulgessinger:
`tqdm.autonotebook` allows automatically showing fancy progress bars. 

Any reason why this wouldn't be a good idea?


lgray:
Hey - the build failure looks spurious, restarting that job.

This looks OK to me, thanks for catching it!

nsmith-:
Is it expected to keep seeing a warning for this?
```/usr/local/lib/python3.7/site-packages/tqdm/autonotebook.py:17: TqdmExperimentalWarning: Using `tqdm.autonotebook.tqdm` in notebook mode. Use `tqdm.tqdm` instead to force console mode (e.g. in jupyter console)
  " (e.g. in jupyter console)", TqdmExperimentalWarning)```

paulgessinger:
I think it is. I didn’t catch this because it didn’t show up on my machine for some reason, but I think I can circumvent it by importing `tqdm.auto`, which I believe doesn’t print the warning. 

Sorry about that.

PR #176, opened 2019-10-11
Add __getattr__ to LazyDataFrame

paulgessinger:
Using `__getattr__`, one can access columns from lazy dataframe like `df.electron_pt` in addition to `df["electron_pt"]`. This just forwards to `__getitem__` which is obviously already implemented.

Pandas does something like this (they're a bit more clever I believe), so I thought it makes sense to have it here as well. (Also I prefer it 😄 )

### DIFF
--- a/coffea/processor/dataframe.py
+++ b/coffea/processor/dataframe.py
@@ -121,6 +121,12 @@ class PreloadedDataFrame(MutableMapping):
         self._accessed.add(key)
         return self._dict[key]
 
+    def __getattr__(self, key):
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            raise AttributeError(key)
+
     def __iter__(self):
         for key in self._dict:
             self._accessed.add(key)
--- a/coffea/processor/dataframe.py
+++ b/coffea/processor/dataframe.py
@@ -52,7 +52,10 @@ class LazyDataFrame(MutableMapping):
             raise KeyError(key)
 
     def __getattr__(self, key):
-        return self.__getitem__(key)
+        try:
+            return self.__getitem__(key)
+        except KeyError:
+            raise AttributeError(key)
 
     def __iter__(self):
         for item in self._dict:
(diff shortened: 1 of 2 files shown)


lgray:
Have you checked that this does not interfere with access to class member functions in the base class or derived classes?

paulgessinger:
As per https://docs.python.org/3/reference/datamodel.html#object.__getattr__, the interpreter tries to look up the given key in the class itself as well as the class tree, before invoking `__getattr__`, so I think it should be safe.

lgray:
Seems to be the same for python2.

Can you add tests of the new functionality to:
https://github.com/CoffeaTeam/coffea/blob/master/tests/test_processor.py
(data frame tests at the bottom)

Thanks!

nsmith-:
Yeah there was a lot of FUD about `__getattr__` that turns out to be mostly related to the much more dangerous `__getattribute__`.  But nevertheless, a google search shows that it can be implemented safely.

coffea/processor/dataframe.py
```diff
@@ -51,6 +51,9 @@ def __getitem__(self, key):
         else:
             raise KeyError(key)
 
+    def __getattr__(self, key):
+        return self.__getitem__(key)
```
nsmith-: So one issue here is that when an item isn't  there, a `KeyError` is raised which leads to a misleading error message and less useful stack trace.  To avoid, maybe better to format as:
```python
        try:
            return self[key]
        except KeyError:
            raise AttributeError
```

nsmith- (changes requested):
While you're there, can you add also the same to PreloadedDataFrame?

coffea/processor/dataframe.py
paulgessinger: Should be this way now.

paulgessinger:
I added tests, basically copied from the ones for `__getitem__` with minor changes. I didn't implement `__setattr__` (this is a bit difficult), so I dropped one assertion in the copied test.

coffea/processor/dataframe.py
lgray: Looks like it's all there now.

lgray (approved):

lgray:
Looks all good to me. Thanks for adding the tests and improve the error reporting!

PR #177, opened 2019-10-16
Plotting updates to address some issues

nsmith-:
Of note,  `Hist.project` is now implemented in its correct form.  Hopefully the FutureWarning from before was enough to make people migrate to `Hist.integrate`.


lgray:
WTF Codecov.... Lemme kick it a bit.

OK - should be updated now. I'll go ahead and merge this.

lgray:
*grumble* --I'll get it to behave, but later.

PR #181, opened 2019-10-16
Implement a dask executor

nsmith-:
Provides a new `dask_executor` suitable for use as an alternative executor in `run_uproot_job`.
For example, the muonspectrum demo notebook can run in a local dask cluster with minimal changes:
```python
from dask.distributed import Client
client = Client()

output = processor.run_uproot_job(fileset,
                                  treename='Events',
                                  processor_instance=FancyDimuonProcessor(),
                                  executor=processor.dask_executor,
                                  executor_args={'client': client, 'flatten': True},
                                  chunksize=100000,
                                 )
```

p.s. can parsl_executor be encapsulated this way?

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -339,8 +339,7 @@ def parsl_executor(items, function, accumulator, **kwargs):
         parsl.clear()
         parsl.load(config)
 
-    # wrap with timeout() ?
-    app = python_app(function)
+    app = timeout(python_app(function))
 
     futures = set(app(item) for item in items)
     _futures_handler(futures, accumulator, status, unit, desc, add_fn)
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -195,7 +195,7 @@ def dask_executor(items, function, accumulator, **kwargs):
     while len(futures) > 1:
         futures = client.map(
             reducer,
-            [futures[i * ntree:(i + 1) * ntree] for i in range(len(futures) // ntree + 1)],
+            [futures[i:i + ntree] for i in range(0, len(futures), ntree)],
             priority=1,
         )
     if status:
@@ -206,7 +206,7 @@ def dask_executor(items, function, accumulator, **kwargs):
 
 
 def parsl_executor(items, function, accumulator, **kwargs):
-    """Execute using parsl pyapp wrapper (TODO)
+    """Execute using parsl pyapp wrapper
 
     Parameters
     ----------
@@ -218,6 +218,8 @@ def parsl_executor(items, function, accumulator, **kwargs):
             An accumulator to collect the output of the function
         config : parsl.config.Config, optional
             A parsl DataFlow configuration object. Necessary if there is no active kernel
+
+            .. note:: In general, it is safer to construct the DFK with ``parsl.load(config)`` prior to calling this function
         status : bool
             If true (default), enable progress bar
         unit : str
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -7,6 +7,7 @@ import pickle
 import sys
 import math
 from tqdm.auto import tqdm
+from collections import defaultdict
 from cachetools import LRUCache
 import lz4.frame as lz4f
 from .processor import ProcessorABC
@@ -22,14 +23,8 @@ from .dataframe import (
 
 try:
     from collections.abc import Mapping, Sequence
-    from functools import lru_cache
 except ImportError:
     from collections import Mapping, Sequence
-
-    def lru_cache(maxsize):
-        def null_wrapper(f):
-            return f
-        return null_wrapper
 
 
 # instrument xrootd source
@@ -486,7 +481,7 @@ def run_uproot_job(fileset,
                 chunks.append(chunk)
     else:
         # get just enough file info to compute chunking
-        nchunks = {}
+        nchunks = defaultdict(int)
         while fileset:
             filemeta = fileset.pop()
             if nchunks[filemeta.dataset] >= maxchunks:
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -1,7 +1,10 @@
 import concurrent.futures
+from functools import partial
 import time
-from tqdm.autonotebook import tqdm
 import uproot
+import pickle
+from tqdm.auto import tqdm
+import lz4.frame as lz4f
 from . import ProcessorABC, LazyDataFrame
 from .accumulator import value_accumulator, set_accumulator, dict_accumulator
 
@@ -17,6 +20,8 @@ except ImportError:
         return null_wrapper
 
 
+OUTPUT_COMPRESSIONLVL = 1
+
 # instrument xrootd source
 if not hasattr(uproot.source.xrootd.XRootDSource, '_read_real'):
     def _read(self, chunkindex):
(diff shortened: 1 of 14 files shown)


nsmith-:
@aminnj I would be grateful if you could try this out on a cluster and see if it works for you.

coffea/processor/executor.py
```diff
@@ -107,6 +107,53 @@ def futures_executor(items, function, accumulator, workers=1, status=True, unit=
...
+            If true (default), enable progress bar
+    """
+    client = kwargs.pop('client')
+    ntree = kwargs.pop('treereduction', 20)
+    status = kwargs.pop('status', True)
+    futures = client.map(function, items, **kwargs)
+    while len(futures) > 1:
+        futures = client.map(
+            _dask_reduce,
+            [futures[i * ntree:(i + 1) * ntree] for i in range(len(futures) // ntree + 1)],
```
nsmith-: The tree reduction here doesn't explicitly bind tasks to the same worker, but it would be good to know if dask does this anyway after analyzing the task graph.

coffea/processor/executor.py
```diff
@@ -107,6 +107,53 @@ def futures_executor(items, function, accumulator, workers=1, status=True, unit=
...
+            A dask distributed client instance
+        treereduction : int, optional
+            Tree reduction factor for output accumulators (default: 20)
+        status : bool, optional
+            If true (default), enable progress bar
+    """
+    client = kwargs.pop('client')
+    ntree = kwargs.pop('treereduction', 20)
+    status = kwargs.pop('status', True)
+    futures = client.map(function, items, **kwargs)
```
nsmith-: For large tasks, the graph construction hangs for an annoyingly long time.  I am not sure if I'm setting this up improperly, or what.

coffea/processor/executor.py
nsmith-: Seems so: https://distributed.dask.org/en/latest/journey.html#step-3-select-a-worker

coffea/processor/executor.py
nsmith-: `item` includes the processor instance, and since the rest of the tuple changes, dask doesn't see that its always the same.  This leads to shipping the processor `N` times for n tasks.  Solution should be to create a closure function carrying the processor so that `item` is lightweight.

lgray:
@nsmith- So parsl and spark executor can be encapsulated this way, but it requires an executor refactor into some classes. I'll add an issue and start on it soon.

coffea/processor/executor.py
```diff
@@ -1,7 +1,10 @@
 import concurrent.futures
+from functools import partial
 import time
-from tqdm.autonotebook import tqdm
 import uproot
+import pickle
+from tqdm.auto import tqdm
```
lgray, line 11: @paulgessinger has made a separate PR to take care of this
@nsmith-  can you remove the tqdm change from this PR and we can go ahead with the other one?

coffea/processor/executor.py
```diff
@@ -99,16 +127,126 @@ def futures_executor(items, function, accumulator, workers=1, status=True, unit=
...
+            priority=1,
+        )
+    if status:
+        from dask.distributed import progress
+        progress(futures, multi=False, notebook=False)
+    accumulator += futures[0].result()
+    return accumulator
+
+
+def parsl_executor(items, function, accumulator, **kwargs):
```
nsmith-, line 289: @lgray I actually started on the parsl conversion, leaving a sketch here. (it's not forwarded to `__init__.py` yet)

lgray: OK - cool. I'll think about spark then.

aminnj:
Hi, this is great! I'm testing it out over HTCondor now. It seems when the number of chunks is divisible by `ntree` then 
https://github.com/CoffeaTeam/coffea/blob/2123a6a178fd7932395df8ff5b74d19df4df88c1/coffea/processor/executor.py#L198
will create an extra empty list which triggers
https://github.com/CoffeaTeam/coffea/blob/2123a6a178fd7932395df8ff5b74d19df4df88c1/coffea/processor/executor.py#L64
This seems to work:
```python
[futures[i:i + ntree] for i in range(0, len(futures), ntree)],
```

coffea/processor/executor.py
```diff
@@ -99,16 +142,136 @@ def futures_executor(items, function, accumulator, workers=1, status=True, unit=
...
+    if cleanup and config is None:
+        raise RuntimeError("No active parsl DataFlowKernel, must specify a config to construct one")
+    elif not cleanup and config is not None:
+        raise RuntimeError("An active parsl DataFlowKernel already exists")
+    else:
+        parsl.clear()
+        parsl.load(config)
+
+    # wrap with timeout() ?
+    app = python_app(function)
```
nsmith-: @lgray I noticed what appears to be a watchdog timer built into parsl now.  Do you know if that is new, or it was there and didn't work?

lgray: It may still not work for python apps. In 0.8.0 it definitely doesn't.
@annawoodard what's the status in parsl master?

annawoodard: @lgray @nsmith: The [parsl timeout](https://github.com/Parsl/parsl/blob/master/parsl/app/python.py#L14-L32) works by starting a timer and injecting an exception into the current thread if the timer goes off. This has the advantage that it doesn't kill the Parsl worker, but it has the disadvantage that it may not work if the hang is happening outside Python (for example, in some underlying C++ code XrootD is running). I recommend you use the timeout implemented [here](https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/parsl/timeout.py) which uses a SIGALRM to blast the whole worker. On master, I've added the machinery to transparently restart the worker. As long as you have `retries > 0` in your Parsl config, any timed-out tasks will automatically be resubmitted. I've tested this works on XrootD hangs.

aminnj:
I'm running my notebook from yesterday after updating this PR, but now I'm getting the error below. Looks like the chunking is done non-locally now. Should stuff work out of the box, or do I need to change my inputs somehow?

```
AttributeError                            Traceback (most recent call last)
<ipython-input-290-f4de6a570dab> in <module>
     11                                   executor=processor.dask_executor,
     12                                   executor_args={'client': c, 'flatten': True, "savemetrics":True,"compression":0},
---> 13                                   chunksize=100000000,
     14 
     15                                  )

~/miniconda3/envs/analysisenv/lib/python3.7/site-packages/coffea/processor/executor.py in run_uproot_job(fileset, treename, processor_instance, executor, executor_args, chunksize, maxchunks)
    394             local_treename = treename
    395         else:
--> 396             raise ValueError('list of filenames in fileset must be a list or a dict')
    397         for filename in filelist:
    398             yield FileMeta(dataset, filename, local_treename)

~/miniconda3/envs/analysisenv/lib/python3.7/site-packages/coffea/processor/executor.py in dask_executor(items, function, accumulator, **kwargs)
    209     """
    210     if len(items) == 0:
--> 211         return accumulator
    212     pool = kwargs.pop('pool', concurrent.futures.ProcessPoolExecutor)
    213     workers = kwargs.pop('workers', 1)

~/miniconda3/envs/analysisenv/lib/python3.7/site-packages/distributed/client.py in result(self, timeout)
    220         result = self.client.sync(self._result, callback_timeout=timeout, raiseit=False)
    221         if self.status == "error":
--> 222             six.reraise(*result)
    223         elif self.status == "cancelled":
    224             raise result

~/miniconda3/envs/analysisenv/lib/python3.7/site-packages/six.py in reraise(tp, value, tb)
    690                 value = tp()
    691             if value.__traceback__ is not tb:
--> 692                 raise value.with_traceback(tb)
    693             raise value
    694         finally:

/srv/temp/workerenv/lib/python3.7/site-packages/coffea/processor/executor.py in _work_function()

AttributeError: 'tuple' object has no attribute 'filename'
```

aminnj:
Not sure what the issue was, but I re-updated the various environments (dask worker, scheduler, local notebook) and now it works. I guess I missed something earlier. Sorry for the noise!

nsmith-:
Yeah, chunking is now farmed out to the executor, or to `pre_executor` if provided.  You can use the latter if you wish chunking to still be a local operation.

lgray:
@aminnj When you're happy with reproducibility I'd be really curious about processing speed / user experience.

Would you be interested in going through what you've worked on with dask next Friday 2pm CST?

aminnj:
Sure, I'll write back here when I can get more condor slots to test at scale (seems something's up with my T2 today). Though yesterday, I was running the `FancyDimuonProcessor` from [here](https://github.com/CoffeaTeam/coffea/blob/master/binder/muonspectrum_v4.ipynb) with 50 workers on DoubleMuon NANOAOD and getting ~1.6MHz. 

One thing I didn't have a chance to look into was a 1-2min delay between preprocessing chunks (done locally at that time) and the start of the dask progress bar. I had O(1k) chunks, so maybe it's a graph construction issue like was mentioned earlier in this thread.

I'm free next Friday. Let me know if you are curious about anything in particular so I can be a bit more prepared.

lgray:
I think I'm mostly interested in user-facing performance and experience.
- How easy was your workflow to setup, what problems did you have setting it up?
- What sort of site reliability issues get in your way?
- What's the variability of getting good throughput?
- Start-to-finish time to run a full analysis with systematics for 2017 or 2018, if you have it?

Essentially, stuff along those lines and an introduction to what you've setup! Feel free to include anything else you think is interesting from your work here.

1.6 MHz is pretty good with 50 workers, but I'm sorta surprised it's only that with FancyDiMuonProcessor. It isn't really doing much! How many cores per worker?

aminnj:
OK, I can address most of those questions on Friday, but I didn't get beyond playing around with toy-ish examples, so no full analysis/systematics :(


I had one core per worker. I ran one instance of the processor in my local notebook just now with a 500k event chunk from one file over xrootd (cold cache) and got this profile:

image

That's ~11s @ ~47kHz per core/worker, which is in the ballpark of ~2MHz if I had 50 remote workers. The 1.6MHz I saw yesterday isn't far from that, so I don't think it's a dask/parallelization issue. I also tried locally without xrootd and it took 8 seconds still (and about 5.5s running it a second time with a warmer cache presumably). 

In your experience, how fast does/should the FancyDimuonProcessor run?

lgray:
Hey @aminnj I am getting ~60 kHz/core/worker (all files local) so I guess we're in the same ball park, must have recalled a simpler Processor.

The low overhead for dask does indeed look quite attractive.

PR #183, opened 2019-10-17
Add tests against python 3.8

lgray:
As in title.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest] #
         java-version: [1.8]
-        python-version: [3.6, 3.7] #
+        python-version: [3.6, 3.7, 3.8] #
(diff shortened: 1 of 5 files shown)


lgray:
Hmm wheels for scipy not available yet...

nsmith-:
I think 3.6 is going to take some time to remove.  It's the latest LCG python.

lgray:
Yeah, fair enough, guess we'll just have more tests for a while.

Makes a good point to move to azure pipelines more free jobs.

lgray:
Some slow improvement here: numpy and scipy now have wheels, waiting on numba and llvmlite now!

nsmith-:
Is there any reason to drop 3.6?

lgray:
Nope and it's not since I rebased it.

pyarrow is still missing python 3.8 wheels, no good reason to add it yet.

lgray:
Damn, pyarrow 0.16.0 is out but not on pypi yet....

PR #188, opened 2019-10-17
Fix issue with JaggedCandidateMethods nested cross

nsmith-:
Code was untouched since some very old version of awkward.
Whatever bug it was working around no longer exists, it seems


lgray:
Yeah there was an old bug where cross wouldn't continue working if some columns became out of order. Seems no longer necessary, thanks!

PR #190, opened 2019-10-22
test and fix for 2D JER SF lookup

jpata:
Added a unit test for 2D lookup of JER SF, needed for the latest Autumn18_V7_MC_SF tag, which fails. An issue will follow shortly :)


jpata:
Here's the related issue: https://github.com/CoffeaTeam/coffea/issues/191

jpata:
Seems like https://github.com/CoffeaTeam/coffea/pull/190/commits/72ca7a125512adf031d5de58493a38a3246fc8d5 fixed it: the output SF arrays are already jagged with the correct structure, so the flattening can be done afterwards, treating 1D as a special case of n-D.

lgray:
@jpata Can you make that automatically happen (if it's not already)?

jpata:
As far as I can tell it is already automatic, in the sense that both the original 1D test and the new 2D test pass and there is no `if/else` required. In the 1D case, the `_parms` ends up being a jagged array with every element being count-1. I checked a few elements by hand and it seems to retrieve the correct values.

lgray:
Indeed, everything else looks good.

I'll go ahead and merge.

lgray:
Ah, right, speed.

Can you do a quick regression test on this for execution speed?

PR #198, opened 2019-11-02
Guard against extraneous axes in Hist.fill

### LINKED ISSUE #197: hist.fill kwargs not guarded
**Describe the bug**

The histogram fill function can be called with arbitrary keyword arguments. Unless there are very strong reasons for this, I think it is dangerous. A simple typo can send optional arguments like `weight` into a void. 

**To Reproduce**
```python
import numpy as np
from coffea.hist import Bin, Cat, Hist

dataset_ax = Cat("dataset", "Primary dataset")
pt_ax = Bin("pt", r"pt", 5, 0, 5)

h = Hist(
        "Counts",
        dataset_ax,
        pt_ax,
        )

h.fill(
    pt=np.array([1,2,3]),
    dataset='test',
    wieght=np.array([4,5,6]) # I mis-spelled "weight"
)
```

**Expected behavior**
Fill should raise an exception if an unknown kwarg is provided.

nsmith-:
Fixes #197

### DIFF
--- a/coffea/hist/hist_tools.py
+++ b/coffea/hist/hist_tools.py
@@ -875,7 +875,7 @@ class Hist(AccumulatorABC):
         Note
         ----
             The reserved keyword ``weight``, if specified, will increment sum of weights
-            by the given column values, which must have the same dimension as all other
+            by the given column values, which must be broadcastable to the same dimension as all other
             columns.  Upon first use, this will trigger the storage of the sum of squared weights.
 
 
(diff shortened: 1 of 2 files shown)


lgray:
Do we need a to raise an exception here or just print a warning/error message?

Some people may want to be lazy and always fill the histograms with the same columns. It wouldn't result in a wrong answer and could be a compact way to fill histograms in some cases.

@AndreasAlbert wants to protect against mistypes, which would be covered by just emitting an error message rather than flat out failing.

Or perhaps the behavior should be configurable?

AndreasAlbert:
thanks for the implementation @nsmith- 

@lgray I personally think an exception is appropriate, since the "use same arguments for all fill calls" tactic seems quite strange to me. If others really like having it as an option, though, a configurable solution would also be fine.

lgray:
@AndreasAlbert Fair enough, just making sure we don't write ourselves into a corner. 

Adding a test and then merging this.

lgray:
@nsmith- FYI you do this all the time in gghbbprocessor :-)

nsmith-:
@lgray https://github.com/nsmith-/coffeandbacon/commit/15faddf5a1c66dde6b5b9118536c4afa8472424d

lgray:
Yeah I fixed it different changes. I was just annoyed at the time. :-p

PR #199, opened 2019-11-02
Allow specification of axes by positional list or by kwarg sequence

lgray:
This was brought up in discussions with @andrzejnovak / @kratsg / @nsmith- that syntax like:
```
h = coffea.hist.Hist("Observed bird count",
                     axes=(coffea.hist.Cat("species", "Bird species"),
                           coffea.hist.Bin("x", "x coordinate [m]", 20, -5, 5),
                           coffea.hist.Bin("y", "y coordinate [m]", 20, -5, 5),
                           )
                     )
```

is much more clear to new users than the positional list syntax that we've been using so far.

This PR allows both styles of histogram instantiation.

### DIFF
--- a/coffea/hist/hist_tools.py
+++ b/coffea/hist/hist_tools.py
@@ -669,16 +669,16 @@ class Hist(AccumulatorABC):
                              coffea.hist.Bin("y", "y coordinate [m]", 20, -5, 5),
                              )
 
-        or
+        # or
 
         h = coffea.hist.Hist("Observed bird count",
                              axes=(coffea.hist.Cat("species", "Bird species"),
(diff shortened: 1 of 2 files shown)


nsmith-:
Would `hist.Hist(hist.Bin(...), ..., label='event count')` be more palatable?

lgray:
@nsmith- I think the main confusion was over the meaning of the axes and categories (as in they didn't know what kind of histogram they were building). It would make sense to let the label be possibly kwargs. I'll add that too.

lgray:
and so `hist.Hist(hist.Bin(...), ..., label='event count')` is already possible anyway so that's good.

lgray:
@nsmith- anything else before I merge this?

PR #200, opened 2019-11-07
Try pyarrow 0.15.1

lgray:
remove version exclusion in travis.


lgray:
and that's a big fat nope.

lgray:
pyarrow from 0.15.0 will work with spark 2.4.3 if you set `ARROW_PRE_0_15_IPC_FORMAT=1` in the environment.

PR #179, opened 2019-10-16
mplhep

andrzejnovak:
Start moving plotting to interface through mplhep

PR #209, opened 2019-11-19
Reverse stacking order

AndreasAlbert:
Follow-up to #177

The current implementation of integral sorting for an axis sorts in descending order, which is opposite of what is desirable for stack plots.


nsmith-:
Does this make the legend order now match the plot order, or do they both get reversed?
Also, thanks.  I remember someone complaining about this already and forgot to go fix it

AndreasAlbert:
Good point, stacking order in the legend is the opposite way. Let's just reverse that one, then, too



image

AndreasAlbert:
Actually, is there even a sensible way to do this? I don't know how other people handle this, but I do something along the lines of

```python
fig, ax, primitives = plot1d(...) # data
plot1d(..., ax=ax, clear=False) # MC
ax.legend()
```

So sure, I can change the legend ordering in `primitives['legend']`, but I am not sure how useful that is given that it will only have the MC or data entries, but never both.

nsmith-:
Well, in principle only the MC order is wrong in the legend.  Whether data is above or below seems less important, no?  So there, it would just be flipping the order of handles and labels in https://github.com/CoffeaTeam/coffea/blob/master/coffea/hist/plot.py#L298

nsmith-:
Ah, legend isn't made in plot1d in this case, hmm
@andrzejnovak thoughts?

nsmith-:
I guess rather than change the return order inside `hist.identifiers`, actually one has to rework `plot1d` to draw the top piece of the stack first, then the next... rather than accumulate the fill_between values as done now.  Perhaps #179 solves this?

AndreasAlbert:
Maybe we are overthinking this. Reversing the order on a legend you make yourself [is a two-liner](https://stackoverflow.com/questions/34576059/reverse-the-order-of-legend), so if someone wants this relatively specific aesthetic feature, its easy to do.

andrzejnovak:
I still feel a bit awkward (get it) about how "inaccessible" "the regular matplotlib interface becomes when plotting through coffea, but I haven't thought of a better way as of yet.

Usually if I want a custom sort (if it matters at all) there's a sorting function in mplhep https://github.com/andrzejnovak/mplhep/blob/master/mplhep/plot.py#L364 which get's used like this: `ax.legend(*hep.plot.sort_legend(ax, label_dict))` label_dict looking like 
```
label_dict = OrderedDict({
    'Data': 'Data',
    'MC': 'MC'})
```

nsmith-:
Semi-related, doesn't sorting by integral mean that depending on the plot you make, the colors assigned to various processes changes?

andrzejnovak:
Hmm, that makes sense to me, if sorting is adjusted, so should color assignment

andrzejnovak:
Good discussion, it looks like plotting has a bug related anyway. I imagine sorting matters mostly for log plots, but calling `plt.semilogy()` screws axes up.

image

image

andrzejnovak:
@nsmith- You are right, in #179 this looks like:

image

i.e. samples and legend are in the same order

AndreasAlbert:
@andrzejnovak the screwed up axes can be fixed by setting axis limits manually

@nsmith- if you want consistent colors, I think you have to set them by manually looping over the primitives anyway. Otherwise, if you e.g. have multiple control regions that share some but not all of their backgrounds, you won't always have identical colors for the same process.

andrzejnovak:
@AndreasAlbert I know, but it shouldn't happen by default.

nsmith-:
The original problem this PR solves is still valid, right? So I'll merge if tests pass

AndreasAlbert:
yes, please

jrueb:
@AndreasAlbert 
> @nsmith- if you want consistent colors, I think you have to set them by manually looping over the primitives anyway. Otherwise, if you e.g. have multiple control regions that share some but not all of their backgrounds, you won't always have identical colors for the same process.

I think the "integral" option is great but I can't have the categories change colors from one to the other plot. Could you explain how exactly to set the colors manually with/after plot1d?

AndreasAlbert:
@jrueb I'm currently doing [this](https://github.com/bu-cms/bucoffea/blob/master/bucoffea/plot/stack_plot.py#L182-L202), which works fine

PR #211, opened 2019-11-27
spark.read.parquet expects multiple args, not list

lgray:
Since we stopped reading parquet files in our test this interface wasn't checked.
It may be better to just use the `format` interface at the end of the day... I don't think there's any advantage of `spark.read.parquet` over `spark.read.format('parquet')`.

@perilousapricot ?


PerilousApricot:
Correct. parquet() is syntax sugar for format("parquet")
-- 
It's dark in this basement.

lgray:
That's very silly they don't preserve the expected argument structure from what it calls... sigh.

PR #212, opened 2019-12-03
Serialization improvements in for hist.Bin

lgray:
Code used to make the benchmark:
image

Histogram definition (filled with HZZ events from `http://scikit-hep.org/uproot/examples/HZZ.root`):
image

Original bins implementation on ~40000 dense bins:
image

without saving the Intervals list, and building it lazily:
image

This has significant benefits for running at scale!
@nsmith-

### DIFF
--- a/coffea/hist/hist_tools.py
+++ b/coffea/hist/hist_tools.py
@@ -438,18 +438,17 @@ class Bin(DenseAxis):
     """
     def __init__(self, name, label, n_or_arr, lo=None, hi=None):
         super(Bin, self).__init__(name, label)
+        self._lazy_intervals = None
         if isinstance(n_or_arr, (list, np.ndarray)):
             self._uniform = False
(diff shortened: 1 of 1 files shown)


nsmith- (approved):
I suppose a fast path for adding histograms is important in reduction stage, vs. re-assembling the full python objects.  A proper serialization format would sidestep it.

lgray:
Yeah a proper serialization format would be welcome... 

The summing operation itself doesn't take much time - ~150us for 40k bin histogram, so take what we can get here I guess.

nsmith-:
Proper serialization will come with boost-histogram. As for this case, it's a bit pathological to have 40000 bins in one axis. The complexity here is sum of nbins, not product, so really for most use it should be negligible.

lgray:
Eh, as people go to higher dim histograms we'll run into this one way or another. *shrug*

PR #215, opened 2019-12-07
Fix #214

### LINKED ISSUE #214: backward non-compatible change to hist.plot1d return value
**Describe the bug**
https://github.com/CoffeaTeam/coffea/blob/5eda3869f046a30dbebba96cf1fdaaf06a8e7e6d/coffea/hist/plot.py#L278
is not compatible anymore with the documentation
https://github.com/CoffeaTeam/coffea/blob/5eda3869f046a30dbebba96cf1fdaaf06a8e7e6d/coffea/hist/plot.py#L163-L168

and causes e.g. this nbook to fail
https://github.com/prasanthkothuri/swan-spark-notebooks/blob/master/muonspectrum_v3_spark.ipynb

was this a typo, or an intentional break of backward compatibility ?
If it was intentional, how can now the user code access the matplot FIG object to manipulate it ?

**To Reproduce**
self evident

**Expected behavior**
1. code matches documentation 
2. public interface of widely used classes like "plot" is stable

andrzejnovak:
Fixes #214

### DIFF
--- a/binder/muonspectrum_v2.ipynb
+++ b/binder/muonspectrum_v2.ipynb
@@ -82,7 +82,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "fig, ax, _ = hist.plot1d(masshist)\n",
+    "ax = hist.plot1d(masshist)\n",
     "ax.set_xscale('log')\n",
     "ax.set_yscale('log')\n",
     "ax.set_ylim(0.1, 1e6)"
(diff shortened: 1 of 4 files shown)


lgray:
@andrzejnovak you need to fix the test for plotgrid and then I think this is good to go. Thanks!

andrzejnovak:
@lgray Yup

PR #216, opened 2019-12-07
Pronunciation

andrzejnovak:
This the right one? 😅


lgray:
Yep, it's the right one :-)

Can you put the pronunciation centered below the name instead of to the right? Otherwise it could confuse people.

Also you can use [skip ci] at the front of the commit message when you're only editing text. :-)

andrzejnovak:
hmm @lgray can that be done in markdown?

andrzejnovak:
hmm @lgray can that be done in markdown?

lgray:
I have no idea!

andrzejnovak:
Ok, I give up RST, is retarded.

PR #217, opened 2019-12-10
add JER hybrid method

dnoonan08:
Fixes a bug in the JER stochastic smearing (missing a max(X,0) inside sqrt) and adds in the hybrid JER smearing:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetResolution#Smearing_procedures

Requires a column of ptGenJet in the jet to do the hybrid matching.

### DIFF
--- a/coffea/jetmet_tools/JetTransformer.py
+++ b/coffea/jetmet_tools/JetTransformer.py
@@ -104,7 +104,7 @@ class JetTransformer(object):
     def uncertainties(self):
         return self._junc.levels if self._junc is not None else []
 
-    def transform(self, jet, met=None):
+    def transform(self, jet, met=None, hybridJER=False):
         """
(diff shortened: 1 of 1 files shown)


lgray:
@dnoonan08 Thanks!

I have two comments:
1) You can reduce the total code size by getting rid of the if/else and just doing `doHybrid = (jet.ptGenJet.content > 0) & hybridJER` and then putting `doHybrid` in each of the `np.where` statements.
2) Make the logic so that if ptGenJet is provided as a column then the hybrid method becomes the default, if there is no ptGenJet attribute then it emits a warning and applies the stochastic smearing. Then change the flag to be `forceStochastic` instead of `doHybrid`. I think this is a bit closer to the JME tools, iirc.

lgray (approved):

lgray:
@dnoonan08 OK - this is now in coffea 0.6.22

PR #221, opened 2019-12-12
A full NanoAOD profile with methods

nsmith-:
This is an implementation of a similar concept to uproot-methods profiles, albeit with some more details.  It may well be appropriate to move to uproot-methods in its current state, but I plan to expand it to use some coffea addons eventually, such as JEC tools.

An example of instantiation:
```python
import uproot
from coffea.nanoaod import NanoEvents
tree = uproot.open('nanofile.root')['Events']
cache = {}
arrays = tree.lazyarrays(chunked=False, entrystart=0, entrystop=20000, cache=cache)
events = NanoEvents.from_arrays(arrays)
print(cache)
```
the empty `cache` dictionary shows the structure is formed without any branch reads.

Although not all features have been implemented, `events.GenPart` has both parent and children self-references, which are lazily computed with numba kernels taking about 2us per event.  For example, to select b quarks coming from a top decay where the W decayed to a muon and corresponding neutrino:
```python
gen = events.GenPart

where = (
    (abs(gen.pdgId) == 5)
    & (np.abs(gen.distinctParent.pdgId) == 6).fillna(False)
    & gen.hasFlags(['isHardProcess'])
)
genb = gen[where]

gentop = genb.distinctParent
hasWmu = (
    (abs(gentop.children.pdgId) == 24)
    & (abs(gentop.children.children.pdgId) == 13).any()
).any()

chosenb = genb[hasWmu]
```
In fact it would be more straightforward to start from the top quark, but this method demonstrates the up- and down-reference.  Note that we can't directly select the W via `gentop.children[gentop.children.pdgId == 24]` as it is a double-jagged index. This will be possible in awkward 1.x.  This whole selection takes about 7us/evt on a semileptonic ttbar sample I'm testing.


lgray:
A structural comment to start off with:

The common vectors and candidates and the GenParticles stuff is really great, thank you for adding it. They're fairly general but currently displayed as CMS-specific stuff.
 
The new vectors and candidate are basic concepts, and the gen-particle concept is at least common across experiments. 

The schema of gen particle tables is also not so experiment dependent, and it's easy to define a consistent interface. I think writing how the gen particle table works can help in further discussion to make a more general tool, it would be useful as a docstring.

Given that, can you relocate `methods.common`, `methods.generator`, and the `_mixin` in `nanoaod.util` over to `analysis_objects`?

Similarly, can you make the gen particle FLAGS a parameter of the GenParticle constructor, defaulted to the NanoAOD flags?

nsmith-:
> The schema of gen particle tables is also not so experiment dependent, and it's easy to define a consistent interface. I think writing how the gen particle table works can help in further discussion to make a more general tool, it would be useful as a docstring.

I'm not sure what you mean by 'how it works'  I need to put docstrings for the methods still, but beyond those what would that entail?

> Given that, can you relocate methods.common, methods.generator, and the _mixin in nanoaod.util over to analysis_objects?

I want to leave this self-contained for now, at least until awkward1 transition.  There will likely be some significant refactor in concurrence with that.

> Similarly, can you make the gen particle FLAGS a parameter of the GenParticle constructor, defaulted to the NanoAOD flags?

These are mixins, they cannot have a constructor since they are mixed into multiple awkward types, and anyway instance properties will be lost on awkward operations.  This might be possible in awkward 1, though.

lgray:
How it works -> I mean write down the specification of the how gen information is encoded in columns (and what you're doing to decode and re-encode it).

I'm fine with waiting until awkward1 on the other stuff.

nsmith-:
This feature can be enabled in `run_uproot_job` by setting `nano=True` in the executor arguments, after which the usual `df` passed into coffea processor becomes a `NanoEvents` object.  The dataset name is moved then to `df.metadata['dataset']`.

Some last notes: the NanoEvents `from_arrays` can be used with a preloaded dataframe in spark.  If `NanoEvents` is constructed from a file (with `cache=None`) the `NanoEvents.materialized` property lists the columns that were accessed, which can be used to fill a preloaded dataframe for later use.

PR #226, opened 2019-12-14
Make dask get correct name for functions

nsmith-:
The display name of a function in the dask webui is [eventually taken from the](https://github.com/dask/dask/blob/master/dask/utils.py#L695-L724) `__str__` method if it has no `__name__`.


lgray:
I'm guessing a proper release can wait for this one?

PR #228, opened 2019-12-17
Remove python2 tests and warn if in python2

lgray:
To merge on 1 Jan 2020.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 os:
 - linux
 python:
-- 2.7
 - 3.6
 - 3.7
 env: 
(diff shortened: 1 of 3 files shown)


andrzejnovak:
is uproot dropping py3?

lgray:
@jpivarski ?

jpivarski:
I'll be supporting Python 2.7 as long as it's easily possible (i.e. we can use conda to install packages). There's a Scikit-HEP declaration that we don't have to, but I will. That doesn't mean you have to.

nsmith-:
I think this might be a bit too restrictive.  We have a large code base that is py2 compatible at the moment.  Maybe we allow it but no longer test.  If a user tries to run it, one could e.g. emit a warning and if something breaks perhaps they will make a PR with the necessary changes.

lgray:
Sure, that's reasonable. Wasn't sure how ... provocative... we wanted to be.

PR #229, opened 2019-12-20
Enforce more modern dependencies

nsmith-:
Since many users only bother with `pip install --upgrade coffea` (if that!) then they don't necessarily receive useful updates to the uproot/uproot-methods/awkward stack.  This updates the stack to the latest and greatest as of now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,13 +56,13 @@ def get_description():
     return description[start:stop].strip()  # before + + after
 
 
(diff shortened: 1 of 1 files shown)


lgray:
Since I don't think anyone uses the old awkward numba stuff (which requires a particular and old version of awkward) I think this is fine.

Similarly awkward 1 is around the corner and all this is gonna change quite a bit.

PR #230, opened 2019-12-20
Start integrating github actions.

lgray:
I hope I can run this as a PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,12 +6,12 @@ on:
       - gha_master
     # tags:
     #   - /v(\d+)?\.(\d+)\.(\d+)/
-  pull_request:
-    branches:
-      - master
+  # pull_request:
+  #   branches:
+  #     - master
   # Run daily at 0:01 UTC
-  schedule:
-    - cron:  '1 0 * * *'
+  # schedule:
+  #   - cron:  '1 0 * * *'
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,9 @@ name: CI/CD
 on:
   push:
     branches:
-      - master
-    tags:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools setuptools_scm wheel
+        python -m pip install codecov
+        python -m pip -q install numpy packaging scipy numba tqdm matplotlib pytest pytest-runner flake8 mplhep --upgrade
+        python -m pip -q install pandas pyarrow cloudpickle lz4 jinja2 pyspark dask distributed --upgrade
+        python -m pip -q install https://github.com/Parsl/parsl/zipball/master
+        python -m pip install ${{ matrix.awkward-version }}
+        python -m pip install uproot
+        python -m pip install -q --no-cache-dir -e .
```
nsmith-: Except for tests-requires, all the rest should be installable via just `python -m pip install -e .[spark,parsl,dask]` then also it obeys the `setup.py` requirements.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+        python -m pip list
+    - name: Lint with flake8
+      run: |
+        python setup.py flake8
+    - name: Test with pytest
+      run: |
+        python setup.py pytest
```
nsmith-, line 62: If `pip install -e` is used, plain `pytest` and `flake8 coffea` can be used, which avoids some weirdness that I've seen happen with setup.py wrapping.

.github/workflows/ci.yml
nsmith-: For purposely downgraded options (e.g. awkward verison), running pip install on that first, then this, makes sure that a) the downgraded version is permitted by `setup.py` and b) that it doesn't get force-upgraded.

PR #239, opened 2020-01-02
Add a tail timeout option to executors

dntaylor:
There is a feature that sometimes a job can get stuck running (testing on condor with parsl) that prevents the processor from proceeding to postprocess.

This adds an option to executor_args (and pre_args) to pass `tailtimeout=N` to timeout all remaining futures if more than a specified amount of time has passed since the last completed job.

I'm also going to prepare a PR for parsl that allows downscaling of condor resources.

The point of this is to prevent wasting resources on long running tails.

There is another change added also, the `_cancel` function to wrap `job.cancel()`, which is not implemented in parsl `AppFuture`. It doesn't really matter if we don't cancel the job, since the dfk cleanup will take care of it later.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -169,7 +169,7 @@ def _futures_handler(futures_set, output, status, unit, desc, add_fn, tailtimeou
                     pbar.update(1)
                     last_job = time.time()
                 time.sleep(0.5)
-                if tailtimeout > 0 and (time.time() - last_job) > tailtimeout and (last_job - start) > 0:
+                if tailtimeout is not None and (time.time() - last_job) > tailtimeout and (last_job - start) > 0:
                     for job in futures_set:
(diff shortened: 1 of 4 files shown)


lgray:
For spark just make it not have a tail timeout, it has its own internal mechanisms for this.

dntaylor:
Will do.
Question, how do you run the tests locally? I'm doing:
`python -m test --testdir ./tests/`
but it fails on
`FileNotFoundError: [Errno 2] No such file or directory: 'tests/samples/Summer16_23Sep2016V3_MC_L1FastJet_AK4PFPuppi.jec.txt.gz'`

lgray:
`python setup.py pytest`

lgray:
Let me know if you're done with this and I'll merge.

dntaylor:
Yes, done.

PR #246, opened 2020-01-10
Add support for rochester lookup table

dntaylor:
This PR adds support for a Rochester correction workflow.

The `loaduncs` flag optionally loads uncertainties in case you don't need the uncertainty and want to save memory.

The interface it intended to exactly match the officially support Rochester C++ class.
A usage example is:

```
    tag = 'roccor.Run2.v3'
    year = '2018'
    rochester_data = lookup_tools.txt_converters.convert_rochester_file(f'tests/samples/rochester/{tag}/RoccoR{year}.txt',loaduncs=True)
    rochester = lookup_tools.rochester_lookup.rochester_lookup(rochester_data)

    # when using data
    events = NanoEvents.from_file(os.path.abspath('tests/samples/nano_dimuon.root'))

    data_k = rochester.kScaleDT(events.Muon.charge, events.Muon.pt, events.Muon.eta, events.Muon.phi)

    # when using mc
    events = NanoEvents.from_file(os.path.abspath('tests/samples/nano_dy.root'))
    mc_rand = np.random.rand(*events.Muon.pt.flatten().shape)
    mc_rand = JaggedArray.fromoffsets(events.Muon.pt.offsets, mc_rand)
    hasgen = ~np.isnan(events.Muon.matched_gen.pt.fillna(np.nan))
    mc_kspread = rochester.kSpreadMC(events.Muon.charge[hasgen], events.Muon.pt[hasgen], events.Muon.eta[hasgen], events.Muon.phi[hasgen],
                                     events.Muon.matched_gen.pt[hasgen])
    mc_rand = JaggedArray.fromoffsets(hasgen.offsets, mc_rand)
    mc_ksmear = rochester.kSmearMC(events.Muon.charge[~hasgen], events.Muon.pt[~hasgen], events.Muon.eta[~hasgen], events.Muon.phi[~hasgen],
                                   events.Muon.nTrackerLayers[~hasgen], mc_rand[~hasgen])
    mc_k = np.ones_like(events.Muon.pt.flatten())
    mc_k[hasgen.flatten()] = mc_kspread.flatten()
    mc_k[~hasgen.flatten()] = mc_ksmear.flatten()
```

Importantly, Rochester corrections use a symmetric double-sided crystal ball function.
This is not in scipy.stats, so I manually created it (well, the fully asymmetric version).
I will make a PR to `scipy` to have it there in the future.


dntaylor:
FYI: I did some validation checks to make sure this double-sided crystal ball implementation matches the official Rochester one. Agreement is within `1e-7`.

cb_invcdf
cb_pdf
cb_cdf

lgray:
@dntaylor Thanks for adding this!

Can you add some execution time benchmarks (how many kHz/core to evaluate, etc.)?  

There are a few places where it's clear you could replace divisions by taking 1/scalar and multiplying.

Can you please add a test for this function? (one stemming from your validation plots would be nice!)

Only the test is really required from my point of view, the rest would be nice though.

lgray:
Ah, and looking at how you read in the files.
If you can condense that into a (few) np.loadtxt statement(s), that would be useful too.

dntaylor:
I didn't think of a good way to do that since the number of columns in a row are irregular with the first few columns defining how many elements are in that row.

lgray:
- I've also tried to consistently support gzipped text-based correction files since the compression ratio is typically enormous and smaller payloads are better.  

- I've also tried to keep the parsing implementation separate from the evaluation implementation. This helps to reduce the size of code that needs to be delivered to workers. Please pull those apart in the rochester code!

I think these are the only other things I'll request as needed along with a test.

You can take a look at the jetmet or b-tagging parsers for examples.

dntaylor:
Can do.

Some quick benchmarks with the current commit:

```
isData True doError False
50000 event 61115 muons
0.03323514740914106 s/iteration

isData True doError True
50000 event 61115 muons
5.3006772212684155 s/iteration

isData False doError False
50000 event 32341 muons
0.21880943272262812 s/iteration

isData False doError True
50000 event 32341 muons
32.73869079782627 s/iteration
```

Which works out to:
Data (w/ error): 1840 (11.5) khz/core (muons)
MC (w/ error): 148 (1.0) khz/core (muons)

dntaylor:
@lgray 
Wanted to get your opinion on this approach.

What I currently do:
Single class that parses the txt file on init and passes the class to the node.

What I think you are suggesting and I will implement:
Create a `rochester_extractor` class that takes a Rochester txt file and returns the needed lookup tables (there are a lot, minimum (w/o error) 12 and up to ~110 * 12 (w error)).
This would then be what is shipped to the node.
Create a `rochester_evaluator` class that takes the lookup tables at runtime and evaluates the scale factor.

Alternatively, you could still split the functions up but just send the txt file to the node and construct the rochester extractor and evaluator each time (~0.5s).

lgray:
@dntaylor Send one huge batch of already processed numpy arrays (i.e. some big tensor) and make the evaluator around that. Numpy arrays serialize pretty well.

lgray:
I'm a bit worried by the performance of `isData False doError True`, that's a big bottleneck. 

We can work on performance after we get this merged.

If you wanna do some profiles to see if anything obvious sticks out, try using https://github.com/joerick/pyinstrument. I've found it to be a very useful tool in python.

dntaylor:
The reason for 100x increase is because you have to do the computation 100 times for that stat uncertainty using the current recommendation. (The syst uncertainties are more reasonable).

lgray:
Yeah, I got that part but if there are hotspots in the code being called 100x maybe we can whittle on them a bit. This can be done later, it's OK. :-)

dntaylor:
@lgray 
How did you add the nanoaod samples in the tests directory?
Specifically, can they be updated to a newer version of NanoAOD?
https://github.com/CoffeaTeam/coffea/blob/master/tests/samples/nano_dy.root
Is missing the Muon_genPartIdx branch which I need for the rochester test validation.

lgray:
@dntaylor We skimmed off the first 10 events of a drell yan sample and I think double muon data, and just tossed it in there with git add. 

I don't think it's too much effort to make your own, and update what we have there.
If you need us to update them, we can. Just let us know which files.

It will require adjusting of some of the tests if the datasets have different GEN.

dntaylor:
I can do it, just wanted to make sure it wasn't something special. I will grab from the latest NanoAODv6 campaign.

lgray:
Cool, thanks!

dntaylor:
@lgray Okay, you can take a look and see if this is more consistent with what you were thinking.
I've updated the tests with new root files.
I have not updated .parquet files to match.

lgray:
Yep this looks good! Only remaining thing is being able to read gzipped rochester corrections files.
2MB per file is a bit heavy, and I'm sure it'll compress very well.

If you can remove the C++ code from the test/samples directory that would be nice too.

Good solid test, thanks a lot.

dntaylor:
Turns out I need to fix something. Forgot to add an error test and it is broken.

nsmith-:
@jpata may be interested in this

dntaylor:
@lgray Okay, gzip support added. Any last changes you would like to see?

lgray:
I think that's it for now! Thanks a lot!

Issue #248, opened 2020-01-14
tailtimeout default is not None in pre-executor


nsmith-:
Based on the line https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L638 it appears that all pre-executor jobs will timeout.  Should this be `None`?

A workaround would then be
```python
run_uproot_job(..., pre_args={'tailtimeout': None})
```

While here, the line https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L592 is inaccurate, as `tailtimeout` is normally passed on to executors and not interpreted by `run_uproot_job` itself.

nsmith-:
Also, the warning `Stopped job early` might be clarified to suggest it is related to the `tailtimeout` setting

nsmith-:
@dntaylor

dntaylor:
Ah, sorry about that. Yes, it should be `None` by default. I stopped testing things without the tailtimeout set to a value so missed this.

nsmith-:
Fixed in #250 
Can we cut a release @lgray ?

lgray:
Please do!

PR #250, opened 2020-01-15
Fix tailtimeout default for preargs

dntaylor:
This fixes issue #248.


lgray:
@dntaylor Can you add a bit more description to the exception this raises when you time out?

Issue #222, opened 2019-12-12
plot1d to automatically pick overlay if only one categorical and one dense axis present


andrzejnovak:

nsmith-:
What if one wants a multi-bar graph per category, where each bar color is a dense bin?

nsmith-:
Note that is also not implemented yet, but could be.
e.g. https://matplotlib.org/3.1.1/gallery/lines_bars_and_markers/barchart.html

andrzejnovak:
I guess that should be a different function

andrzejnovak:
hmm or a flag for sth like `overlay_method` ?

PR #255, opened 2020-01-28
Auto overlay for 1dense1sparse

andrzejnovak:
Adding  #222


nsmith- (approved):
Ok, still leaves room for having a category x axis if overlay is not None.

lgray:
Ready for merge or do we want anything else here?

PR #261, opened 2020-01-30
Dask improvements

nsmith-:
Introduce an opt-in caching layer with worker affinity for jobs, automatic retries, and some small usability things.


nsmith-:
In the interest of not forking too much, I'm going to merge it as-is.  No changes to user interface are expected, although some new opt-in arguments exist now.  Critically, dask retries jobs by default now.

PR #263, opened 2020-02-06
add automatic test to see btag CSV reader failure

irenedutta23:
**This PR is for reporting a bug.** 
I noticed that the B tag evaluator crashes while trying to implement the new TuneCP5 btag shape scale factors (DeepCSV_2016LegacySF_V1_TuneCP5.csv) from here:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/BtagRecommendation2016Legacy

It also fails while trying to evaluate the new 2017 V5 scale factors (DeepCSV_94XSF_V5_B_F.csv) from here:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/BtagRecommendation94X

This is a small snippet of code that reproduces this problem. Perhaps someone can find a bug fix for this.
Thanks in advance for looking into this!


lgray:
Thanks for catching this, I'll take a look into the crash! 

The b-tag formats seem to have been in flux recently so not surprising.

Can you gzip the csv file before we add it to our test repository though (preferably with a rebase).

irenedutta23:
I have added this gzip file here:
tests/samples/DeepCSV_2016LegacySF_V1_TuneCP5.btag.csv.gz

Do you need some other file also ?

lgray:
Please remove the non-gzipped file :-)

lgray:
Can you squash that down into one commit with rebase? Otherwise the full file sticks around in the git history.

Thanks!

nsmith-:
I also noticed this for 2017 DeepCSV scale factors, and wrote the following parser using pandas:
https://github.com/nsmith-/boostedhiggs/blob/dev/boostedhiggs/btag.py#L70
I will eventually move it into coffea after testing other years.

lgray:
@nsmith- Nice, looking forward to its addition.

lgray:
@irenedutta23 @nsmith- So the underlying problem is that the binning for the discriminator changes as a function of eta and pt bin. This was not the case up until these files, and so was assumed as a property of the file format.

I'll figure out a fix, or Nick may submit his version of this in the mean time, whichever comes first!

nsmith-:
@irenedutta23 I've merged the file you added into #267 
Let me know if that looks like it will work for you

irenedutta23:
@nsmith- I think this works for me! Thanks! Just for clarification : For the old evaluator, I had to pass `abs(eta)` to get the right SF. Is this also the case with the new ` BTagScaleFactor` class ? I see that may not be the case here: [validation/btag/validate.py R52](https://github.com/CoffeaTeam/coffea/pull/267/files/4616ddea06fbbf8d98474d1f4b76424ac71101d5..0c3987bc4d9f16133b4eb344bba5a00750c67e8d#diff-038dbf231b88f3daaf4b484aa6cded60R52)

andrzejnovak:
iirc the SFs are always using abs eta, so it would possibly make sense to hardcode that

nsmith-:
It appeared to me that the intention was for the BTV class to support signed eta, but then no CSV file I've seen so far has signed eta SFs.  I left it in for convenience, and put abs(eta) for those SF where no negative eta bin is found, see https://github.com/CoffeaTeam/coffea/blob/btagsf/coffea/btag_tools/btagscalefactor.py#L110-L113 
But if BTV has no plans to use signed eta, then yes it would make sense to hardcode an abs or require the use to do so.

andrzejnovak:
Yeah, i don't see how scale factors should be signed eta dependent. But if you want to future proof you could add a signedEta flag.

nsmith-:
The example file was included in the tests of #267

PR #267, opened 2020-02-14
Introduce dedicated b-tagging scale factor class

nsmith-:
This class simplifies implementation of b-tagging scale factors by allowing a single lookup on (flavor, eta, pt, discriminant) column vectors.  It also is tested to load and work on all currently available csv files that I could find on the TWiki.
What has not yet been tested is a validation against the non-column tool at https://github.com/cms-sw/cmssw/tree/CMSSW_10_0_X/CondTools/BTau/test


nsmith-:
Some basic validation against the BTV class shows pretty decent agreement for one particular SF set.

Edit: synced to float precision for appropriate input bounds.  Seems the eval_auto_bounds defaults to 1. when `abs(eta)>2.5` or for some cases also when `eta<0`

Issue #272, opened 2020-02-18
plot1d(stack/density)


andrzejnovak:
When both density and stack are `True`, `plot1d` currently does density and then stack. Should also have an option to do stack first and then density.

PR #273, opened 2020-02-20
feat: plot stack behaviour, externalize some code to mplhep

andrzejnovak:
mplhep now enables two types of interaction between stack/density options, so let coffea use that


andrzejnovak:
Also closes #272

andrzejnovak:
Ready to merge

PR #276, opened 2020-03-11
Fix Cornercases: Empty formula and multiple formulas for btag SF

pfackeldey:
Dear coffea developers,

this PR implements fixes for two corner cases:

1. set the default scale factor formula to "1." in case the `this_bin` mask is empty
2. raise an Exception in case multiple formulas are found for the same bin

Thanks a lot in advance for reviewing!

PS:
I stumbled over the first corner case with the current deep flavour btag csv file in CMS (2017), where i just got an index out of bound error, since the `this_bin` mask was an empty array.

### DIFF
--- a/coffea/lookup_tools/csv_converters.py
+++ b/coffea/lookup_tools/csv_converters.py
@@ -58,11 +58,26 @@ def convert_btag_csv_file(csvFilePath):
         for i, eta_bin in enumerate(etaBins[:-1]):
             for j, pt_bin in enumerate(ptBins[:-1]):
                 for k, discr_bin in enumerate(discrBins[:-1]):
-                    this_bin = np.where((all_names == label) &
-                                        (corrections[columns[4]] == eta_bin) &
-                                        (corrections[columns[6]] == pt_bin) &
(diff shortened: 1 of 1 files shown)


nsmith-:
Hi, thanks for solving the parsing error!
You might be interested in looking at the revised b-tagging class implemented in #267 and documented in https://coffeateam.github.io/coffea/api/coffea.btag_tools.BTagScaleFactor.html#coffea.btag_tools.BTagScaleFactor

pfackeldey:
Cool! Looks like I can also use this `BTagScaleFactor` class to extract the scale factors. Is this way recommended in comparison to using the `extractor` from `lookup_tools` ? 
(btw i fixed the flake8 linting)

nsmith-:
You might find it easier to use, since it handles apply a SF to all flavors at once along the lines of the BTV class `eval_auto_bounds`

pfackeldey:
Thanks a lot! I will try it out. :)

lgray:
In the mean time, I'm going to merge this. Thanks a lot for the contribution!

Good to keep the older stuff running too, but I'll add a deprecation warning to the old btagging SF classes.

PR #277, opened 2020-03-12
Revert "Fix Cornercases: Empty formula and multiple formulas for btag SF"

lgray:
Reverts CoffeaTeam/coffea#276

The fix causes another bug.

### DIFF
--- a/coffea/lookup_tools/csv_converters.py
+++ b/coffea/lookup_tools/csv_converters.py
@@ -61,21 +61,8 @@ def convert_btag_csv_file(csvFilePath):
                     this_bin = np.where((all_names == label) &
                                         (corrections[columns[4]] == eta_bin) &
                                         (corrections[columns[6]] == pt_bin) &
-                                        (corrections[columns[8]] == discr_bin))[0]
-                    if len(this_bin) == 1:
-                        vals[k, j, i] = corrections[this_bin][columns[10]][0]
-                    elif len(this_bin) > 1:
-                        raise Exception(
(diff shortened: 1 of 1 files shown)


lgray:
@pfackeldey The fix you provided stalls on the Legacy btag SFs.

Can you please provide a functioning test of the patch?

pfackeldey:
@lgray Could you point me to the Legacy btag SFs? I will try then to provide a functioning test of the patch. Thank you :)

lgray:
@pfackeldey Try out https://github.com/CoffeaTeam/coffea/blob/master/tests/samples/DeepCSV_2016LegacySF_V1_TuneCP5.btag.csv.gz

Thanks!

pfackeldey:
@lgray For me it works for the legacy btag SFs

```python
from coffea.lookup_tools import csv_converters
sfs=csv_converters.convert_btag_csv_file("tests/samples/DeepCSV_2016LegacySF_V1_TuneCP5.btag.csv.gz")
```
works for me (just takes a few minutes)

lgray:
OK - then it was so slow I gave up. If it doesn't parse in a few seconds there's pretty much no reason to use this over the Pandas implementation. I guess that's just a huge file?

pfackeldey:
Yes exactly. That was why I "precompiled" those corrections and saved the evaluator in a `.coffea` file in my analysis. But I also switched to the Pandas implementation, which is way faster.

lgray:
Yeesh, someone felt the kitchen sink was needed.
Very well. 

I'll instead hook Nick's stuff into the evaluator / extractor schema, add a deprecation warning for this.

Thanks!

PR #286, opened 2020-03-27
small processor usability improvement

ghost:
This PR introduces few features:
   * more printout in case of bad files
   * the current filename is passed to the LazyDataFrame as an additional field, useful for debugging rare crashes
   * ```run_uproot_job``` can run on something other than TTrees. In that case, in the  preprocessing the number of entries per file is set to 1, assuming constant execution time and instead of a LazyDataFrame, a dictionary with a key ```'obj'``` is passed to the processor


lgray:
@jpriscia please fix the following linter issues
```
coffea/processor/executor.py:522:22: E261 at least two spaces before inline comment
coffea/processor/executor.py:524:30: E203 whitespace before ':'
coffea/processor/executor.py:525:34: E203 whitespace before ':'
coffea/processor/executor.py:526:35: E203 whitespace before ':'
coffea/processor/executor.py:606:65: E261 at least two spaces before inline comment
coffea/processor/executor.py:606:110: W291 trailing whitespace
```

You can reproduce this list locally by running `python setup.py flake8` in the root directory of coffea.

lgray:
@nsmith- you may want to check this out to see if you're happy with it, or if we want to do this in a more clean way.

nsmith-:
Why not utilize the executors directly? The main functionality provided by `run_uproot_job` is doing the splitting into even-sized tasks. For a one-per-file task, the splitting is trivial and you can just pass a list of files to a user-defined function (that possibly wraps `_work_function`) to do the computation.  The executors provide the map-reduce mechanism.

nsmith-:
Do you want to still pursue this PR or did you find a good solution with using the executors directly?

nsmith-:
Closing due to inactivity

PR #290, opened 2020-04-03
fix doc string for candidatesfromoffsets

jmduarte:
fix doc string for JaggedCandidateArray.candidatesfromoffsets


lgray:
thanks for the catch!

PR #293, opened 2020-04-08
speed up btag SF initialization

jpata:
Improves the initialization time of BTagScaleFactor for the RESHAPE option (4min -> 8s for one file).

Seems like `numba.vectorize` is rather slow in compilation time and simple `numba.jit` works as well, if we take care of masks, thanks @nsmith- for the tip.

### DIFF
--- a/coffea/btag_tools/btagscalefactor.py
+++ b/coffea/btag_tools/btagscalefactor.py
@@ -135,16 +135,9 @@ class BTagScaleFactor:
             except KeyError:
                 if 'x' in formula:
                     feval = eval('lambda x: ' + formula, {'log': numpy.log, 'sqrt': numpy.sqrt})
-                    out = numba.vectorize([
-                        numba.float32(numba.float32),
-                        numba.float64(numba.float64),
-                    ])(feval)
+                    out = numba.jit()(feval)
                 else:
-                    val = eval(formula, {'log': numpy.log, 'sqrt': numpy.sqrt})
-
-                    def duck(_, out, where):
-                        out[where] = val
-                    out = duck
+                    out = eval(formula)
                 BTagScaleFactor._formulaCache[formula] = out
                 return out
 
@@ -187,6 +180,7 @@ class BTagScaleFactor:
         except KeyError:
             functions = [BTagScaleFactor._compile(f) for f in self._corrections[systematic][-1]]
             self._compiled[systematic] = functions
+
         try:
             flavor.counts
             jin, flavor = flavor, flavor.flatten()
@@ -212,7 +206,13 @@ class BTagScaleFactor:
                 var = numpy.clip(discr, corr[2][0], corr[2][-1])
             else:
                 var = numpy.clip(pt, corr[1][0], corr[1][-1])
-            func(var, out=out, where=(mapidx == ifunc))
+            
+            where=(mapidx == ifunc)
+            if isinstance(func, float):
+                out[where] = func
+            else:
+                tmp = func(var)
+                out[where] = tmp[where]
 
         if jin is not None:
             out = jin.copy(content=out)
--- a/coffea/btag_tools/btagscalefactor.py
+++ b/coffea/btag_tools/btagscalefactor.py
@@ -210,8 +210,8 @@ class BTagScaleFactor:
             if isinstance(func, float):
                 out[where] = func
             else:
-                tmp = func(var)
-                out[where] = tmp[where]
+                tmp = func(var[where])
+                out[where] = tmp
 
         if jin is not None:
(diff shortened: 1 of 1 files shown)


lgray:
Nice optimization!

coffea/btag_tools/btagscalefactor.py
```diff
@@ -212,7 +206,12 @@ def eval(self, systematic, flavor, abseta, pt, discr=None, ignore_missing=False)
                 var = numpy.clip(discr, corr[2][0], corr[2][-1])
             else:
                 var = numpy.clip(pt, corr[1][0], corr[1][-1])
-            func(var, out=out, where=(mapidx == ifunc))
+            where = (mapidx == ifunc)
+            if isinstance(func, float):
+                out[where] = func
+            else:
+                tmp = func(var)
```
nsmith-: How about `tmp = func(var[where])` and `out[where] = tmp` ? Saves evaluation on unnecessary inputs

nsmith- (changes requested):

coffea/btag_tools/btagscalefactor.py
jpata: sure. I didn't find a speed difference in a test on 1M random inputs, but might be nicer.

nsmith-: Ha!  Goes to show how assumptions about performance are often wrong.. turns out we can do extra math about as fast as allocating a smaller array
Feel free to change or leave it if you tested.

lgray:
@nsmith- can you approve? or does it need more review?

nsmith- (approved):

nsmith-:
Failed inb4

jpata:
Folks, I did some more testing in the actual analysis, and I think this might have to be reverted, as the overall analysis runtime is slower... 😞😔Now I timed it in two steps.

Load CSV, run  `eval` with an array of length one for all systematics to initiate numba JIT:
- `numba.vectorize`: ~250 s (could be reduced by 50% by restricting signatures to e.g. float32 only)
- `numba.jit`: ~9 s
- BTagCalibrationStandalone via Python CFFI: ~95 s

Evaluation on 1.3M jets for one systematic:
- `numba.vectorize`: ~0.9 s
- `numba.jit`: ~16 s
- BTagCalibrationStandalone via Python CFFI: ~0.8 s

In the original test above, I ran it just once with a large random input array, such that the init+eval times were not measured separately.

Systematics: `["jes", "lfstats1", "lfstats2", "hfstats1", "hfstats2", "cferr1", "cferr2", "lf", "hf"]`
SF file: `DeepCSV_102XSF_V1.csv`

nsmith-:
hmm are we sure things like `numba.jit()(lambda x: x**2 + 3)` are actually doing what we want on numpy arrays? Can you try `numba.njit()` ?

nsmith-:
I expected the `jit` function to require a manual loop inside

nsmith-:
ok nevermind looks fine
```
In [5]: f = numba.jit()(lambda x: numpy.log(x)**2 + 3)

In [6]: f(x)
Out[6]:
array([8.17241946, 8.62260485, 8.51762165, ..., 8.42205341, 8.14520123,
       7.49257713])

In [7]: %timeit f(x)
455 µs ± 4.14 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
 vs.

```
In [10]: def doit(x):
    ...:     out = numpy.zeros_like(x)
    ...:     for i in range(len(x)):
    ...:         out[i] = numpy.log(x[i])**2 + 3
    ...:     return out
    ...:

In [11]: f = numba.njit()(doit)

In [12]: f(x)
Out[12]:
array([8.17241946, 8.62260485, 8.51762165, ..., 8.42205341, 8.14520123,
       7.49257713])

In [13]: %timeit f(x)
455 µs ± 4.03 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

jpata:
Yeah I think in recent numba versions `jit==njit` basically, object mode has to be turned on by by a flag. Maybe what's happening is that it's using python list comprehension to produce an array? Or the function is getting compiled every time it's called, not sure...

nsmith-:
hmm I'm testing on 100k events and not seeing quite the same, with your latest change:
```
load time: 0.5287809371948242
first eval time: 7.369645833969116

In [1]: %timeit sf.eval('central', jet.hadronFlavour, jet.eta, jet.pt, jet.btagDeepB, ignore_missing=True)
249 ms ± 1.75 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I played a bit and got it down to 117 ms, but not really much improvement.

PR #300, opened 2020-04-27
Fix recursion error in LazyDataFrame

### LINKED ISSUE #299: RecursionError when copying LazyDataFrame
**Describe the bug**
When one tries to create a shallow copy a `coffea.processor.LazyDataFrame` with Python's standard `copy.copy` one gets a RecursionError.

**To Reproduce**
```
import uproot
import coffea
import copy
f = uproot.recreate("/tmp/roottest.root")
f["t"] = uproot.newtree({"branch": "int32"})
df = coffea.processor.LazyDataFrame(f["t"])
copy.copy(df)
```

**Additional context**
One can more directly trigger the error by
```
coffea.processor.LazyDataFrame.__new__(coffea.processor.LazyDataFrame).something
```
 `copy` seems to be calling the built-in `__new__` for `LazyDataFrame` which creates an instance without calling `__init__`. The error then seems to be happening because `LazyDataFrame.__getattr__` forwards to `LazyDataFrame.__getitem__`, which in turn needs attributes such as `_dict`, which causes another `__getattr__` call.
Maybe it would be better to have `__getitem__` forward to `__getattr__` as latter is more fundamental. Although preferably I think there should only be one way (see the Zen of Python) to access the items of a `LazyDataFrame`, thus getting rid of the `LazyDataFrame.__getattr__` definition altogether would solve everything.

nsmith-:
Encountered when python attempts dunder methods on a __new__-initialized
instance, e.g. during copy.copy() calls.
Fixes #299

### DIFF
--- a/coffea/processor/dataframe.py
+++ b/coffea/processor/dataframe.py
@@ -9,7 +9,9 @@ except ImportError:
 class LazyDataFrame(MutableMapping):
     """Simple delayed uproot reader (a la lazyarrays)
 
-    Keeps track of values accessed, for later parsing.
+    One can access branches either through ``df["bname"]`` or ``df.bname``, although
+    the latter is restricted to branches that do not start with a leading underscore.
+    Keeps track of values accessed, in the `materialized` attribute.
 
     Parameters
     ----------
@@ -52,6 +54,8 @@ class LazyDataFrame(MutableMapping):
             raise KeyError(key)
 
     def __getattr__(self, key):
+        if key.startswith("_"):
+            raise AttributeError(key)
         try:
             return self.__getitem__(key)
         except KeyError:
@@ -129,6 +133,8 @@ class PreloadedDataFrame(MutableMapping):
         return self._dict[key]
 
     def __getattr__(self, key):
+        if key.startswith("_"):
+            raise AttributeError(key)
         try:
             return self.__getitem__(key)
         except KeyError:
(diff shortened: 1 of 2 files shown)


lgray:
Are we sure we want to do it like this? What if someone's ntuples have leading '_' in branch names?

nsmith-:
They can use getitem then. I left a note in the docstring to that effect as well.

nsmith-:
The alternative is playing dunder method whack-a-mole forever

lgray:
That's fine, it's a rare case to begin with.

Why not change the `.startswith("_")` to `.startswith("__")`, which will be even more rare and should still solve the problem?

nsmith-:
Actually that would not solve this problem. The minimal restriction would be `key.startswith("__") or key in ["_dict", "_tree", "_materialized", "_branchargs"]` as those are each attribute accessed in the `__getitem__` call. Seems to me more straightforward to ban all leading-underscore than have magic restrictions the user may have a harder time diagnosing.

nsmith-:
Well I take that back, banning dunder would work, but then all branch names named `_*` _except_ `_tree`, etc. would be valid. Again, I think blanket ban over magic names seems a better policy. As always, `df["_mybranch"]` would work.

lgray:
OK - I think this will be fine for now. 

I wanna poke around with overloading `__copy__()` and `__deepcopy__()` to see if the restriction can just be removed. If it can I'll post a PR.

nsmith-:
Well the private members (`tree_`, etc.) would always be restricted even then, which is my main point.
Might want to do `__setstate__` etc. as well.

PR #304, opened 2020-05-26
Work Queue Executor: global queue

keylime-unicorn:
Implemented a global queue to reduce data transfer and improve overall performance of Coffea when running with Work Queue.


lgray:
@camicarballo Would you mind coming to the next coffea dev meeting and talking about how the performance improvements you implemented here work?

Otherwise code changes look fine. Thanks!

keylime-unicorn:
@lgray I can definitely talk about the changes at the next meeting :)

PR #305, opened 2020-06-04
feat: Add gen matching to nanoevents for fatjets

andrzejnovak:
Self explanatory, hopefully it will be followed up by matching for candidates


andrzejnovak:
@nsmith- This is ready.

coffea/nanoaod/methods/jets.py
```diff
@@ -4,6 +4,7 @@
 
 class FatJet(LorentzVector):
     '''NanoAOD large radius jet object'''
+    _enable_genjet = False
     subjetmap = {'FatJet': 'SubJet'}  # V6 has 'GenJetAK8': 'SubGenJetAK8', maybe better to put in generator.py
```
nsmith-, line 8: Ah I think here is why I had set `GenJetAK8` to map to FatJet class: since you can link the subjets. But as the comment says, this probably makes more sense as a specialized class. For now LorentzVector is probably good enough.

coffea/nanoaod/methods/jets.py
```diff
@@ -4,6 +4,7 @@
 
 class FatJet(LorentzVector):
     '''NanoAOD large radius jet object'''
+    _enable_genjet = False
```
nsmith-, line 7: I guess we should add at some point a version specifier to the NanoEvents constructor to turn on these features for known good NanoAOD versions. The only reason the enable flag exists rather than simply checking for existence of the index map is because in v5 the index map had a bug and some indices were invalid (> length of genjet).

nsmith- (approved):
I left some comments but they are TODOs for another time.

coffea/nanoaod/methods/jets.py
andrzejnovak: Yeah the flag is not great and afaict not documented so it was a bit of digging at first.

coffea/nanoaod/methods/jets.py
andrzejnovak: Hmm yeah wanting that makes as much sense as wanting FatJet to GenJet match, tbh I am not quite sure why this was a problem.

PR #314, opened 2020-06-30
Fix binwnorm for "Sum unc." in plot1d

bfis:
The plotting of "Sum unc." band in `hist.plot.plot1d` did not incorporate the effect of `binwnorm`.


lgray:
@bfis Could you add a test that confirms this is working as expected? Thanks!

nsmith- (approved):
I suppose `density=True` isn't a valid configuration when using `stack=True` so we don't need that to change?

lgray:
Guess we can add a test if it breaks :-)

PR #319, opened 2020-07-12
Awkward1 NanoEvents

nsmith-:


lgray:
What's the default NanoEvents that people get when they import it right now?

It may be a good idea to phase the transition similar to how Jim is planning for awkward1.

nsmith-:
They're in completely separate modules, `coffea.nanoaod` vs. `coffea.nanoevents`.
The naming is opaque but the documentation makes it very explicit.
I'm going to merge this now to have the docs uploaded. Release is optional, I will just be screen sharing tomorrow.

lgray:
Ah, sorry, didn't catch that.

PR #321, opened 2020-07-16
hist.Hist.fill method performance boost

pfackeldey:
Dear coffea developers,

this PR basically substitutes the `np.add.at` DenseAxis filling with `np.bincount`. This results in a significant performance boost:
image (1)
Here the `fill` method is the current version, while the `fast_fill` method is the `np.bincount` implementation (this method naming is just for the performance showcase).

Real world performance gain:
In our full analysis setup we are processing chunks of 100k events and fill O(100k) 1D-histograms per chunk. With the new `fill` (with `np.bincount`) we were able to speed up the filling of those histograms from ~80seconds to ~5seconds (while selection, reconstruction, etc. is just about ~10 seconds) per chunk. This significantly improves the runtime for a analysis at scale!
Have a look at our performance measures in our analysis case ("Histogramming").
with  `np.add.at`:
image (2)
with `np.bincount`:
image (3)

This implementation is also very close to the `np.histogramdd` implementation: https://github.com/numpy/numpy/blob/v1.19.0/numpy/lib/histograms.py#L942-L1128

I know this PR affects a very central method, which under no circumstances should break. I've made plenty of local tests successfully, also the unittests were successful. Therefore I'd like to request a careful review, especially regarding corner cases, which I might not have catched/tested.

One downside: The values, which are filled into the DenseAxes have to be of type `np.ndarray`! I've put an Exception to catch this in the code.

Thank you very much for reviewing!
Best, Peter

### DIFF
--- a/coffea/hist/hist_tools.py
+++ b/coffea/hist/hist_tools.py
@@ -963,14 +963,28 @@ class Hist(AccumulatorABC):
                 self._sumw2[sparse_key] = np.zeros(shape=self._dense_shape, dtype=self._dtype)
 
         if self.dense_dim() > 0:
-            dense_indices = tuple(d.index(values[d.name]) for d in self._axes if isinstance(d, DenseAxis))
+            dense_indices = ()
+            for d in self._axes:
+                if isinstance(d, DenseAxis):
+                    if not isinstance(values[d.name], np.ndarray):
+                        raise Exception("Values in DenseAxes need to be numpy arrays. Got '{}' (type: {})".format(values[d.name], type(values[d.name])))
+                    dense_indices += (d.index(values[d.name]),)
+            xy = np.ravel_multi_index(dense_indices, self._dense_shape)
             if "weight" in values:
-                np.add.at(self._sumw[sparse_key], dense_indices, values["weight"])
-                np.add.at(self._sumw2[sparse_key], dense_indices, values["weight"]**2)
+                self._sumw[sparse_key][:] += np.bincount(
+                    xy, weights=values["weight"], minlength=np.array(self._dense_shape).prod()
+                ).reshape(self._dense_shape)
+                self._sumw2[sparse_key][:] += np.bincount(
+                    xy, weights=values["weight"] ** 2, minlength=np.array(self._dense_shape).prod()
+                ).reshape(self._dense_shape)
             else:
-                np.add.at(self._sumw[sparse_key], dense_indices, 1.)
+                self._sumw[sparse_key][:] += np.bincount(
+                    xy, weights=None, minlength=np.array(self._dense_shape).prod()
+                ).reshape(self._dense_shape)
                 if self._sumw2 is not None:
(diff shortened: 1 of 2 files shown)


lgray:
Excellent work! Thanks for the contribution. We'll figure out a proper review soon.

nsmith-:
Interesting, I seem to recall someone (Henry?) comparing the options and finding add.at to be the best, but clearly not any more, or the tests were not applicable. This reminds me that, really, we should migrate to boost-histogram. I believe the last of the missing functionality was added as of v0.8, so we should check it out.

lgray:
@nsmith- I see a decent definition of a 0.7. Awkward1 + boost-histogram transition.

coffea/hist/hist_tools.py
```diff
@@ -963,14 +963,28 @@ def fill(self, **values):
                 self._sumw2[sparse_key] = np.zeros(shape=self._dense_shape, dtype=self._dtype)
 
         if self.dense_dim() > 0:
-            dense_indices = tuple(d.index(values[d.name]) for d in self._axes if isinstance(d, DenseAxis))
+            dense_indices = ()
+            for d in self._axes:
+                if isinstance(d, DenseAxis):
+                    if not isinstance(values[d.name], np.ndarray):
```
nsmith-: You could consider instead to try `numpy.asarray(values[d.name])` to give a chance for compatible arrays to be cast to numpy. What is not clear to me is where integer arguments are rejected. What numpy function didn't accept them?

nsmith- (changes requested):

coffea/hist/hist_tools.py
lgray: I was about to make the same comment. We should preserve original behavior, and it doesn't look like it is that hard.

lgray: It's np.bincount - it expects array of dimension 1 exactly (and single numbers are dim 0).

nsmith-: Ah, so the solution is to wrap the `xy` indexer with `np.atleast_1d`

pfackeldey: Sure, good idea :) I can turn `values[d.name]` into a 1d array in case it is a list or a single digit with `np.atleast_1d`.

nsmith- (approved):
Thanks, this looks good

coffea/hist/hist_tools.py
```diff
@@ -964,13 +964,22 @@ def fill(self, **values):
 
         if self.dense_dim() > 0:
             dense_indices = tuple(d.index(values[d.name]) for d in self._axes if isinstance(d, DenseAxis))
+            xy = np.atleast_1d(np.ravel_multi_index(dense_indices, self._dense_shape))
```
nsmith-, line 967: A micro-optimisation would be to re-use the dense_indices variable so the input arrays can be deallocated

PR #324, opened 2020-07-21
Fix #320: Use flatten, copy and counts

### LINKED ISSUE #320: FactorizedJetCorrector incorrectly handles JaggedArrays
**Describe the bug**
The `getSubCorrections` and subsequently the `getCorrection` methods of `FactorizedJetCorrector` incorrectly handle JaggedAttays. It modifies the content of the inputs and raises an error when using a starts-stops JaggedArray.

**To Reproduce**
```
import awkward
import coffea
extract = coffea.lookup_tools.extractor()
extract.add_weight_sets(['* * tests/samples/Summer16_23Sep2016V3_MC_L2Relative_AK4PFPuppi.jec.txt.gz'])
extract.finalize()
corrector = coffea.jetmet_tools.FactorizedJetCorrector(**evaluator)
eta = awkward.JaggedArray.fromcounts([1, 1], [0., 0.])
rho = awkward.JaggedArray.fromcounts([1, 1], [30., 30.])
pt = awkward.JaggedArray.fromcounts([1, 1], [100., 100.])
area = awkward.JaggedArray.fromcounts([1, 1], [0.5, 0.5])
corrector.getCorrection(JetPt=pt, JetEta=eta, JetA=area, Rho=rho)
print(pt)
```
`pt` will surprisingly be changed by the `getCorrection` call and will be around 109 now.
On top of that with a starts-stops JaggedArray:
```
pt = awkward.JaggedArray([2, 3], [3, 4], [float("nan"), float("nan"), 100., 100.])
corrector.getCorrection(JetPt=pt, JetEta=eta, JetA=area, Rho=rho)
```
one will get a `ValueError`.

jrueb:
This fixes #320 and also adds a test case for the bug.

### DIFF
--- a/coffea/jetmet_tools/FactorizedJetCorrector.py
+++ b/coffea/jetmet_tools/FactorizedJetCorrector.py
@@ -163,12 +163,12 @@ class FactorizedJetCorrector(object):
             raise Exception('No variable to correct, need JetPt or JetE in inputs!')
         firstarg = localargs[self._signature[0]]
         cumulativeCorrection = 1.0
-        offsets = None
+        counts = None
         if isinstance(firstarg, awkward.JaggedArray):
-            offsets = firstarg.offsets
-            cumulativeCorrection = firstarg.ones_like().content
+            counts = firstarg.counts
+            cumulativeCorrection = firstarg.ones_like().flatten()
             for key in localargs.keys():
-                localargs[key] = localargs[key].content
+                localargs[key] = localargs[key].flatten().copy()
         else:
             cumulativeCorrection = np.ones_like(firstarg)
         corrections = []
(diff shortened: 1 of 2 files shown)


lgray:
Thanks!

My one comment is why you need to use counts rather than offsets?

jrueb:
You will get a value error for trying to get the offsets of something like `awkward.JaggedArray([0, 2], [1, 3], [7, 8, 9])`.
In the case of `FactorizedJetCorrector` this would happen if the eta is such a jagged array that is defined by starts and stops with gaps in-between.

lgray:
Ah right, this was only working on compact arrays. 
You could use starts and stops directly, but I think the performance difference for this function is minimal. 

Otherwise looks good. Thanks!

lgray:
Do you need a release cut for this now or can we wait for a few more items to come in?

jrueb:
If you want, you can wait. I don't mind.

PR #326, opened 2020-07-24
Add FsrPhoton support to NanoEvent

dntaylor:
This PR adds FsrPhoton handling to NanoEvent v0 and v1.

The FsrPhoton collection has existed since NanoAODv6 (I think).
It has match to muons (only, so far) that is also added (and no match to jet/gen, hence the removal of the Common lepton mixin).


nsmith- (approved):
Thanks, looks good

PR #329, opened 2020-08-10
make hist.export1d output pickleable

bfis:
Hoisted the inline class definitions in `hist.export1d` out to module level to enable `pickle` i.e. for `multiprocessing`.


bfis:
the checks failed due to an (unrelated) hash mismatch during dependency installation:
https://github.com/CoffeaTeam/coffea/runs/966866596?check_suite_focus=true#step:7:19

PR #335, opened 2020-08-20
work_queue_executor: fixed issue with absolute path names for the environment file argument

keylime-unicorn:
The work_queue_executor contained a bug in which it would not allow an absolute pathname to be specified for the environment file argument to the executor. 

Fixed with a simple call to basename().


lgray:
Thanks!

PR #339, opened 2020-08-26
Don't use colon in uproot open anymore

nsmith-:
Since it might go away: https://github.com/scikit-hep/uproot4/pull/81

### DIFF
--- a/coffea/nanoevents/factory.py
+++ b/coffea/nanoevents/factory.py
@@ -72,7 +72,7 @@ class NanoEventsFactory:
         if not issubclass(schemaclass, BaseSchema):
             raise RuntimeError("Invalid schema type")
         if isinstance(file, str):
-            tree = uproot4.open(file + ":" + treepath)
+            tree = uproot4.open(file)[treepath]
         elif isinstance(file, uproot4.reading.ReadOnlyDirectory):
             tree = file[treepath]
         if entry_start is None or entry_start < 0:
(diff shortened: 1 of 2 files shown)


jpivarski:
Turns out, it's not going away. That was a long back-and-forth, but the colon is still allowed, with two ways to disable it: `pathlib.Path("filename.root")` and the `{"filename.root": "object/in/file"}`. If you do the wrong thing, you'll get an error message that looks like this:

```
FileNotFoundError: file not found

    '/home/jpivarski/miniconda3/lib/python3.8/site-packages/skhep_testdata/data/uproot-issue63.root:WtLoop_nominal'

Files may be specified as:
   * str/bytes: relative or absolute filesystem path or URL, without any colons
         other than Windows drive letter or URL schema.
         Examples: "rel/file.root", "C:\abs\file.root", "http://where/what.root"
   * str/bytes: same with an object-within-ROOT path, separated by a colon.
         Example: "rel/file.root:tdirectory/ttree"
   * pathlib.Path: always interpreted as a filesystem path or URL only (no
         object-within-ROOT path), regardless of whether there are any colons.
         Examples: Path("rel:/file.root"), Path("/abs/path:stuff.root")

Functions that accept many files (uproot4.iterate, etc.) also allow:
   * glob syntax in str/bytes and pathlib.Path.
         Examples: Path("rel/*.root"), "/abs/*.root:tdirectory/ttree"
   * dict: keys are filesystem paths, values are objects-within-ROOT paths.
         Example: {"/data_v1/*.root": "ttree_v1", "/data_v2/*.root": "ttree_v2"}
   * already-open TTree objects.
   * iterables of the above.
```

PR #340, opened 2020-08-27
Update nanoevents.py

pfackeldey:
Dear Coffea developers,
this PR fixes docstrings for the NanoAOD format such, that e.g.:
```python
print(events.LHEPdfWeight.__doc__)
>> ''
```
becomes:
```python
print(events.LHEPdfWeight.__doc__)
>> 'LHE pdf variation weights (w_var / w_nominal) for LHA IDs 91400 - 91432'
```
Unfortunately some of the "n"+Name branches in NanoAOD have empty docstrings and thus they will be empty, but actually they are not (see: https://cms-nanoaod-integration.web.cern.ch/integration/master/mc94X_doc.html#LHEPdfWeight)


nsmith- (changes requested):
Hi,
Thanks for this! Can you also implement it for the awkward1 NanoEvents?
The relevant lines are https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/schemas.py#L254 and https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/schemas.py#L262

pfackeldey:
Hi,
I am not very familiar with awkward1 yet. From some debugging I noticed that the docstrings are still correct and fine at this point: https://github.com/CoffeaTeam/coffea/blob/f1b53330a6fdcd374790cf0700db7839b6860813/coffea/nanoevents/factory.py#L149
image

What comes out of `from_arrayset` although has wrong docstrings. I even saw the full `ak.Array` docstring instead of the one in `self._schema.form`:
image
Do you have an idea/hint what goes wrong in this step?

pfackeldey:
At the relevant lines, which you send, I do not see any problems with the docstring. They seem to be correct.

pfackeldey:
I just found that this one is correct:
image
For some reason the bound `__doc__` method gets overwritten though, I don't see a reason for this to fail yet... (https://github.com/scikit-hep/awkward-1.0/blob/3365376ba6bec56accd000a55483eabe2438f64a/src/awkward1/highlevel.py#L286-L288)

nsmith-:
Oh, I assumed that I would have implemented things incorrect in the same way between the ak0 and ak1, so just pointed to where I expected a change. If no change is needed and `NanoAODSchema` makes the right form already then nevermind. What you found looks more like a bug in awkward1, thanks for starting to look into it!

nsmith- (approved):

nsmith-:
I recognize the issue with the extremely long docstring, it's because on first access to `events.X` what awkward1 does is actually set up a new virtual array that has a slice generator. However, it doesn't have a complete `form` and therefore `purelist_parameter` does not see the correct value. For `__record__` this was a big issue, which I fixed in https://github.com/scikit-hep/awkward-1.0/pull/390 Looks like we should add `__doc__` as well since it is also used in each highlevel array construction.

nsmith-:
Should be fixed by https://github.com/scikit-hep/awkward-1.0/pull/422

PR #341, opened 2020-08-27
Implemented multi axes Hist.scale

bfis:
Enables Hist.scale to use a dict with keys referring to multiple sparse axes.


nsmith- (approved):
This looks good, thanks for the added feature!

PR #343, opened 2020-09-04
Some nanoevents improvements

nsmith-:


nsmith-:
The work_queue test is apparently now broken, which should be unrelated to this PR

PR #344, opened 2020-09-04
Awkward1 lookup_tools wrap

nsmith-:
This might be it?!

### DIFF
--- a/coffea/lookup_tools/lookup_base.py
+++ b/coffea/lookup_tools/lookup_base.py
@@ -1,5 +1,7 @@
-from ..util import awkward
-from ..util import numpy as np
+import numpy
+import awkward
+import awkward1
+import numbers
 
 
 class lookup_base(object):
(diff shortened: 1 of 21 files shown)


coffea/lookup_tools/lookup_base.py
```diff
@@ -8,7 +10,33 @@ def __init__(self):
...
+                    raise NotImplementedError("support for cupy/jax/etc. numpy extensions")
+                result = self._evaluate(*[nplike.asarray(x) for x in inputs])
+                return lambda: (awkward1.layout.NumpyArray(result),)
+            return None
+
+        behavior = awkward1._util.behaviorof(*args)
+        args = [awkward1.operations.convert.to_layout(arg, allow_record=False, allow_other=True) for arg in args]
+        out = awkward1._util.broadcast_and_apply(args, getfunction, behavior)
+        assert isinstance(out, tuple) and len(out) == 1
+        return awkward1._util.wrap(out[0], behavior)
```
nsmith-: @jpivarski I lifted most of this from the numpy [ufunc](https://github.com/scikit-hep/awkward-1.0/blob/master/src/awkward1/_connect/_numpy.py#L40) implementation. Does it look general enough to put in awkward as a public method? I think maybe a nice extension would be some sort of single-dispatch based on `nplike`. In this context, we mandate all items are broadcastable numeric arrays, so I left out the `adjust` part, is that important in this context?

jpivarski: Is `lookup_base` an abstract class and something else defines `_evaluate`?

jpivarski: I haven't been able to figure out yet what this does.

nsmith-: Correct, the derived classes implement `_evaluate(*numpy_arrays)`, e.g. https://github.com/CoffeaTeam/coffea/blob/master/coffea/lookup_tools/dense_lookup.py#L31

nsmith-:
This covers most all `lookup_tools` awkward1 conversions. There are some references to awkward0 in some places that I don't know their purpose. Ideally any class deriving from `lookup_base` should only receive numpy-like arrays.

PR #348, opened 2020-09-21
Update executor to uproot4 and switch to using NanoEventsFactory

aperloff:
Make it so that the executor/processor can use the NanoEventsFactory (uproot4/awkward1) module in addition to the NanoEvents (uproot3/awkward0) module. The NanoEventsFactory works with multiple flat ntuple schemas; the primary one being the NanoAODSchema, but also comes with the TreeMakerShema. An example call to the executor using the new schema option will look like:
```python
    output = processor.run_uproot_job(sample_dict,
                                      treename='TreeMaker2/PreSelection',
                                      processor_instance=SampleProcessor(),
                                      executor=processor.iterative_executor,
                                      executor_args={
                                          'skipbadfiles':True,
                                          'schema': processor.NanoAODSchema, 
                                          'flatten':True, 
                                          'workers': 4},
                                      chunksize=50000, maxchunks=100
                                     )
```
While I believe the code is in good shape, this PR will need #350 to be fully functional. This was initially tested using the TreeMakerSchema and successfully rendered all of the branches and processed multiple files. However, that test included some code which transitioned more of the processor to uproot4/awkward1, which didn't go into this PR. Without those changes, we can see no obvious signs of failure, but not enough of the code has been transitioned in order to perform a full test.

### DIFF
--- a/coffea/nanoevents/factory.py
+++ b/coffea/nanoevents/factory.py
@@ -2,7 +2,7 @@ import warnings
 import weakref
 import json
 import awkward1
-import uproot4
+import uproot4 as uproot
 from coffea.nanoevents.util import quote, key_to_tuple, tuple_to_key
 from coffea.nanoevents.mapping import TrivialOpener, UprootSourceMapping, CachedMapping
 from coffea.nanoevents.schemas import BaseSchema, NanoAODSchema
@@ -49,8 +49,8 @@ class NanoEventsFactory:
 
         Parameters
         ----------
-            file : str or uproot4.reading.ReadOnlyDirectory
-                The filename or already opened file using e.g. ``uproot4.open()``
+            file : str or uproot.reading.ReadOnlyDirectory
+                The filename or already opened file using e.g. ``uproot.open()``
             treepath : str, optional
                 Name of the tree to read in the file
             entry_start : int, optional
@@ -69,14 +69,18 @@ class NanoEventsFactory:
             metadata : dict, optional
                 Arbitrary metadata to add to the `base.NanoEvents` object
             uproot_options : dict, optional
-                Any options to pass to ``uproot4.open``
+                Any options to pass to ``uproot.open``
         """
         if not issubclass(schemaclass, BaseSchema):
             raise RuntimeError("Invalid schema type")
         if isinstance(file, str):
-            tree = uproot4.open(file, **uproot_options)[treepath]
-        elif isinstance(file, uproot4.reading.ReadOnlyDirectory):
+            tree = uproot.open(file, **uproot_options)[treepath]
+        elif isinstance(file, uproot.reading.ReadOnlyDirectory):
             tree = file[treepath]
+        elif "<class 'uproot.rootio.ROOTDirectory'>" == str(type(file)):
+            raise RuntimeError("The file intance (%s) is an uproot3 type, but this module is only compatible with uproot4 or higher" % (str(type(file))))
+        else:
+            raise RuntimeError("Invalid file instance (%s)" % (str(type(file))))
         if entry_start is None or entry_start < 0:
             entry_start = 0
         if entry_stop is None or entry_stop > tree.num_entries:
@@ -114,7 +118,7 @@ class NanoEventsFactory:
             if len(branch):
                 continue
             form = branch.interpretation.awkward_form(None)
-            form = uproot4._util.awkward_form_remove_uproot(awkward1, form)
+            form = uproot._util.awkward_form_remove_uproot(awkward1, form)
             form = json.loads(form.tojson())
             if (
                 form["class"].startswith("ListOffset")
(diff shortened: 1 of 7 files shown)


coffea/nanoevents/factory.py
```diff
@@ -69,14 +69,18 @@ def from_file(
...
         if not issubclass(schemaclass, BaseSchema):
             raise RuntimeError("Invalid schema type")
         if isinstance(file, str):
-            tree = uproot4.open(file, **uproot_options)[treepath]
-        elif isinstance(file, uproot4.reading.ReadOnlyDirectory):
+            tree = uproot.open(file, **uproot_options)[treepath]
+        elif isinstance(file, uproot.reading.ReadOnlyDirectory):
             tree = file[treepath]
+        elif "<class 'uproot.rootio.ROOTDirectory'>" == str(type(file)):
+            raise RuntimeError("The file intance (%s) is an uproot3 type, but this module is only compatible with uproot4 or higher" % (str(type(file))))
```
nsmith-: `"The file instance (%r) is an uproot3 type, but this module is only compatible with uproot4 or higher" % file)`

coffea/nanoevents/factory.py
```diff
@@ -69,14 +69,18 @@ def from_file(
...
+        else:
+            raise RuntimeError("Invalid file instance (%s)" % (str(type(file))))
```
nsmith-: Invalid file type, `TypeError`

coffea/processor/__init__.py
```diff
@@ -27,6 +27,13 @@
     defaultdict_accumulator,
     column_accumulator,
 )
+from ..nanoevents.schemas import (
```
nsmith-: I know existing coffea code has relative imports but I'm trying to stick to absolute imports going forward

coffea/processor/executor.py
```diff
@@ -774,18 +773,38 @@ def localsource(path):
...
+                        schemaclass=schema,
+                        metadata={
+                            'dataset': item.dataset,
+                            'filename': item.filename,
+                        },
+                    )
+                    df = factory.events()
+                elif issubclass(schema, NanoEvents):
+                    file = uproot.open(item.filename, localsource=localsource, xrootdsource=xrootdsource)
+                    df = NanoEvents.from_file(
```
nsmith-, line 793: For awkward0 NanoEvents I'm pretty sure you need to pass an uproot3 file object for it to work. How about we just pass the filename to `NanoEvents` so that it can open with uproot3 and then we never import `uproot3` inside this module. You can set `file = None` and skip the later checks and performance counters when this path is taken.

coffea/processor/spark/spark_executor.py
```diff
@@ -110,8 +111,13 @@ def spex_accumulator(total, result):
                 co_udf = coffea_udf
                 if flatten:
                     co_udf = coffea_udf_flat
-                if nano:
-                    co_udf = coffea_udf_nano
+                if schema is not None:
+                    if issubclass(schema, NanoEvents):
+                        co_udf = coffea_udf_nano
+                    elif issubclass(schema, schemas.BaseSchema):
+                        raise ValueError("The uproot4 version of NanoEvents (%s) has not been implemented yet" % (str(schema.__name__)))
```
nsmith-: NotImplementedError

nsmith- (changes requested):
Some nitpicks mostly. I think this PR might need to also take care of updating the performance counters to use uproot4 as well.

coffea/processor/executor.py
aperloff: Since the `uproot3` module is used in other places like:
```python
            from uproot.source.xrootd import XRootDSource
            xrootdsource = XRootDSource.defaults
            xrootdsource['timeout'] = xrootdtimeout
```
or
```python
    if mmap:
        localsource = {}
    else:
        opts = dict(uproot.FileSource.defaults)
        opts.update({'parallel': None})

        def localsource(path):
            return uproot.FileSource(path, **opts)
```
we're going to have to import `uproot3` anyway. That is, at least until #350 is merged in. Do you still want me to make this change?

aperloff:
Shouldn't the performance counters be updated to use uproot4 in #350 and not in this PR?

coffea/processor/executor.py
nsmith-: You're right, in fact now I realize that the awkward1/uproot4 version is the only one not working here.

coffea/processor/executor.py
```diff
@@ -1042,7 +1070,7 @@ def run_uproot_job(fileset,
     savemetrics = executor_args.pop('savemetrics', False)
     flatten = executor_args.pop('flatten', False)
     mmap = executor_args.pop('mmap', False)
-    nano = executor_args.pop('nano', False)
```
nsmith-, line 1045: To help with migration, can we leave `nano` here, and issue a deprecation warning? e.g.
```python
if nano:
    warnings.warn("Please use 'schema': processor.NanoEvents rather than 'nano': True to enable awkward0 NanoEvents processing", DeprecationWarning)
    schema = NanoEvents
```

nsmith- (approved):
Looks good. I'll start digging into #350 this afternoon

PR #350, opened 2020-09-22
Make the processor work using uproot4/awkward1

aperloff:
This pull request will make the processor/executor work with uproot4/awkward1, but will still retain the ability to return awkward0 arrays for those people who are still using awkward0. This PR is necessary to make #348 usable. It is based on #348, to make it easier to test.

This is still somewhat of a work in progress. There are some features from uproot3/awkward0 that I couldn't find a replacement for in uproot4/awkward1. However, they may not be needed anymore.

Places where I know I need some guidance:
1. In `_work_function()` the file open line used to be:
```python
uproot.open(item.filename, localsource=localsource, xrootdsource=xrootdsource)
```
In uproot4 the options localsource and xrootdsource no longer exist. This seemingly means the following block of code from the same function isn't necessary:
```python
    if mmap:
        localsource = {}
    else:
        opts = dict(uproot.FileSource.defaults)
        opts.update({'parallel': None})

        def localsource(path):
            return uproot.FileSource(path, **opts)
```
and that the `mmap` option isn't needed. Are there valid uproot4/awkward1 replacements or was I correct in just removing these lines? Note, I left the `mmap` option for now just for compatibility.
2. The `_read` attribute of `uproot.source.xrootd.XRootDSource` no longer exists in uproot4 and I couldn't find a replacement. Is this block of code still needed?
```python
# instrument xrootd source
if not hasattr(uproot.source.xrootd.XRootDSource, '_read_real'):
    def _read(self, chunkindex):
        self.bytesread = getattr(self, 'bytesread', 0) + self._chunkbytes
        return self._read_real(chunkindex)

    uproot.source.xrootd.XRootDSource._read_real = uproot.source.xrootd.XRootDSource._read
    uproot.source.xrootd.XRootDSource._read = _read
```
3. In `processor/dataframe.py` I'm not sure what to do with this line:
```python
self._branchargs = {'library': awkward, 'flatten': flatten}
```
'flatten' is no longer an available option.
4. There are lots more, but I only got part of the way through the transition.


nsmith-:
Item 1 was solved via https://github.com/CoffeaTeam/coffea/pull/350/files#diff-2910e237aa690c4f123d26b3749619c3R787-R790 it appears the `timeout` is now a top-level option to `uproot4.open`, and the source can be controlled by passing the appropriate class.
I'm dropping support for the `bytesread` metric in awkward0-based NanoEvents to avoid importing uproot3 inside the executor module. The LazyDataFrame and awkward1-based NanoEvents still have it because uproot4 now has a built-in performance counter: https://github.com/CoffeaTeam/coffea/pull/350/files#diff-2910e237aa690c4f123d26b3749619c3L818
To fix item 4, I rewrote the LazyDataFrame to properly flatten awkward1 arrays or cast them as awkward0 JaggedArrays via https://github.com/CoffeaTeam/coffea/pull/350/files#diff-6a09dd7e5c8e03cd6def2d8ff9452ce6R52-R55

aperloff:
Why did you get rid of https://github.com/CoffeaTeam/coffea/pull/350/files#diff-2910e237aa690c4f123d26b3749619c3L824-L825? Does uproot4 have the ability to close a file that goes out of context?

nsmith-:
Right, uproot4 files follow PEP 343 it appears https://www.python.org/dev/peps/pep-0343/

coffea/processor/dataframe.py
```diff
@@ -26,15 +23,23 @@ class LazyDataFrame(MutableMapping):
...
-        self._branchargs = {'awkwardlib': awkward, 'flatten': flatten}
-        entrystart, entrystop = uproot.tree._normalize_entrystartstop(tree.numentries, entrystart, entrystop)
-        self._branchargs['entrystart'] = entrystart
-        self._branchargs['entrystop'] = entrystop
-        self._available = {k.decode('ascii') for k in self._tree.keys()}
+        self._branchargs = {
+            "decompression_executor": uproot4.source.futures.TrivialExecutor(),
+            "interpretation_executor": uproot4.source.futures.TrivialExecutor(),
+        }
+        if entrystart is None or entrystart < 0:
```
aperloff, line 36: Why not use https://github.com/scikit-hep/uproot4/blob/f1e66b552e930df4876c32545819918fab669cf9/uproot4/behaviors/TBranch.py#L2805-L2821?

coffea/processor/dataframe.py
```diff
@@ -26,15 +23,23 @@ class LazyDataFrame(MutableMapping):
...
         self._tree = tree
         self._flatten = flatten
-        self._branchargs = {'awkwardlib': awkward, 'flatten': flatten}
-        entrystart, entrystop = uproot.tree._normalize_entrystartstop(tree.numentries, entrystart, entrystop)
-        self._branchargs['entrystart'] = entrystart
-        self._branchargs['entrystop'] = entrystop
-        self._available = {k.decode('ascii') for k in self._tree.keys()}
+        self._branchargs = {
+            "decompression_executor": uproot4.source.futures.TrivialExecutor(),
+            "interpretation_executor": uproot4.source.futures.TrivialExecutor(),
```
aperloff, line 34: What do these executors do which is like the former 'awkwardlib' and 'flatten' options?

nsmith-:
@aperloff let me know if you're happy with this and I'll merge

tests/test_spark.py
```diff
@@ -72,44 +71,10 @@ def test_spark_executor():
...
-
-    filelist = {'ZJets': {'files': ['file:'+osp.join(os.getcwd(),'tests/samples/nano_dy.root')], 'treename': 'Events' },
-                'Data'  : {'files': ['file:'+osp.join(os.getcwd(),'tests/samples/nano_dimuon.root')], 'treename': 'Events'}
-                }
-
-    from coffea.processor.test_items import NanoEventsProcessor
-    from coffea.processor.spark.spark_executor import spark_executor
-
-    columns = ['nMuon','Muon_pt','Muon_eta','Muon_phi','Muon_mass', 'Muon_charge']
-    proc = NanoEventsProcessor(columns=columns)
```
aperloff, line 110: Why isn't any of the above needed anymore?

tests/test_spark.py
```diff
@@ -72,44 +71,10 @@ def test_spark_executor():
...
+    proc = NanoEvents0Processor(columns=columns)
     hists = run_spark_job(filelist, processor_instance=proc, executor=spark_executor, spark=spark, thread_workers=1,
-                          executor_args={'file_type': 'root', 'schema': processor.NanoAODSchema})
+                          executor_args={'file_type': 'root', 'nano': True})
```
aperloff, line 77: Why put back the soon to be deprecated option?

coffea/processor/dataframe.py
nsmith-: We have simpler logic here (no fancy negative numbers for wrapping) and that is a private function so its not guaranteed to be available/stable.

coffea/processor/dataframe.py
nsmith-: These are more or less independent issues. This forces uproot to be single-threaded in this context (which we want since the coffea processing is already multi-thread and we don't want to accidentally occupy more CPU than the user requests)

tests/test_spark.py
nsmith-: It looks to me like it's more straightforward to just copy the `NanoEvents0Processor` section but with `NanoEventsProcessor` once the spark execution is available for the new NanoEvents. I'll work on that in a separate PR.

tests/test_spark.py
nsmith-: Actually no reason, but this is also the only place where the deprecated call style is tested so I left it for now.

coffea/processor/dataframe.py
aperloff: Okay, good reasons.

coffea/processor/dataframe.py
aperloff: Okay, and there is somewhere else in the code where the `flatten` argument is passed or is it unnecessary becuase you're asking all processors to be able to work with jaggedarrays?

tests/test_spark.py
aperloff: Okay

tests/test_spark.py
aperloff: Okay, I guess that's a good enough reason.

coffea/processor/dataframe.py
nsmith-: Indeed I marked `flatten` as deprecated: https://github.com/CoffeaTeam/coffea/pull/350/files#diff-2910e237aa690c4f123d26b3749619c3R1083 but it is still applied by the following code: https://github.com/CoffeaTeam/coffea/pull/350/files#diff-6a09dd7e5c8e03cd6def2d8ff9452ce6R57-R58

aperloff: Okay, fair enough.

aperloff:
@nsmith- This PR looks good to me. I think it's ready to be merged (with a name change).

PR #356, opened 2020-10-07
fix: filemeta not respecting skipbadfiles

andrzejnovak:
`run_uproot_job` was crashing when `maxchunks` was set and bad files were present along with `skipbadfiles` turned on, due to misordered logic

This fixes it and piggy backs a test on `test_nanoevents_analysis`, let me know if I should write a separate test instead


andrzejnovak:
@nsmith- seems ready for review

nsmith- (approved):
Thanks looks good

nsmith-:
There's some weirdness with macOS tests, also on master. I'll just merge

PR #362, opened 2020-10-29
fix genJetIdx handling

bfis:
`genJetIdx` may point beyond the available `GenJet`s, since the latter are 
truncated according to a minimum pT.
A non-negative `genJetIdx` indicates the existence of a `GenJet` and may 
be used to infer their pT ordering (in case the corresponding `GenJet`s 
were truncated), thus it should be kept available to the user.

### DIFF
--- a/coffea/nanoaod/methods/jets.py
+++ b/coffea/nanoaod/methods/jets.py
@@ -139,7 +139,6 @@ class Jet(LorentzVector):
             )
             embedded_genjet.__doc__ = genjet.__doc__
             self['matched_gen'] = embedded_genjet
-            del self['genJetIdx']
 
         if 'JetPFCands' in events.columns:
             pfcand_link = events['JetPFCands']
--- a/coffea/nanoaod/nanoevents.py
+++ b/coffea/nanoaod/nanoevents.py
@@ -131,7 +131,7 @@ class NanoCollection(awkward.VirtualArray):
         index.type.takes = self.array.offsets[-1]
         index = awkward.JaggedArray.fromoffsets(self.array.offsets, content=index)
         globalindex = (index + destination.array.starts).flatten()
-        invalid = (index < 0).flatten()
+        invalid = ((destination.array.counts <= index) | (index < 0)).flatten()
         if any(invalid):
             globalindex[invalid] = -1
             # note: parent virtual must derive from this type
--- a/coffea/nanoevents/transforms.py
+++ b/coffea/nanoevents/transforms.py
@@ -109,6 +109,7 @@ def local2global(stack):
     target_offsets = stack.pop()
     index = stack.pop()
     index = index.mask[index >= 0] + target_offsets[:-1]
+    index = index.mask[index < target_offsets[1:]]
     out = numpy.array(awkward1.flatten(awkward1.fill_none(index, -1)))
     if out.dtype != numpy.int64:
         raise RuntimeError


nsmith-:
Isn't this a bug in NanoAOD? I would be worried that if the indices exceed the destination collection size, that there may be other filtering that could invalidate the mapping. I suppose pT filtering is relatively safe since the collection is (presumably) pT-ordered, but are we sure this is the case? Have you done a delta-R sanity check?

Nevertheless, added protections may be a good idea. Would you be willing to make the equivalent change also to the awkward1 version of NanoEvents? It should be as simple as adding a line after https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/transforms.py#L111 like `index = index.mask[index < target_offsets[1:]]`

bfis:
> Isn't this a bug in NanoAOD?

No, this is intended and  [documented](https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookNanoAOD#The_Events_TTree), albeit confusing, behavior.

> I would be worried that if the indices exceed the destination collection size, that there may be other filtering that could invalidate the mapping.

Yes, this will become a problem if/when the [current filtering](https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/NanoAOD/python/jets_cff.py#L593) changes. However, this seems to be more of a (format-)definition issue on the NanoAOD production side.

> I suppose pT filtering is relatively safe since the collection is (presumably) pT-ordered, but are we sure this is the case?

As far as I understand the documentation, this is the case.

> Have you done a delta-R sanity check?

I have not.

> Would you be willing to make the equivalent change also to the awkward1 version of NanoEvents? It should be as simple as adding a line after https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/transforms.py#L111 like `index = index.mask[index < target_offsets[1:]]`

Can do. However, I do not have the means to properly test this.

andrzejnovak:
If the protection does go in, it should not happen silently. There should be a warning if there are any imprecisions being hidden

nsmith-:
Yeah, this bothers me because this kind of check has caught bugs in earlier Nano-like nTuples. But we would have to ignore the warning when it applies to GenJets.

bfis:
> If the protection does go in, it should not happen silently.

It's less of a protection and more like a correction of behavior. The current (buggy) implementation causes GenJets to silently jump between events in almost maximally problematic way i.e. a Jet with an associated GenJet with pT<10 "receives" one of the GenJets of the following event(s), which (on average) have much higher pT.

> There should be a warning if there are any imprecisions being hidden

I don't think this is too helpful, as the NanoAOD format itself defines these cutoffs. Practically all object collections have some min. pT cutoff. Then, this could (in principle - I've not checked the cms-sw code there) also affect the muons/electrons for a Jet. I don't think a user would profit from excessive warnings, sine this occurrence is rather common in certain datasets (e.g. "TTGJets")

nsmith-:
yikes, I guess since it is a global index then yeah it would cross boundaries.. certainly this PR is an improvement then

nsmith-:
I checked the awkward1 impl, seems to work:
```python
In [15]: ak.any(events.Jet.genJetIdx >= ak.num(events.GenJet))
Out[15]: True
```
and histogramming `events.Jet.matched_gen.delta_r(events.Jet)` gives:
matched_gen

for a random TTBar file I had

PR #363, opened 2020-10-31
Add py39 to tests

lgray:
Starting this PR to see where things fail.


lgray:
weird, pytest doesn't install on windows for py39... Guess we need to wait a bit...

PR #364, opened 2020-11-02
JetMET tools that work with awkward1

lgray:
This is the first of two PRs.

1 - make it so the current tools are agnostics to underlying awkward version
2 - integrate the JetMET tools with nanoaod/nanoevents in a nice way.

This PR implements:

- [x] Changes to the low level JetMET tools that make them not care about array library type (and also removes all in-place operations)
- [x] Introduces a new tool for managing JECs, leaving the JetTransformer as it is. 
- [x] Ditto for MET
- [x] Allow the stack to work without a FactorizedJetCorrector defined.

### DIFF
--- a/coffea/jetmet_tools/CorrectedJetsFactory.py
+++ b/coffea/jetmet_tools/CorrectedJetsFactory.py
@@ -1,0 +1,370 @@
+from .JECStack import JECStack
+from coffea.util import awkward
+from coffea.util import awkward1
+from coffea.util import numpy as np
+from collections import namedtuple
+import warnings
+from copy import copy
+from functools import partial
+
+_stack_parts = ['jec', 'junc', 'jer', 'jersf']
+_MIN_JET_ENERGY = 1e-2
+
+
+def rand_gauss_ak0(arr, var):
+    item = arr[var]
+    wrap, _ = awkward.util.unwrap_jagged(item,
+                                         item.JaggedArray,
+                                         (item, ))
+    return wrap(np.random.normal(size=item.content.size))
+
+
+def rand_gauss_ak1(arr, var):
+    item = arr[var]
+
+    def getfunction(layout, depth):
+        if (isinstance(layout, awkward1.layout.NumpyArray)
+            or not isinstance(layout, (awkward1.layout.Content,
+                                       awkward1.partition.PartitionedArray)
+           )):
+            return lambda: awkward1.layout.NumpyArray(np.random.normal(size=awkward1.count(item)))
+        return None
+
+    out = awkward1._util.recursively_apply(
+        awkward1.operations.convert.to_layout(item),
+        getfunction)
+    assert out is not None
+    return awkward1._util.wrap(out, awkward1._util.behaviorof(item))
+
+
+def jer_smear(arr, variation, forceStochastic, ptGenJet, ptJet, etaJet):
+    pt_gen = arr[ptGenJet] if not forceStochastic else None
+    jetPt = arr[ptJet]
+
+    if isinstance(jetPt, awkward.array.base.AwkwardArray):
+        if forceStochastic:
+            wrap, arrays = awkward.util.unwrap_jagged(jetPt,
+                                                      jetPt.JaggedArray,
+                                                      (jetPt, ))
+            pt_gen = wrap(np.zeros_like(arrays[0]))
+    elif isinstance(jetPt, awkward1.highlevel.Array):
+        def getfunction(layout, depth):
+            if (isinstance(layout, awkward1.layout.NumpyArray)
+                or not isinstance(layout, (awkward1.layout.Content,
+                                           awkward1.partition.PartitionedArray)
+               )):
+                return lambda: awkward1.layout.NumpyArray(np.zeros_like(size=awkward1.count(jetPt)))
+            return None
+        if forceStochastic:
+            pt_gen = awkward1._util.recursively_apply(
+                awkward1.operations.convert.to_layout(jetPt),
+                getfunction)
+            pt_gen = awkward1._util.wrap(pt_gen, awkward1._util.behaviorof(jetPt))
+    else:
+        raise Exception('\'arr\' must be an awkward array of some kind!')
+
+    jersmear = arr['jet_energy_resolution'] * arr['jet_resolution_rand_gauss']
+    jersf = arr['jet_energy_resolution_scale_factor'][:, :, variation]
+    doHybrid = pt_gen > 0.
+
+    detSmear = 1. + (jersf - 1.) * (-pt_gen + jetPt) / jetPt  # because of #367
+    stochSmear = 1. + np.sqrt(np.maximum(jersf**2 - 1., 0.)) * jersmear
+
+    min_jet_pt = _MIN_JET_ENERGY / np.cosh(arr[etaJet])
+    min_jet_pt_corr = min_jet_pt / jetPt
+
+    smearfact = None
+    if isinstance(arr, awkward.array.base.AwkwardArray):
+        wrap, arrays = awkward.util.unwrap_jagged(jetPt,
+                                                  jetPt.JaggedArray,
+                                                  (jetPt, min_jet_pt_corr))
+        smearfact = np.where(doHybrid.content, detSmear.content, stochSmear.content)
+        smearfact = np.where((smearfact * arrays[0]) < arrays[1], arrays[1], smearfact)
+        smearfact = wrap(smearfact)
+    elif isinstance(arr, awkward1.highlevel.Array):
+        smearfact = awkward1.where(awkward1.flatten(doHybrid),
+                                   awkward1.flatten(detSmear),
+                                   awkward1.flatten(stochSmear))
+        smearfact = awkward1.where((smearfact * awkward1.flatten(jetPt)) < awkward1.flatten(min_jet_pt),
+                                   awkward1.flatten(min_jet_pt_corr),
+                                   smearfact)
+
+        def getfunction(layout, depth):
+            if (isinstance(layout, awkward1.layout.NumpyArray)
+                or not isinstance(layout, (awkward1.layout.Content,
+                                           awkward1.partition.PartitionedArray)
+               )):
+                return lambda: awkward1.layout.NumpyArray(smearfact)
+            return None
+        smearfact = awkward1._util.recursively_apply(
+            awkward1.operations.convert.to_layout(jetPt),
+            getfunction)
+        smearfact = awkward1._util.wrap(smearfact, awkward1._util.behaviorof(jetPt))
+    else:
+        raise Exception('\'arr\' must be an awkward array of some kind!')
+
+    return smearfact
+
+
+class JetUncertainty:
+    def __init__(self, jets, has_jer, forceStochastic, name_map, VirtualType, uncertainty, lazy_cache):
+        self.jets = jets
+        self.has_jer = has_jer
+        self.forceStochastic = forceStochastic
+        self.name_map
+        self.VirtualType = VirtualType
+        self.uncertainty = uncertainty
+        self.lazy_cache = lazy_cache
+        if uncertainty is None:  # JER uncertainty are [central, up, down]
+            self.UP = 1
+            self.DOWN = 2
+        else:  # JES uncertainties are [up, down]
+            self.UP = 0
+            self.DOWN = 1
+
+    def build_unc(self, up_down):
+        out = None
+        if isinstance(self.jets, awkward.array.base.AwkwardArray):
+            out = self.jets.copy()
+        elif isinstance(self.jets, awkward1.highlevel.Array):
+            out = copy(self.jets)
+        else:
+            raise Exception('\'jets\' must be an awkward array of some kind!')
+
+        if self.uncertainty is None:  # JER uncertainties
+            def jer_smeared_corr(arr, variation, forceStochastic, ptGenJet, ptJet, etaJet):
+                return jer_smear(arr, variation, forceStochastic, ptGenJet, ptJet, etaJet)
+
+            out['jet_energy_resolution_correction'] = self.VirtualType(jer_smeared_corr,
+                                                                       args=(out, up_down, self.forceStochastic,
+                                                                             self.name_map['ptGenJet'],
+                                                                             self.name_map['JetPt'],
+                                                                             self.name_map['JetEta']),
+                                                                       length=len(out),
+                                                                       cache=self.lazy_cache)
+
+            def jer_smeared_val(arr, varName):
+                return arr['jet_energy_resolution_correction'] * arr['jet_energy_correction'] * arr[varName]
+
+            out[self.name_map['JetPt']] = self.VirtualType(jer_smeared_corr,
+                                                           args=(out, self.name_map['ptRaw']),
+                                                           length=len(out),
+                                                           cache=self.lazy_cache)
+            out[self.name_map['JetMass']] = self.VirtualType(jer_smeared_corr,
+                                                             args=(out, self.name_map['massRaw']),
+                                                             length=len(out),
+                                                             cache=self.lazy_cache)
+        else:  # JES uncertainties
+            out['jet_energy_uncertainty'] = self.uncertainty
+
+            def junc_smeared_val(arr, up_down, has_jer, varName):
+                base = arr['jet_energy_correction'] * arr[varName]
+                if has_jer:
+                    base = arr['jet_energy_resolution_correction'] * base
+                return out['jet_energy_uncertainty'][up_down] * base
+
+            out[self.name_map['JetPt']] = self.VirtualType(jer_smeared_corr,
+                                                           args=(out, up_down, self.has_jer,
+                                                                 self.name_map['ptRaw'],
+                                                                 ),
+                                                           length=len(out),
+                                                           cache=self.lazy_cache)
+            out[self.name_map['JetMass']] = self.VirtualType(jer_smeared_corr,
+                                                             args=(out, up_down, self.has_jer,
+                                                                   self.name_map['massRaw']
+                                                                   ),
+                                                             length=len(out),
+                                                             cache=self.lazy_cache)
+
+        return out
+
+    def up(self):
+        return self.build_unc(self.UP)
+
+    def down(self):
+        return self.build_unc(self.DOWN)
+
+
+class JetUncertaintyHandler:
+    def __init__(self, jets, has_jer, forceStochastic, name_map, VirtualType, uncertainties=None, lazy_cache=None):
+        self.jets = jets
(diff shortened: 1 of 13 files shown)


bfis:
If you plan to introduce a new tool/interface for JEC anyway, it would be nice if it handled shifts with an iterator/generator, such that one can process all shifts without the need to have them in memory at the same time.

NJManganelli:
> If you plan to introduce a new tool/interface for JEC anyway, it would be nice if it handled shifts with an iterator/generator, such that one can process all shifts without the need to have them in memory at the same time.

What's the estimated memory usage of e.g. O(100) shifts? Using e.g. iterative execution + generator would necessitate having to load everything into memory repeatedly, for each iteration of the data to process, if I'm thinking this through correctly.

lgray:
@bfis @NJManganelli My plan here was use go for lazy evaluation with caching of the corrections. Given the cited issue with memory, I can make the cache LRU with maxsize=10 or so (default, but tunable). This should be a suitable tradeoff in speed vs. memory given the typical use cases. The systematics tend to not be needed for very long in code for the most part. 

I will likely cache just the corrections and do all the multiplication on demand, since the last step of multiplication by pt/mass is fast.

Syntax I am thinking:
```python3
cJets = CorrectedJets(events.Jets,
                      <map to standardized jet/met property names> , 
                      <stack of JECs>) # FYI Jet/MET inputs can be non-NanoEvents
cJets.pt # central value, corrected
cJets['syst_name'].up.pt # pt shifted up for a given systematic on JES or JERSF
cJets['syst_name'].down.pt # pt shifted down '''                            '''
cJets['syst_name'].up.shift # access to bare correction
# iterable over keys 

cMET = CorrectedMet(events.MET, cJets, <name mapping>)
# ditto
```

bfis:
@NJManganelli

> What's the estimated memory usage of e.g. O(100) shifts?

It should be `N_events * N_shifts * 2 * (mean(N_jets) + 1)` floats/doubles, so in the order of `N_jets * N_shifts` (e.g. O(1000)). We currently use 100k events chunks and end up with upto 1GB of "overhead".

> Using e.g. iterative execution + generator would necessitate having to load everything into memory repeatedly, for each iteration of the data to process, if I'm thinking this through correctly.

That's correct, however those could be cached using e.g. the [`ColumnCache`](https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/dask/__init__.py#L9).

@lgray 

That cache idea seems great! As long as the cache can be tuned/limited even a (default) unbounded cache seems a good choice.

lgray:
@bfis Yeah, LRU cache can take a maxsize=None to revert to no-timeout mode. I think it's probably the best choice, since it offers a pretty good knob to turn.

bfis:
Also, would it be possible provide/emit corrected Jet/MET objects, that behave identical to the original ones i.e. "copies" with only the pT, mass, etc. overridden by the corrected values? This way one can use the same code to process nominal and shifted jets/events and only need to differentiate between them later i.e. during histogram filling.

lgray:
@bfis Ah gotcha. I think this is covered already by the structure mentioned above + a little shim. 

Would aim for a setup like this:

```python3
cJets = CorrectedJets(events.Jets,
                      <map to standardized jet/met property names> , 
                      <stack of JECs>) # FYI Jet/MET inputs can be non-NanoEvents
cJets.pt # cJets acts like a NanoJets object with some overrides
shifted = cJets['syst_name'].up # shifted '''                '''
```

Unfortunately, this won't act completely like an awkward mixin, but should be close enough to not break the typical idioms.

nsmith-:
On the subject of caching, be sure to use the existing runtime array cache of the NanoEvents object, e.g. `obj.cache` because creating a new cache object means they now contend with each other for space, and the user has to manage two quotas. Personally I would lean towards the generator style, ensuring the corrected arrays get decref'd to zero and get cleaned up eagerly rather than having them get evicted by a cache.

Regarding awkward mixin, what prevents you from using mixins and/or awkward virtual arrays?

lgray:
Working on it... might be able to work something out... Wasn't intending this to be NanoEvents specific at this stage.
May have to deal with two caches for a little bit.

lgray:
OK - final design is like this:

Switch everything to lazy evaluation.

1) cache JECs and JER(SF)s if available (these see heavy re-use and there's 7-8 columns to store at most)
2) For JEC Uncertainties since they are largely used *once* these will not be cached unless you opt them in

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -0,0 +1,370 @@
...
+
+def rand_gauss_ak1(arr, var):
+    item = arr[var]
+
+    def getfunction(layout, depth):
+        if (isinstance(layout, awkward1.layout.NumpyArray)
+            or not isinstance(layout, (awkward1.layout.Content,
+                                       awkward1.partition.PartitionedArray)
+           )):
+            return lambda: awkward1.layout.NumpyArray(np.random.normal(size=awkward1.count(item)))
```
nsmith-: `size=len(layout)` in this context, since the length of the input may be longer due to None-padding

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -0,0 +1,370 @@
...
+                                                      jetPt.JaggedArray,
+                                                      (jetPt, ))
+            pt_gen = wrap(np.zeros_like(arrays[0]))
+    elif isinstance(jetPt, awkward1.highlevel.Array):
+        def getfunction(layout, depth):
+            if (isinstance(layout, awkward1.layout.NumpyArray)
+                or not isinstance(layout, (awkward1.layout.Content,
+                                           awkward1.partition.PartitionedArray)
+               )):
+                return lambda: awkward1.layout.NumpyArray(np.zeros_like(size=awkward1.count(jetPt)))
```
nsmith-: Same story here, but really this `zeroes_like/random_like` business should be a helper function

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -0,0 +1,370 @@
...
+    min_jet_pt = _MIN_JET_ENERGY / np.cosh(arr[etaJet])
+    min_jet_pt_corr = min_jet_pt / jetPt
+
+    smearfact = None
+    if isinstance(arr, awkward.array.base.AwkwardArray):
+        wrap, arrays = awkward.util.unwrap_jagged(jetPt,
+                                                  jetPt.JaggedArray,
+                                                  (jetPt, min_jet_pt_corr))
+        smearfact = np.where(doHybrid.content, detSmear.content, stochSmear.content)
+        smearfact = np.where((smearfact * arrays[0]) < arrays[1], arrays[1], smearfact)
```
nsmith-, line 108: Why not use `ak.where` and avoid all the fragile wrapping-unwrapping?

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -0,0 +1,370 @@
+from .JECStack import JECStack
```
nsmith-: I think we should do absolute imports

coffea/jetmet_tools/FactorizedJetCorrector.py
```diff
@@ -2,8 +2,10 @@
 import warnings
 import re
 from ..util import awkward
+from ..util import awkward1
```
nsmith-: I think we can drop the `from util import X` now that the whole `cupy/numpy` switch business is done in another way.

coffea/jetmet_tools/JECStack.py
```diff
@@ -0,0 +1,113 @@
+from .FactorizedJetCorrector import FactorizedJetCorrector, _levelre
```
nsmith-: absolute import

coffea/jetmet_tools/JetCorrectionUncertainty.py
```diff
@@ -150,11 +151,19 @@ def getUncertainty(self, **kwargs):
...
-                args.append(kwargs[input])
-            uncs.append(func(*tuple(args)))
+            args = tuple(kwargs[input] for input in sig)
+
+            if isinstance(args[0], awkward.array.base.AwkwardArray):
+                uncs.append(awkward.VirtualArray(func, args=args))
+            elif isinstance(args[0], np.ndarray):
+                uncs.append(func(*args))  # np is non-lazy
+            elif isinstance(args[0], awkward1.highlevel.Array):
+                uncs.append(awkward1.virtual(func, args=args, length=len(args[0]), cache=cache))
```
nsmith-, line 165: You should provide a `form` to ak1 virtual so that it knows what is in the content. In this case, it is just a number right? You can use a simple dtype as the form argument in that case

nsmith- (changes requested):

coffea/jetmet_tools/CorrectedJetsFactory.py
lgray: there is no ak.where for awkward0? 

Anyway - making it so everything is flat before we start monkeying around with arrays so this should change.

coffea/jetmet_tools/JetCorrectionUncertainty.py
lgray: only len is needed here.

lgray:
OK - decided to try and make things more correctly integrated. Doesn't work for ak0 yet.. May never?

Anyway now, all uncertainty variations are accessed as thought originally.
Everything is now held in Jets array, all evaluation is as lazy as possible.

```
# <somewhere earlier in the code
corrector = CorrectedJets(<map to standardized jet/met property names> , 
                          <stack of JECs>) # FYI Jet/MET inputs can be non-NanoEvents

# <in actual analysis code>
cJets = corrector.build(events.Jet)
cJets.pt # central value, corrected
cJets['syst_name'].up.pt # pt shifted up for a given systematic on JES or JERSF
cJets['syst_name'].down.pt # pt shifted down '''                            '''
#cJets['syst_name'].up.shift # access to bare correction (not yet) but not impossible
```

In particular, `cJets['whatever'].up` is a fully fledged nanoevents Jet object, so you can run the same code, etc on it.

Right now it takes at 2.6s to setup a full stack of JECs and uncertainties, which is a bit slow.

Now to do the MET.

lgray:
@bfis @NJManganelli either one of you want to give this a quick spin and lemme know how it feels? I'm more or less happy with the general workflow now. You can get an idea of how to use it from the testing function.

The one thing that garnered complaints that I had to keep from the old implementation: you need to broadcast `event.<whatever rho you're using>` into your jets and have it as a per-jet object. Trying to hide that particular part of the process made things messy.

At least with awkward1 that should be an indexing operation and less of a memory hog.

pfackeldey:
Hi @lgray ,

Thank you for taking care of a new implementation! I have a small wish if you don't mind:

So far all systematic shifts, produced by the `Weight` object are suffixed with "Up" and "Down". It would be nice if this could be also the case for the JES/JER/Unclustered shifts. Especially since the Combine tool looks only for the "Up" and "Down" suffix. Right now JES/JER/Unclustered shifts have as suffix "_up" and "_down", which means that they need to get a special treatment throughout the analysis chain. I guess it is just a small change, but it simplifies the handling with e.g. fitting in the end.

Does that make sense to you?

Best, Peter

bfis:
> [...] either one of you want to give this a quick spin and lemme know how it feels?

I could take a look, but an actual test would take some time - since we are still using an [older coffea](https://github.com/bfis/coffea/tree/usable_fixes) version (due to some memory leak) and awkward0 (with the corresponding older `nanoaod/nanoevents.py`).

> In particular, `cJets['whatever'].up` is a fully fledged nanoevents Jet object, so you can run the same code, etc on it.

This is pretty much what I imagined, so I'd be very happy.

> Right now it takes at 2.6s to setup a full stack of JECs and uncertainties, which is a bit slow.

Is this for the setup of the corrector instances i.e. would not apply if one were to just (un)pickle them during the main workload?
Or is this a per event or per JEC uncertainty overhead?

> The one thing that garnered complaints that I had to keep from the old implementation: you need to broadcast `event.<whatever rho you're using>` into your jets and have it as a per-jet object. Trying to hide that particular part of the process made things messy.

I'm not sure whether I understand this correctly, but couldn't you just pass this along during corrected jet building (e.g. `cJet = corrector.build(events.Jet, event.whatever_rho_to_use)` or lies the issue with the internal broadcasting mechanics?

> So far all systematic shifts, produced by the `Weight` object are suffixed with "Up" and "Down". It would be nice if this could be also the case for the JES/JER/Unclustered shifts. [...] Right now JES/JER/Unclustered shifts have as suffix "_up" and "_down", [...]

As far as I understand the changes, the corrector would not perform concatenation of shift name and direction anymore. So we would need to do it on our own anyway, thus we can avoid the inconsistency from the get-go.

lgray:
@pfackeldey So now the JES/JER uncertainties are treated a bit more object-like, so I think the naming conventions are up to the user.

I'm thinking the pattern by which people will use this is something along the lines of:
```python3
for uncertainty in corrector.uncertainties:
    for variation in ['up', 'down']:
        weights[packed_name(uncertainty, variation)] = some_processing_code(cJets[uncertainty][variation])
```

Does this match reality or do I need to change assumptions?

pfackeldey:
Alright, perfect :) Thank you for the clarification. This looks perfect to me 👍

lgray:
I'm going to wait for further information concerning: https://github.com/scikit-hep/awkward-1.0/issues/538
before finishing this. 

Right now I don't think anyone would use this tool since setting it up is far slower than most analysis, better to wait and know what's the right way to implement.

jpivarski:
There won't be any changes in your implementation. The fixes I'm putting in to Awkward would keep the interface the same.

Since `Content::setitem_field` is not at fault, fixing this performance bug won't involve any restructuring that might have forced an interface change. So far, it looks like it's silly stuff that shouldn't have been this slow to begin with.

lgray:
@jpivarski OK - that's good to know! I think I'll still wait so people don't get a bad taste in their mouths when they try to use the tool, or put in a warning about versions.

lgray:
OK - finally the MET corrector. 

@nsmith- anything you'd like altered in the implementation?

lgray:
@bfis @pfackeldey please have a look at this once it's merged.

All the parts are there now, it'll need a shake down though.

kondratyevd:
@lgray Does this new implementation allow to apply JER without specifying JEC? (https://github.com/CoffeaTeam/coffea/issues/327)

lgray:
Ah, that can be added without too much trouble.

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def uncertainties(self):
+        out = ['JER'] if self.jec_stack.jer is not None else []
+        if self.jec_stack.junc is not None:
+            out.extend(['JES_{0}'.format(unc) for unc in self.jec_stack.junc.levels])
+        return out
+
+    def build(self, jets, lazy_cache):
+        if lazy_cache is None:
+            raise Exception('CorrectedJetsFactory requires a awkward-array cache to function correctly.')
```
nsmith-, line 178: For NanoEvents, one can likely assume that `jets._caches[0]` is a good lazy cache. (e.g. check `list(events.Jet._caches[0].keys())` out) I realize now that `array.cache` disappeared as a public attribute, so now this is private. Another option is to check for `jets.behavior["__events_factory__"]._cache` which is also private but coffea-private rather than awkward-private.

coffea/jetmet_tools/CorrectedMETFactory.py
```diff
@@ -0,0 +1,161 @@
...
+        if 'yMETRaw' not in name_map or name_map['yMETRaw'] is None:
+            warnings.warn('There is no name mapping for massRaw,'
+                          ' CorrectedJets will assume that <object>.x is raw pt!')
+            name_map['yMETRaw'] = name_map['METy'] + '_raw'
+
+        self.name_map = name_map
+
+    def build(self, MET, corrected_jets, lazy_cache):
+        if lazy_cache is None:
+            raise Exception('CorrectedMETFactory requires a awkward-array cache to function correctly.')
```
nsmith-, line 27: Same option here to use `jets.behavior["__events_factory__"]._cache` if the key exists.

tests/test_lookup_tools.py
```diff
@@ -36,30 +40,38 @@ def test_extractor_exceptions():
...
-        assert(e.args[0] == 'do not mix JaggedArrays and numpy arrays when calling a derived class of lookup_base')
+        if awkwardlib == "ak0":
+            assert(e.args[0] == 'Do not mix JaggedArrays and other arrays when calling derived class of lookup_base')
+        if awkwardlib == "ak1":
+            assert(isinstance(e, ValueError))
 
-def test_evaluator_exceptions():
+def test_evaluate_noimpl():
     from coffea.lookup_tools.lookup_base import lookup_base
     try:
```
nsmith-, line 68: `with pytest.raises(NotImplementedError)` might work too. Not sure what it tests though?

nsmith- (approved):
Some optional items, but I think any further checks would require playing with it. (which I have to do anyway for analysis)

coffea/jetmet_tools/CorrectedJetsFactory.py
nsmith-: The magic `__events_factory__` key is used by NanoEvents mixins to be able to resolve cross-references even from sliced arrays or records, so it will stay around for a while. In most cases its `_cache` attribute is the same runtime cache passed to `ak.from_arrayset`

lgray:
I'll finish the last to-do item tomorrow.

tests/test_jetmet_tools.py
```diff
@@ -252,27 +343,113 @@ def test_jet_transformer():
...
+    name_map = jec_stack.blank_name_map
+    name_map['JetPt'] = 'pt'
+    name_map['JetMass'] = 'mass'
+    name_map['JetEta'] = 'eta'
+    name_map['JetA'] = 'area'
+    
+    jets = events.Jet
+    
+    jets['pt_raw'] = jets['rawFactor'] * jets['pt']
+    jets['mass_raw'] = jets['rawFactor'] * jets['mass']
```
kondratyevd, line 379: Shouldn't it be `(1 - jets['rawFactor'])` instead?
I'm getting incorrect results otherwise.

andrzejnovak: Yes, it should

lgray: My bad. Can you post a PR and I'll get it integrated, do a backport for the tests.

kondratyevd: https://github.com/CoffeaTeam/coffea/pull/387

PR #365, opened 2020-11-04
fix: protect for "compound" name aka Jet and JetPF

andrzejnovak:
Minor bugfix
- NanoAODSchema indexing will crash if any collection names are a subset of others like `Jet` and `JetPFCands`


andrzejnovak:
@nsmith- The test doesn't look related and this PR actually allows to read the PFNano in and not crash.

PR #368, opened 2020-11-11
Executor to work with unbinned outputs (Pandas and Dask DataFrames)

kondratyevd:
This is the initial implementation of an executor that allows writing **unbinned outputs** of individual jobs in `pandas.DataFrame` format (as opposed to `AccumulatorABC`), and consider them as partitions of a single distributed Dask DataFrame. 

The output is lazy (`dask.Delayed` object referring to the partitioned Dask DataFrame). It can be passed to the next steps of the workflow, or saved as `parquet` files. In all cases, there is no need for merging, therefore no data transfer happens between individual workers, which reduces overhead and allows to keep the workflow parallelized.

The implementation is done via an executor argument `use_dataframes`. When set to `True`:
 
- The output format is different (Dask DataFrame instead of AccumulatorABC)
- "Reduce" step is not performed for the processing tasks. Instead, the outputs of individual jobs are collected into a Dask DataFrame using `dask.dataframe.from_delayed()`
- Not saving metrics: the output contains only the DataFrame


kondratyevd:
Ubuntu abd MacOS checks terminated with `Error: The operation was canceled`, I'm not sure why. @lgray is there a way to restart them?

lgray:
@kondratyevd Your commits appear to be missing the new function `run_uproot_pandas_job`. Could you please commit those changes as well?

Thanks!

kondratyevd:
@lgray it was a typo in function name from the previous version. Fixed now!

lgray:
@nsmith- anything else we want here aside from the functionality?

coffea/processor/executor.py
```diff
@@ -1093,6 +1120,7 @@ def run_uproot_job(fileset,
     mmap = executor_args.pop('mmap', False)
     schema = executor_args.pop('schema', None)
     nano = executor_args.pop('nano', False)
+    use_dataframes = executor_args.pop('use_dataframes', False)
```
nsmith-, line 1131: Can you add a small section to the docstring describing this feature? Also this would be a good spot to check if the executor is `dask_executor` since that is the only executor that supports this option

nsmith- (changes requested):

coffea/processor/executor.py
kondratyevd: Sure!

nsmith- (approved):

PR #371, opened 2020-11-23
Integrate Coffea executor with SkyhookDM backend

JayjeetAtGithub:
Add a factory method `from_arrow` in the NanoEvents module to generate NanoEvents from Arrow IPC Tables.
Also, add support for reading RADOS objects from a SkyhookDM cluster from inside executors.
Tables in IPC format are read from SkyhookDM and are converted to NanoEvents to make them ready for further
processing from within executors.

### DIFF
--- a/coffea/nanoaod/nanoevents.py
+++ b/coffea/nanoaod/nanoevents.py
@@ -1,6 +1,8 @@
 import numpy
 import uproot
 import awkward
+import awkward1
+import pyarrow
 from .methods import collection_methods
 from .util import _mixin
 from ..util import _hex, _ascii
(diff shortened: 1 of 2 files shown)


nsmith-:
FYI the `nanoaod` submodule is the awkward 0.x version and will be deprecated soon (coffea 0.7)
We should probably start on an integration using `nanoevents` submodule.

JayjeetAtGithub:
Hi @nsmith- , Thanks for pointing that out. We will change it. Btw, sorry for opening and closing the PR. We are still working on it in our fork and hope to open it soon.

PR #373, opened 2020-11-26
Add entrystart and entrystop to events metadata

jrueb:
As discussed on the mailing list, here are my suggested additions.
This just adds the `entrystart` and `entrystop` numbers to the events metadata, which can be very useful for debugging.
I noticed at this point one might also add `treename` and `fileuuid`, as these are the only fields of `WorkItem` that would still be missing from the metadata. Do you think these should be added?
Jim Pivarski suggested using uproot4 Report object. I think it would be nice to have, however the scope is missing a reference to the tree and it doesn't know the global offset. Plus a tree object isn't picklable, which I could imagine to produce problems here.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -791,15 +791,21 @@ def _work_function(item, processor_instance, flatten=False, savemetrics=False,
                     timeout=xrootdtimeout,
                     file_handler=uproot4.MemmapSource if mmap else uproot4.MultithreadedFileSource,
                 )
+            metadata = {
+                'dataset': item.dataset,
+                'filename': item.filename,
+                'treename': item.treename,
+                'entrystart': item.entrystart,
+                'entrystop': item.entrystop,
+                'fileuuid': list(item.fileuuid)  # list to be JSON serializable
+            }
             with filecontext as file:
                 if schema is None:
                     # To deprecate
                     tree = file[item.treename]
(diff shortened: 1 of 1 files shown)


lgray:
@jrueb please go ahead and add `treename` and `fileuuid`, it's not like a few bytes more of useful information is a bad thing.

As for the uproot4 report object - if there's anything extra we can extract from it on error, etc. we should try. However using it more than that locks us into a particular reader of a particular file format. This is not the best idea for some longer term goals.

jrueb:
It seems the metadata needs to be JSON serializable in order to work with awkward's `from_arrayset`. Thus I included the `fileuuid` as a list instead of bytes.

lgray:
can't do `decode('ascii')` and serialize that string?

jrueb:
`fileuuid` is not an ASCII string, nor an UTF-8 one. To me it looks like random bytes.

lgray:
Ah, then it should be convertible to a python integer, which you can then stuff in a string.
`x = int.from_bytes(fileuuid)`

Base of the number up to you, I'd prefer hex, but it's not a big deal.

jrueb:
This is probably the most rigorous way to convert it (using `uuid`). `from_bytes` asks to specify the endianness. `uuid` already has everything implemented (including putting hyphens where they belong).

lgray:
Even better! Thanks!

PR #381, opened 2020-12-10
Refactor of NanoEvents file ingestion and add arrow via parquet

lgray:
Largely it looks like it's pretty easy to make anything look roughly a root file interpreted by uproot with a few little shim classes.
The present objective is to get `NanoEventsFactory.from_file` to ingest parquet files.

This means there's one big BaseSourceMapping which handles the DSL and file handle caching.
Right now I've got NanoEvents working on root files again with all the parquet versions written and needing to be tested properly.

Given that we have multiple ways to extract forms depending on the file type, the `_extract_base_form` class method is now part of specialized source mapping class.

Arrow from parquet files can give some weird answers right now because of UnmaskedArray and stuff like that coming from of nullable, but non-masked arrow arrays.
We'll have to work out a few of those issues in awkward1 as time goes by, but for now we can still get to the point of testing.
I think for instance if `some_arrow_array.null_count == 0` then we can safely discard any notion of masking, however we may instead way to think carefully upon how we write the schemas for our parquet files and arrow tables (which would be more robust in the long term).

Things to do:
- [x] Thing that determines file type based on file magic-bytes
- [x] Refactor SourceMappings and Openers
- [x] Get uproot readers working again
- [x] Write parquet mappings and openers
- [x] Test parquet mappings and openers

@nsmith- @BenGalewsky @jpivarski


lgray:
@jpivarski It looks like the result of this will be some further discussions over how to interpret pyarrow into awkward array and back again. We need to be mindful of fields and types and what `mask == None` means in various contexts and how that translated to awkward types to keep things clear and to have well defined round trips of awkward -> arrow -> awkward, and arrow -> awkward -> arrow.

@BenGalewsky - this PR can read parquet files that look like NanoAOD (and other things too if you use the base schema). It *really* *only* reads NanoAOD, so flat arrays and singly-jagged arrays. In order to get anything more rich we have to address what I wrote to Jim above. Similarly you will have to be very careful about how you write arrow_schemas if you want the correct awkward to pop out in the future (since arrow array structure does not necessarily imply a schema, but if you have a schema that does imply an exact structure).

So we get to have some long drawn out discussion about everybody's favorite subject: interfaces and standards. 🥳

lgray:
@nsmith- this is ready for a review.

lgray:
I've also included some proposals for how decorate parquet files to aid in the construction of user friendly parts NanoEvents and to encode awkward forms in parquet if we wanna do that.

coffea/nanoevents/factory.py
```diff
@@ -33,7 +40,7 @@ def __setstate__(self, state):
         self._events = lambda: None
 
     @classmethod
-    def from_file(
+    def from_root(
```
nsmith-, line 108: We should have a stub `from_file` that calls `from_root` and deprecate it for 0.8

coffea/nanoevents/factory.py
```diff
@@ -87,71 +92,153 @@ def from_file(
...
+            raise TypeError("Invalid file type (%s)" % (str(type(file))))
+
+        if entry_start is None or entry_start < 0:
+            entry_start = 0
+        if entry_stop is None or entry_stop > table_file.metadata.num_rows:
+            entry_stop = table_file.metadata.num_rows
+
+        pqmeta = table_file.schema_arrow.metadata
+        pquuid = None if pqmeta is None else pqmeta.get(b'uuid', None)
+        pqurl = str(None if pqmeta is None else pqmeta.get(b'url', None))
```
nsmith-: Is this a relative path within the file or does it also include the file itself? If so, that should be stripped, and uuidpfn mapping should allow retrieval of the PFN. The point is that, if these PFNs were user-supplied, an earlier step could resolve the UUID and perhaps use a different PFN (e.g. the file is hosted locally)

coffea/nanoevents/factory.py
```diff
@@ -87,71 +92,153 @@ def from_file(
...
+
+        partition_meta = (
```
nsmith-: The idea is that the partition key, along with a uuidpfn map, can be used all by itself to locate and load the data. Obviously in a `from_file` situation that is trivial, but the idea is that a database holds only the partition keys.

coffea/nanoevents/factory.py
```diff
@@ -87,71 +92,153 @@ def from_file(
...
+
+        return cls._from_mapping(mapping, partition_meta, base_form,
+                                 runtime_cache, persistent_cache,
+                                 schemaclass, metadata)
+
+    @classmethod
+    def _from_mapping(
+        cls,
+        mapping,
+        partition_meta,
```
nsmith-: I think we should call this a `partition_key`

coffea/nanoevents/factory.py
```diff
@@ -87,71 +92,153 @@ def from_file(
...
+        base_form,
+        runtime_cache,
+        persistent_cache,
+        schemaclass,
+        metadata,
+    ):
+        """Quickly build NanoEvents from a root file
+
+        Parameters
+        ----------
```
nsmith-, line 294: base_form is missing from the docstring

coffea/nanoevents/mapping.py
```diff
@@ -119,6 +176,212 @@ def __getitem__(self, key):
...
+    def preload_column_source(self, uuid, path_in_source, source):
+        """To save a double-open when using NanoEventsFactory.from_file"""
+        key = self.key_root() + tuple_to_key((uuid, path_in_source))
+        self._cache[key] = source
+
+    def get_column_handle(self, columnsource, name):
+        return ParquetSourceMapping.UprootLikeShim(columnsource, name)
+
+    def extract_column(self, columnhandle, start, stop):
+        # make sure uproot is single-core since our calling context might not be
```
nsmith-: stray comment

tests/test_nanoevents.py
```diff
@@ -42,11 +46,15 @@ def test_read_nanomc():
     crossref(events[ak.num(events.Jet) > 2])
     crossref(events)
 
-    assert ak.any(events.Photon.isTight, axis=1).tolist()[:9] == [False, True, True, True, False, False, False, False, False]
```
nsmith-, line 45: Why does the result depend on the file type?

coffea/nanoevents/mapping.py
```diff
@@ -21,7 +36,42 @@ def open_uuid(self, uuid):
...
+        pquuid = None if pqmeta is None else pqmeta.get(b'uuid', None)
+        pfn = None if pqmeta is None else pqmeta.get(b'url', None)
+        if str(pquuid) != uuid:
+            raise RuntimeError(
+                f"UUID of file {pfn} does not match expected value ({uuid})"
+            )
+        return TrivialParquetOpener.UprootLikeShim(parfile)
+
+
+class BaseSourceMapping(Mapping):
```
nsmith-, line 74: So I see you abstracted by subclassing, I was originally thinking about nesting: everything but `!load` would be in a "DSL interpreter" class and it would be constructed with the source map class (which only implements the `!load` via getitem) We can save that for another time though: my motivation was to implement a skim layer (basically the form part of the persistent key would start with a `!filter` expression, referencing another column that was joined with the existing source using a "split-source" mapping beneath the dsl interpreter layer)

nsmith- (changes requested):

setup.py
```diff
@@ -60,6 +60,7 @@ def get_description():
                     'uproot>=3.11.0,<4',
                     'awkward1>=0.4.3,<1',
                     'uproot4>=0.0.17,<1',
+                    'pyarrow>=0.10.0,!=0.14.0',
```
nsmith-: pyarrow is up to 2.0.0 now, are we confident this requirement can be as loose as it is?

jpivarski: Awkward's requirement on pyarrow is `>=1.0.0` in `awkward1==0.4.3` and `>=2.0.0` in `awkward1==0.4.4`:

https://github.com/scikit-hep/awkward-1.0/commit/d92f75291906b5dc29447a9c6d5b06dbe0b1efa2

Although it's an optional dependency, only loaded at runtime if used, the `ak.to_arrow`/`ak.from_arrow` functions fail if not 2.0 (now). The same PR (above) fixes some errors in UnionArray handling, probably not directly relevant for NanoEvents.

coffea/nanoevents/mapping.py
lgray: Reminds me I need to turn the parquet multi-core read off by default!

tests/test_nanoevents.py
lgray: We didn't remake the nano_dy.parquet from the newer nano_dy.root.

coffea/nanoevents/factory.py
lgray: Right now it's actually NoneType.

This is here right now mostly as a placeholder for what we actually want to do, with some indication as to what is possible and similar to a root file. We have not agreed as to what we will fill parquet/arrow metadata with when it comes from serviceX and what its format will be (and our test file does not have this metadata anyway).

coffea/nanoevents/mapping.py
lgray: Seems reasonable and not hard - can do it in a later PR since this is future functionality.

setup.py
lgray: I'll push it up to 1.0.0 for coffea.

tests/test_nanoevents.py
nsmith-: Ok, I guess doing so might break other tests

nsmith- (approved):

PR #383, opened 2020-12-14
Deprecation warnings for awkward 0

lgray:
With a target date for changing to errors ~January 2021
The warnings can be flipped to errors with the `awkward1.deprecations_as_errors` flag.

### DIFF
--- a/coffea/processor/dataframe.py
+++ b/coffea/processor/dataframe.py
@@ -1,6 +1,7 @@
 from collections.abc import MutableMapping
 import awkward1
 import uproot4
+from coffea.util import deprecate_detected_awkward0
 
 
 class LazyDataFrame(MutableMapping):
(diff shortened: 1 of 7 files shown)


coffea/processor/dataframe.py
```diff
@@ -57,6 +58,7 @@ def __getitem__(self, key):
             if self._flatten and isinstance(awkward1.type(array).type, awkward1.types.ListType):
                 array = awkward1.flatten(array)
             array = awkward1.to_awkward0(array)
```
nsmith-, line 60: Actually LazyDataFrame could be made awkward1-compatible via opt-in flag that just makes it skip this line. In the long run, though, maybe it makes more sense to support just NanoEvents via BaseSchema

nsmith- (approved):

coffea/processor/dataframe.py
lgray: Yep - that's why the support warning rather than the will be deleted one. `0.7` here we come.

PR #386, opened 2020-12-15
Remove use of awkward0 in all coffea classes, minimize use of uproot3

### LINKED ISSUE #382: Disable cross-reference errors in NanoAODSchema
Sometimes, due to a skim step, certain collections are dropped but not the cross-ref branches. This makes it impossible to open with NanoEvents using the default NanoAODSchema.

Set a class member option to turn https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/schemas.py#L225 into a warning.

**Describe alternatives you've considered**
A workaround is to drop the offending cross-ref using a new schema class:
```python

class HackSchema(NanoAODSchema):
    def __init__(self, base_form):
        base_form["contents"].pop("Muon_fsrPhotonIdx", None)
        super().__init__(base_form)


events = NanoEventsFactory.from_file(
    "tests/samples/nano_dy.root",
    schemaclass=HackSchema,
).events()
```

@dryrun

### LINKED ISSUE #385: Add awkward1 support to spark executor
Starting from #355 there was a request to have the spark executor be able to use awkward1.

This requires adding the appropriate functionality to the UDFs to produce awkward1 arrays, similar to what's there for awkward0 right now.

### LINKED ISSUE #395: Bug in delta_r coffea.0.6.47
**Describe the bug**
delta_r does not work if the second array is also nested. 

**To Reproduce**
```
events = NanoEventsFactory.from_file("root://cmseos.fnal.gov//store/user/cmsdas/2021/long_exercises/TTGamma/TestFiles/TTGamma_1l.root", schemaclass=NanoAODSchema).events()
events.Muon.delta_r(events.Photon)
```

**Output**
Error message:
```
ValueError                                Traceback (most recent call last)
<ipython-input-127-c220807612a8> in <module>
----> 1 events.Muon.delta_r(events.Photon)

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/coffea/nanoevents/methods/vector.py in delta_r(self, other)
    416         :math:`\sqrt{\Delta\eta^2 + \Delta\phi^2}`
    417         """
--> 418         return numpy.sqrt(self.delta_r2(other))
    419 
    420     def nearest(

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/coffea/nanoevents/methods/vector.py in delta_r2(self, other)
    409     def delta_r2(self, other):
    410         """Squared `delta_r`"""
--> 411         return (self.eta - other.eta) ** 2 + self.delta_phi(other) ** 2
    412 
    413     def delta_r(self, other):

/cvmfs/sft.cern.ch/lcg/views/LCG_96python3/x86_64-centos7-gcc8-opt/lib/python3.6/site-packages/numpy/lib/mixins.py in func(self, other)
     23         if _disables_array_ufunc(other):
     24             return NotImplemented
---> 25         return ufunc(self, other)
     26     func.__name__ = '__{}__'.format(name)
     27     return func

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/awkward1/highlevel.py in __array_ufunc__(self, ufunc, method, *inputs, **kwargs)
   1367         See also #__array_function__.
   1368         """
-> 1369         return awkward1._connect._numpy.array_ufunc(ufunc, method, inputs, kwargs)
   1370 
   1371     def __array_function__(self, func, types, args, kwargs):

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/awkward1/_connect/_numpy.py in array_ufunc(ufunc, method, inputs, kwargs)
    213 
    214     out = awkward1._util.broadcast_and_apply(
--> 215         inputs, getfunction, behavior, allow_records=False
    216     )
    217     assert isinstance(out, tuple) and len(out) == 1

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/awkward1/_util.py in broadcast_and_apply(inputs, getfunction, behavior, allow_records, left_broadcast, right_broadcast, numpy_to_regular)
   1008     else:
   1009         isscalar = []
-> 1010         out = apply(broadcast_pack(inputs, isscalar), 0)
   1011         assert isinstance(out, tuple)
   1012         return tuple(broadcast_unpack(x, isscalar) for x in out)

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/awkward1/_util.py in apply(inputs, depth)
    771                         nextinputs.append(x)
    772 
--> 773                 outcontent = apply(nextinputs, depth + 1)
    774                 assert isinstance(outcontent, tuple)
    775 

/uscms_data/d3/ahall/DAS2021/fork/TTGamma_LongExercise/ttgenv/lib/python3.6/site-packages/awkward1/_util.py in apply(inputs, depth)
    812                         nextinputs.append(fcn(x, offsets))
    813                     elif isinstance(x, listtypes):
--> 814                         nextinputs.append(x.broadcast_tooffsets64(offsets).content)
    815                     # handle implicit left-broadcasting (unlike NumPy)
    816                     elif left_broadcast and isinstance(x, awkward1.layout.Content):

ValueError: in ListOffsetArray64, cannot broadcast nested list

(https://github.com/scikit-hep/awkward-1.0/blob/0.4.5/src/cpu-kernels/awkward_ListArray_broadcast_tooffsets.cpp#L27)`
```
 - OS: SL7 on the LPC
 - Version 0.6.47

**Additional context**
My temporary workaround is to define my own method using ak.cartesian:
```
        def dR(one,two):
            pairs = ak.cartesian({"i0":one,"i1":two},nested=True)
            diffPhi = abs(pairs.i0.phi - pairs.i1.phi)
            diffEta = abs(pairs.i0.eta - pairs.i1.eta)
            return np.sqrt(diffPhi*diffPhi + diffEta*diffEta)
```

lgray:
Fixes #385 
Fixes #382
Fixes #395

uproot3 is still needed for one thing - export1d in the hist package, since uproot4 cannot write root files.

I've made `import coffea` a relatively empty import except for a new deprecation warnings-as-errors.

Another major change is that the original dataframe shim interface is completely removed, replaced by at least `nanoevents.BaseSchema`, which helps a lot in working with awkward1. The "flatten" option is now removed, since `ak.zip` is great.

NanoEvents now uses `ak.from_buffers` rather than `ak.from_arrayset` since the latter will be removed.

All executors function as they did before, and emit errors if you try to use awkward0.

All tests now function, no longer pertinent tests are removed.


lgray:
@nsmith- this is ready for a look.

lgray:
Fixes #385

lgray:
Could also include #382 in this.

coffea/hist/__init__.py
```diff
@@ -17,9 +17,6 @@
     Cat,
     StringBin,
 )
-from .export import (
```
nsmith-, line 20: Is this because uproot3 is an optional install dependency? Otherwise it seems we could leave this in. If it needs removing, we need to update the tutorial notebook on histograms, which demos this feature.

coffea/hist/hist_tools.py
```diff
@@ -9,7 +9,7 @@
 import numbers
 import re
 import warnings
-import awkward1
+import awkward as ak
```
nsmith-: Here and throughout, I personally prefer using `awkward` in this library rather than shorthand (same goes for `numpy` vs. `np`). I think it is generally a good idea for library code to be a bit more verbose. I'd invite @jpivarski to comment and break the tie since I know you must have made this decision to switch in a lot of places.

coffea/jetmet_tools/CorrectedMETFactory.py
```diff
@@ -25,12 +24,9 @@ def __init__(self, name_map):
...
             raise Exception('CorrectedMETFactory requires a awkward-array cache to function correctly.')
-        if (isinstance(MET, awkward.array.base.AwkwardArray) or
-            isinstance(corrected_jets, awkward.array.base.AwkwardArray)):
-            raise Exception('Awkward0 not supported by CorrectedMETFactory, please use the JetTransformer or awkward1')
-        elif (not isinstance(MET, awkward1.highlevel.Array) or
-              not isinstance(corrected_jets, awkward1.highlevel.Array)):
-            raise Exception('\'MET\' must be an awkward array of some kind!')
+        if (not isinstance(MET, ak.highlevel.Array) or
+            not isinstance(corrected_jets, ak.highlevel.Array)):
+            raise Exception('\'MET\' and \'corrected_jets\' must be an awkward array of some kind!')
```
nsmith-: Double quotes on the outside? (it is what black uses)

coffea/jetmet_tools/JetCorrectionUncertainty.py
```diff
@@ -1,8 +1,7 @@
 from ..lookup_tools.jec_uncertainty_lookup import jec_uncertainty_lookup
 import warnings
 import re
-from ..util import awkward
-from ..util import awkward1
+import awkward as ak
 from ..util import numpy as np
```
nsmith-: This can be just `import numpy as np` (or maybe `import numpy` per previous comments) right? I believe that if we do in the future try for `cupy` et al we'll have an alternative implementation (awkward1 has approached this differently, deciding the library in the dunder method callback)

coffea/lookup_tools/jersf_lookup.py
```diff
@@ -98,9 +98,9 @@ def _evaluate(self, *args):
             eval_values.append(np.clip(eval_vals[eval_name], clamp_mins, clamp_maxs))
 
         # get parameter values
-        parm_values_central = self._parms[0][bin_tuple].flatten()
-        parm_values_up = self._parms[1][bin_tuple].flatten()
-        parm_values_down = self._parms[2][bin_tuple].flatten()
+        parm_values_central = np.array(self._parms[0][bin_tuple]).squeeze()
```
nsmith-, line 101: Looks unrelated to ak1 or the mentioned bugfixes?

coffea/lookup_tools/jme_standard_function.py
```diff
@@ -1,8 +1,8 @@
 from .lookup_base import lookup_base
 
-from ..util import awkward
-from ..util import numpy
-from ..util import numpy as np
+import awkward as ak
+import numpy
+import numpy as np
```
nsmith-: Pending previous comment, probably should pick one of the two numpy names

coffea/lookup_tools/txt_converters.py
```diff
@@ -183,8 +183,8 @@ def _build_standard_jme_lookup(
                     theBins = np.union1d(binMins, binMaxs[-1:])
                 allBins = np.append(allBins, theBins)
                 counts = np.append(counts, theBins.size)
-            bins[layout[i + offset_name]] = awkward.JaggedArray.fromcounts(
-                counts, allBins
+            bins[layout[i + offset_name]] = ak.unflatten(
```
nsmith-: Did you cross check the minimum awkward version for `unflatten` in setup.py?

coffea/nanoevents/factory.py
```diff
@@ -43,67 +44,6 @@ def __setstate__(self, state):
         self._cache = None
         self._events = lambda: None
 
-    @classmethod
```
nsmith-, line 46: Can you update the tutorial notebook(s) to use `from_root`?

coffea/nanoevents/factory.py
```diff
@@ -329,15 +344,19 @@ def events(self):
...
             behavior = dict(self._schema.behavior)
             behavior["__events_factory__"] = self
-            events = awkward1.from_arrayset(
+            events = ak.from_buffers(
                 self._schema.form,
+                len(self),
                 self._mapping,
-                prefix=self._partition_key,
-                sep="/",
+                key_format=key_formatter,
```
nsmith-, line 363: I missed this development, thanks @jpivarski ! Does this allow also control of the non-content attributes such as `offsets`? That would simplify the DSL a bit by removing the need for the `!skip` command (because otherwise awkward1 forces us to add an offsets that is sometimes not appropriate (e.g. if the offsets are coming from a different branch)

coffea/nanoevents/mapping/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .uproot import TrivialUprootOpener, UprootSourceMapping
```
nsmith-, line 1: I'd prefer if we do absolute imports

coffea/nanoevents/methods/vector.py
```diff
@@ -417,6 +417,42 @@ def delta_r(self, other):
         """
         return numpy.sqrt(self.delta_r2(other))
 
+    def metric_table(self,
+                     other,
+                     axis=1,
+                     metric=lambda a, b: a.delta_r(b),
+                     return_combos=False
```
nsmith-: I'd prefer the more verbose `return_combinations`

coffea/nanoevents/methods/vector.py
```diff
@@ -446,27 +482,22 @@ def nearest(
...
         if axis is None:
-            a, b = self, other
             # NotImplementedError: ak.firsts with axis=-1
             axis = other.layout.purelist_depth - 2
-        else:
-            a, b = awkward1.unzip(
-                awkward1.cartesian([self, other], axis=axis, nested=True)
-            )
-        mval = metric(a, b)
         # prefer keepdims=True: awkward-1.0 #434
```
nsmith-, line 486: scikit-hep/awkward-1.0#434 is resolved, so now it could be 
```python
mmin = ak.argmin(mval, axis=axis + 1, keepdims=True)
```

coffea/nanoevents/schemas/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .base import BaseSchema
```
nsmith-, line 1: I'm not sure I see the motivation in splitting this module out at this stage. The file was not that large.
Did you make any other changes than just dividing the code into files?

coffea/nanoevents/transforms.py
```diff
@@ -1,10 +1,19 @@
 import copy
 import numpy
 import numba
-import awkward1
+import awkward as ak
 from coffea.nanoevents.util import concat
 
 
+def data(stack):
```
nsmith-, line 8: What kind of array does this get called on? pyarrow?
Most of the rest are specifically for awkward layout types (we should probably label them to indicate what type of input is valid)

coffea/nanoevents/transforms.py
```diff
@@ -265,16 +274,16 @@ def nestedindex(stack):
     """
     indexers = stack[:]
     stack.clear()
-    # return awkward1.concatenate([idx[:, None] for idx in indexers], axis=1)
+    # return ak.concatenate([idx[:, None] for idx in indexers], axis=1)
```
nsmith-: I believe concatenate for axis=1 is now implemented. Care to try it?

coffea/processor/dataframe.py
```diff
@@ -13,26 +12,21 @@ class LazyDataFrame(MutableMapping):
...
-        tree : uproot4.TTree
+        tree : uproot.TTree
             Tree to read
         entrystart : int, optional
             First entry to read, default: 0
         entrystop : int, optional
             Last entry to read, default None (read to end)
         preload_items : iterable
             Force preloading of a set of columns from the tree
-        flatten : bool
```
nsmith-, line 24: I don't think flatten is used in any tutorials now, but might be worth checking

coffea/processor/executor.py
```diff
@@ -888,37 +862,21 @@ def _work_function(item, processor_instance, flatten=False, savemetrics=False,
                 'fileuuid': str(uuid.UUID(bytes=item.fileuuid))
             }
             with filecontext as file:
-                if schema is None:
-                    # To deprecate
```
nsmith-, line 892: I see LazyDataFrame is still in the source. I think it does cover some use cases where a file is just not readable via NanoEvents with base schema presently. Perhaps the better solution is to make https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/mapping.py#L195 more robust, but in the meantime any strong motivation not to leave the LazyDataFrame (via setting schema to None) ?

coffea/processor/test_items/NanoEventsProcessor.py
```diff
@@ -32,6 +33,19 @@ def process(self, events):
 
         dataset = events.metadata["dataset"]
 
+        mapping = events.behavior["__events_factory__"]._mapping
+        events.Muon.pt
+        if isinstance(mapping, nanoevents.mapping.CachedMapping):
```
nsmith-, line 38: Thanks for adding this. I was feeling lazy when I first went through

tests/test_preloaded.py
```diff
@@ -2,25 +2,29 @@
...
 
-    tree = uproot.open(os.path.abspath('tests/samples/nano_dy.root'))['Events']
-    arrays = tree.arrays(columns, flatten=True, namedecode='ascii')
-    df = processor.PreloadedDataFrame(tree.numentries, arrays)
+    rootdir = uproot.open(os.path.abspath('tests/samples/nano_dy.root'))
+    tree = rootdir['Events']
+    arrays = tree.arrays(columns, how=dict)
+    src = SimplePreloadedColumnSource(arrays,
+                                      rootdir.file.uuid,
+                                      len(arrays['nMuon']),
```
nsmith-: Could do `tree.numentries` as well right?

nsmith- (changes requested):
One general comment:`coffea/nanoevents` and `coffea/lookup_tools` had previously been black-formatted. Can you run it again for those to keep them formatted? Eventually we should get the whole repo autoformatted, but I wanted to hold off on that for a dedicated PR.

As an aside, I checked out locally and see an error in tests:
```
FAILED tests/test_jetmet_tools.py::test_corrected_jets_factory - AttributeError: no field named 'x'
```
It may be due to some unreleased change in awkward, but I will debug a bit further. (edit: https://github.com/scikit-hep/awkward-1.0/issues/629) I also see a numba typing error coming from the nanoaod transform.

coffea/hist/__init__.py
lgray: Mostly I didn't want the awkward0 deprecation warning showing up any time anyone imports hist.
I can't seem to turn of the warning being emitted.

Since it inherits from exception I guess I can catch it or something? I'll poke at it.

coffea/nanoevents/transforms.py
lgray: This shows up in from_buffers even for root files, `/data` is more or less equivalent to `!content`.

coffea/lookup_tools/jersf_lookup.py
lgray: It was getting an extra empty dimension for some reason.

coffea/hist/__init__.py
jpivarski: Can you filter it with the [warnings filter](https://docs.python.org/3/library/warnings.html#describing-warning-filters)? It's a `FutureWarning`.

jpivarski: Yes, you can do this:

```python
>>> import warnings
>>> with warnings.catch_warnings():
...     warnings.simplefilter("ignore")
...     import awkward0
... 
>>> awkward0
<module 'awkward0' from '.../python3.8/site-packages/awkward0/__init__.py'>
```

lgray: OK - so long as that doesn't set the global warning filter state and a subsequent import of awkward0 into the global namespace would cause a warning.

jpivarski: This sets the warning state only temporarily—after the `with` block, `FutureWarning` would be shown again. However, once `awkward0` is loaded, `import awkward0` is a pass-through (just because of how Python's import works), so users will never see the Awkward0 warning.

lgray: Ah, hmm. I think I'll take it since we're removing awkward0 support from the rest of coffea and we shouldn't have major issues from version mixups.

jpivarski: This warning-hiding might be relevant for importing Uproot 3, which is still needed for a couple of things. Or just leave the warnings in and explain them when they come up, or catch Uproot 3's warning and emit your own, saying, "We rely on Uproot 3 for this and that, but there's no way around that for now."

The thing a user needs to know is whether an action needs to be taken. If they're manually importing Uproot 3, the answer is often "yes" (they should upgrade if they are just starting a project), but if they're getting it as a dependency through Coffea, the answer is "no" (Coffea will upgrade underneath when it becomes possible; the user has no action to take).

nsmith-: Can we hide the `TH1` class declaration that requires importing uproot3 inside the export1d function so that uproot3 is only imported (and awkward0 along with) when the function is called?

lgray: That makes the TH1 class unpickleable (we actually have a test for this, contributed by the Aachen group).

lgray: I think I'll just go with wrapping the import, since in time this will switch to uproot4.

coffea/nanoevents/transforms.py
nsmith-: From https://github.com/scikit-hep/awkward-1.0/pull/592/files#diff-e9f2e3c299511a19f2535f6fdaaca21cb69f3893a3d7fe57aa775ef7110e5317L3410-L3418 the NumpyArray form is now keyed with `data` where it used to not be keyed with any attribute. FYI @jpivarski was this intended? This is fine, though I think the NanoEvents DSL could just as well tack `!skip` after every `!content` rather than make this no-op call (would save some tiny amount of time)
What we can also do is change `key_formatter` to ignore any attribute arguments on the assumption that the form key already fully resolves the required attribute. Then we wouldn't need `!skip` or `!data` DSL commands

coffea/nanoevents/factory.py
nsmith-: Indeed it does, per docs:
key_format (str or callable) – Python format string containing "{partition}", "{form_key}", and/or "{attribute}" or a function that takes these as keyword arguments and returns a string to use as keys for buffers in the container. The partition is a partition number (non-negative integer, passed as a string), the form_key is a string associated with each node in the Form, and the attribute is a hard-coded string representing the buffer’s function (e.g. "data", "offsets", "index").

lgray: So the problem here is that `/offsets` was generated by the old `from_arrayset` and the others are new. Dropping the attribute results in an incorrectly unpacked array. It may be better to account for the new attributes and work with @jpivarski to tune in a well defined and useful behavior.

coffea/lookup_tools/txt_converters.py
lgray: 1.0.1 is good and in setup.py

coffea/processor/executor.py
lgray: I'll make None give you LazyDataFrame, but default to BaseSchema.

lgray: Tagging this for rc2

coffea/processor/dataframe.py
lgray: it's not

coffea/nanoevents/transforms.py
lgray: tagging this for rc2

coffea/lookup_tools/txt_converters.py
lgray: This raises a question if we want awkward release candidates as pins.
Right now I have to use uproot4 rc because otherwise it reads histograms wrong.

Delaying this to rc2.

coffea/hist/hist_tools.py
lgray: I think I got most of this.

coffea/nanoevents/factory.py
lgray: done

coffea/nanoevents/factory.py
lgray: delaying this to rc2

lgray:
There are a few unresolved issues from this PR. I'll make a tracking issue or expand the 0.7.0 issue.

Closing this will result in rc1. We should aim to keep it to 2 or 3 rcs before releasing 0.7.0.

coffea/nanoevents/mapping/__init__.py
```diff
@@ -1,9 +1,18 @@
...
-           'SimplePreloadedColumnSource', 'PreloadedOpener', 'PreloadedSourceMapping'
-           'CachedMapping']
+__all__ = [
+    "TrivialUprootOpener",
+    "UprootSourceMapping",
+    "TrivialParquetOpener",
+    "ParquetSourceMapping",
+    "SimplePreloadedColumnSource",
+    "PreloadedOpener",
+    "PreloadedSourceMapping" "CachedMapping",
```
nsmith-, line 17: I think this is missing a comma, so the two strings just get concatenated

nsmith- (commented):
Guess its already merged but I did see one likely bug

coffea/nanoevents/mapping/__init__.py
lgray: weird it doesn't cause any tests to fail?

jpivarski: Perhaps none of the tests involve PreloadedSourceMapping or CachedMapping? That's a reason why this is a bad syntax rule, but it's a hold-over from C.

nsmith-: This is a great example of how a code formatter can make certain bugs more obvious, since the missing comma was previously disguised by a line break.

lgray: Yeah - it's certainly useful from that perspective. However both of those classes are tested! Anyway. I'll fix it up and start rc2.

nsmith-: Specifying `__all__` in a module is kind of like overriding `__dir__` on a class: it doesn't actually change the behavior of getattr but does provide a way to declare a (possibly restricted) list of "public" attributes

coffea/nanoevents/schemas/__init__.py
lgray: No - just thinking forward to automatic schema generators and stuff like that. If a random person wants to make their own schema it's nice to have it compartmentalized, no other distractions in the file, same for 'examples' with the TreeMaker and NanoAOD schemas.

nsmith-: I agree it needs to happen eventually, just not sure about now (also a bit sad to lose the change history for it)
But didn't you need to make a change to implement #382

nsmith-: I found it https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/schemas/nanoaod.py#L128-L135 looks fine (though `type(self)` imo looks nicer than dunder, but it is to taste)

coffea/nanoevents/methods/vector.py
nsmith-: I'll take this

PR #390, opened 2020-12-16
WQ Executor Fixups

dthain:
Several updates to improve the ergonomics of the Work Queue executor:
- The progress bar display was broken and only showed the creation of tasks.  It now shows one bar for creation, then a second bar for actual execution.
- Added options to control verbose mode, output mode, and worker password, so they don't clobber the actual output.  By default, you only get excess output when a task fails in some way.

(Some further PRs incoming to update the WQ installation and examples.)


lgray:
same for here.

dthain:
Ok, ready to merge here.

PR #391, opened 2020-12-16
Update WQ Executor Installation and Examples

dthain:


lgray:
@dthain lemme know when you're happy here and I'll merge!

dthain:
Appreciate the quick response, I think we are close, I just want to check a few more things.

We are trying to carefully match up what we are running in integration tests with the documentation and actual user practice.
I see that coffea is available via conda now, is there a reason to prefer that (or not) over pip?

lgray:
@dthain The conda version of coffea is there for people that prefer to use conda, there's no preference on our end.

dthain:
Ready to merge here.

PR #392, opened 2020-12-17
WQ Executor: Pipelining and Garbage Collection

dthain:
The existing WQ executor first materializes all tasks and then waits for them all to complete.  This works, but causes trouble if you have a very large number of tasks such that the filesystem becomes a bottleneck with all of the temp files generated.

This PR pipelines execution so that submission and execution are interleaved.  If more than 100 tasks are idle in the queue, the executor starts waiting for some to execute or complete before generating more.  Temporary files are deleted as we go along.


dthain:
Ok, this one is ready on my end.

PR #400, opened 2020-12-24
See if CI can be done binary only

lgray:


lgray:
weird, incompatible with spark. ah well.

PR #403, opened 2021-01-01
0.7.0 Release Candidate 2

lgray:
Continuation of release candidate development.

- [x] https://github.com/CoffeaTeam/coffea/pull/386#pullrequestreview-560604697
- [x] https://github.com/CoffeaTeam/coffea/pull/386#discussion_r550369516 (confirmed that unflatten min. ver. is 1.0.1)
- [x] https://github.com/CoffeaTeam/coffea/pull/386#discussion_r550370291 (deferred to 0.7.x and discussion with Jim)
- [x] https://github.com/CoffeaTeam/coffea/pull/386#discussion_r550371659
- [x] https://github.com/CoffeaTeam/coffea/pull/386#discussion_r550373063 (schema=None is LazyDataFrame, default is BaseSchema)


coffea/processor/__init__.py
```diff
@@ -5,7 +5,6 @@
 from .processor import ProcessorABC
 from .dataframe import (
     LazyDataFrame,
-    PreloadedDataFrame,
```
nsmith-, line 8: Does this mean the spark executor only supports NanoEvents? Or just that it uses some other mechanism for generic root files?

nsmith- (approved):

coffea/processor/__init__.py
lgray: Spark executor works on the internals of spark which are pandas or arrow tables, which will always be convertible to pretty simple awkward. There is little reason to support anything but nanoevents here.

Laurelin or whatever else is used as the input is completely decoupled.

nsmith-: Ah right, so the base form is always reasonable

PR #404, opened 2021-01-02
Fix bug in NanoAOD electron working point attributes

nsmith-:
Unlike for photons, electrons don't use a bitmap to pack their values,
but rather a monotone increasing sequence.


lgray:
I'll save this for rc2 unless you desperately need it for the DAS?

PR #415, opened 2021-01-13
Remove reference cycles from CorrectedJetsFactory.build 

nsmith-:
Still maintains laziness (getitem on a recordarray will not materialize virtual child nodes, rather it just returns them)

The issue is that a reference cycle like:
(highlevel array) -> (recordarray node) -> (virtual node) -> (generator) -> (args) -> (highlevel array)
can be formed quite easily, e.g.
```python
import awkward as ak

arr = ak.zip({"x": ak.Array([1, 2, 3]), "y": ak.Array([4, 5, 6])})
arr["z"] = ak.virtual(lambda arr: arr["x"] + arr["y"], (arr,), length=len(arr))
del arr

import gc
print([o for o in gc.get_objects() if "awkward.highlevel" in str(type(o))])

# make gc try harder
for _ in range(10):
    gc.collect()
print([o for o in gc.get_objects() if "awkward.highlevel" in str(type(o))])
```
which returns:
```
[<Array [{x: 1, y: 4, z: 5}, ... y: 6, z: 9}] type='3 * {"x": int64, "y": int64, ...'>]
[<Array [{x: 1, y: 4, z: 5}, ... y: 6, z: 9}] type='3 * {"x": int64, "y": int64, ...'>]
```
you can see that the garbage collector cannot seem to detect the cycle created here, probably because the generator is a pybind11 object that then holds a reference to a pure python object.

In practice this leads to runaway memory usage (4 GB in 5 minutes!) on large processing jobs. FYI @dryrun

### DIFF
--- a/coffea/jetmet_tools/CorrectedJetsFactory.py
+++ b/coffea/jetmet_tools/CorrectedJetsFactory.py
@@ -4,6 +4,7 @@ import numpy
 import warnings
 from copy import copy
 from functools import partial
+import operator
 
 
 _stack_parts = ['jec', 'junc', 'jer', 'jersf']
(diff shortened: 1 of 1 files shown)


lgray:
Thanks for the catch @nsmith-!

PR #421, opened 2021-01-15
Backport WQ Executor Changes to V0.6.X

dthain:
Combines PRs #390 #391 #392 into V0.6.X branch.


dthain:
Hmm, this commit history looks more complex than I intended.  Let me see if there is a better way to get the desired result.

lgray:
Yeah as it is this also pulls in the 0.7.0 marker! Probably just some quick scrutiny with a rebase -i will fix things up.

dthain:
See #426 instead

PR #422, opened 2021-01-15
Allow Regrouped* in JEC uncertainty name

nsmith-:


lgray:
Can you add a zipped uncertainty file with this naming scheme to the tests?

nsmith-:
Would be nice to know a regex that matches these, maybe @aperloff knows who to ask?
My initial guess would be `(Regrouped_|RegroupedV._|)(.*)_(MC|DATA)_([^_]+)_([^_]+)` but this is quickly getting out of hand. Maybe wait for json format? :)

aperloff:
I've never seen a regex to parse the filenames themselves, which is what we're talking about. For reference, I've written down the parts of the string in [1]. Until now I've never seen anyone put something before the era part of the string. I don't have any code which parses the filename. We all just built it from its components. That said, the JERC conveners and their AlCaDB contact would be in charge of this.

[1] RegroupedV2_Fall17_17Nov2017_V32_MC_UncertaintySources_AK4PFchs.txt
notation? = RegroupedV2
era = Fall17_17Nov2017_V32_MC
correction/type = UncertaintySources
jet collection = AK4PFchs

PR #425, opened 2021-01-18
fix: LumiMask to take ak

andrzejnovak:
To allow `LumiMask(json)(events.run, events.luminosityBlock)` without manually casting.

### DIFF
--- a/coffea/lumi_tools/lumi_tools.py
+++ b/coffea/lumi_tools/lumi_tools.py
@@ -1,5 +1,6 @@
 from ..util import numpy as np
 import json
+import awkward as ak
 
 from ..util import numba
 from numba import types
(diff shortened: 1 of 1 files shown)


coffea/lumi_tools/lumi_tools.py
```diff
@@ -117,6 +118,10 @@ def __call__(self, runs, lumis):
                 An array of dtype `bool` where valid (run, lumi) tuples
                 will have their corresponding entry set ``True``.
         """
+        if isinstance(runs, ak.highlevel.Array):
```
nsmith-, line 121: FYI for future: `numpy.asarray` can be used as well, avoiding the awkward import (and generalizing to other array extension types)

PR #426, opened 2021-01-19
WQ Backports to V0.6.X Branch (Take 2)

dthain:
Backport Work Queue executor improvements PRs #390 #391 #392 to V0.6.X branch.

PR #430, opened 2021-01-27
Updated all awkward/uproot-methods/uproot requirements and imports to new names.

jpivarski:
I have updated all the required packages in setup.py and all `imports` under the "coffea" and "tests" directories to use the post-transition names.

My intention is for this PR to be merged into `backports-v0.6.x` and have that released as Coffea v0.6.51. (I've also updated the version number.)

Let's see if the tests pass!

The goal would be for the Coffea package to split into green and yellow on the right of the transition line below, with Coffea v0.6.51 being yellow (old), but to the right of the line, and Coffea v0.7.0rcN to be green (new), also to the right.

transition-version-numbers-methods


jpivarski:
The tests pass now. If you don't want users to see the warnings, wrap all `import uproot3`/`import uproot3_methods`/`import awkward0` variants (highlighted by this PR diff) with

```python
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    import uproot3 as uproot
    import uproot3_methods as uproot_methods
    import awkward0 as awkward
```

(or more carefully singling out the FutureWarning [using this recipe](https://docs.python.org/3/library/warnings.html#testing-warnings)).

"test coffea (macOS-latest) - python 3.6, JDK1.8" takes a long time because Awkward 1.x doesn't have a binary wheel for Python 3.6 on MacOS. When your CI installs it, it has to compile from source. ([This deployment script](https://github.com/scikit-hep/awkward-1.0/blob/620aa92e462b8e96da00ee6ce93df8d88b84af7a/.ci/azure-deploy-awkward.yml#L266-L296) didn't work for some reason, long ago, so I gave up on that OS/Python combination.) Isn't Python 3.6 on MacOS is an unusual combination?

**Please release this as Coffea v0.6.51 soon, and encourage everyone to pip --update to the new names!**

Thanks!

lgray:
@jpivarski would you mind if we merge #426 ahead of this and rebase here to make sure there aren't any typos caused one way or another?

jpivarski:
Sure: chances are, they're mutually mergeable. I only changed imports, not the code that uses the imported libraries further down in the code, so they're localized differences.

lgray:
@jpivarski That is what I too am hoping!

nsmith-:
#426 is merged and this branch is updated. LGTM?

PR #435, opened 2021-02-01
Fix reading from CSV for numpy 1.20.0

jrueb:
In version 1.20.0 of numpy a fix to np.genfromtxt was introduced that makes it return a list if the result is not transposable and unpack is True.
See https://github.com/numpy/numpy/commit/3329d26c5d1d53f7fca1dfe253fdc43e93f0f6aa

### DIFF
--- a/coffea/lookup_tools/csv_converters.py
+++ b/coffea/lookup_tools/csv_converters.py
@@ -50,7 +50,6 @@ def convert_btag_csv_file(csvFilePath):
         },
         delimiter=",",
         skip_header=1,
-        unpack=True,
         encoding="ascii",
     )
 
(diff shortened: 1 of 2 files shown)


jrueb:
I have tested the effects of the `unpack` parameter in an older version of numpy and it doesn't seem to make a difference whether it's `True` or `False` in Coffea. The bug fix in numpy was that `unpack` had no effect if `dtype` was structured, which is the case for Coffea. I am unsure why it was set to `True` here at all.
I did not test `convert_effective_area_file` though.

lgray:
Thanks for the fix! I'll mint the release soon.

PR #444, opened 2021-02-05
Remove expectation of warning for uproot3

lgray:


nsmith-:
This seems to have been a change in uproot3 3.14.3 from 3.14.2, and I was hoping the warning would be re-introduced. @jpivarski is this expected that we no longer see a warning?
```python
    def test_export1d():
        with pytest.warns(FutureWarning):
>           import uproot3
E           Failed: DID NOT WARN. No warnings of type (<class 'FutureWarning'>,) was emitted. The list of emitted warnings is: [].
tests/test_hist_tools.py:164: Failed
```

nsmith-:
Answered in https://github.com/scikit-hep/uproot3/issues/533

PR #447, opened 2021-02-09
Extend Nanoevents vector and add tests

### LINKED ISSUE #445: Extending Nanoevents vector
(Continued from the mailing list)
The classes in `coffea.nanoevents.methods.vector` are still missing some common functions. What I have in mind here

- Subtractions, just like there is addition for two vector objects
- Division with numbers, just like there is multiplication
- Dot product
- `ThreeVector.cross`, returning the cross product with another `ThreeVector`
- `TwoVector.unit` and `ThreeVector.unit`, returning the corresponding unit
vector
- `LorentzVector.p3`, returning the corresponding momentum `ThreeVector`
- `LorentzVector.boostp3`, like `p3` but divided by the energy so that it
can be used for boosting
- `LorentzVector.boost`, returning the Lorentz boost using a `ThreeVector`

The naming here agrees with what is used in `uproot3_methods`. Besides `dot`, I can not find these in scikit-hep/vector. On the other hand one might argue that `p3` is too ambiguous as there already is `p2`, which is something different.

Note that `LorentzVector.boostp3` does what `TLorentzVector::BoostVector` in ROOT does, which is `LorentzVector.p3 / LorentzVector.t` and thus not equivalent to `LorentzVector.p3.unit`, which is `LorentzVector.p3 / LorentzVector.rho`.

jrueb:
Closes #445

Besides adding the features mentioned in #445, this also adds a `negative` mixin, allowing thngs like `b = -a`. Also added `PolarTwoVector.multiply` and `SphericalThreeVector.multiply`.
Additionally I fixed the `PtEtaPhiMLorentzVector.multiply` and `PtEtaPhiELorentzVector.multiply` functions, which gave correct results only for non-negative factors.
This also adds a number of unit tests for the vectors.
Finally I used @nsmith- suggestion for naming `pvec`, but this can still easily be adjusted.

### DIFF
--- a/coffea/nanoevents/methods/vector.py
+++ b/coffea/nanoevents/methods/vector.py
@@ -173,7 +173,7 @@ class TwoVector:
 
     @property
     def unit(self):
-        """Unit vector, a vector of length 1 pointing the same direction"""
+        """Unit vector, a vector of length 1 pointing in the same direction"""
         return self / self.r
 
 
@@ -381,7 +381,7 @@ class ThreeVector(TwoVector):
 
     @property
     def unit(self):
-        """Unit vector, a vector of length 1 pointing the same direction"""
+        """Unit vector, a vector of length 1 pointing in the same direction"""
         return self / self.rho
 
 
--- a/coffea/nanoevents/methods/vector.py
+++ b/coffea/nanoevents/methods/vector.py
@@ -154,11 +154,10 @@ class TwoVector:
 
     @awkward.mixin_class_method(numpy.divide, {numbers.Number})
     def divide(self, other):
-        """Divide this vector by a scalar elementwise using `x` and `y` components"""
-        return awkward.zip(
-            {"x": self.x / other, "y": self.y / other},
-            with_name="TwoVector",
-        )
+        """Divide this vector by a scalar elementwise using its cartesian components
+
+        This is realized by using the multiplication functionality"""
+        return self.multiply(1 / other)
 
     def delta_phi(self, other):
         """Compute difference in angle between two vectors
@@ -232,20 +231,6 @@ class PolarTwoVector(TwoVector):
             {
                 "r": self.r * abs(other),
                 "phi": self.phi % (2 * numpy.pi) - (numpy.pi * (other < 0))
-            },
-            with_name="PolarTwoVector",
-        )
-
-    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
-    def divide(self, other):
-        """Divide this vector by a scalar elementwise using using `x` and `y` components
-
-        In reality, this directly adjusts `r` and `phi` for performance
-        """
-        return awkward.zip(
-            {
-                "r": self.r / abs(other),
-                "phi": self.phi % (2 * numpy.pi) + (numpy.pi * (other < 0))
             },
             with_name="PolarTwoVector",
         )
@@ -356,14 +341,6 @@ class ThreeVector(TwoVector):
             with_name="ThreeVector",
         )
 
-    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
-    def divide(self, other):
-        """Divide this vector by a scalar elementwise using `x`, `y`, and `z` components"""
-        return awkward.zip(
-            {"x": self.x / other, "y": self.y / other, "z": self.z / other},
-            with_name="ThreeVector",
-        )
-
     def dot(self, other):
         """Compute the dot product of two vectors"""
         return self.x * other.x + self.y * other.y + self.z * other.z
@@ -444,21 +421,6 @@ class SphericalThreeVector(ThreeVector, PolarTwoVector):
         return awkward.zip(
             {
                 "rho": self.rho * abs(other),
-                "theta": (numpy.sign(other) * self.theta + numpy.pi) % numpy.pi,
-                "phi": self.phi % (2 * numpy.pi) - numpy.pi * (other < 0)
-            },
-            with_name="SphericalThreeVector",
-        )
-
-    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
-    def divide(self, other):
-        """Divide this vector by a scalar elementwise using `x`, `y`, and `z` components
-
-        In reality, this directly adjusts `r`, `theta` `phi` for performance
-        """
-        return awkward.zip(
-            {
-                "rho": self.rho / other,
                 "theta": (numpy.sign(other) * self.theta + numpy.pi) % numpy.pi,
                 "phi": self.phi % (2 * numpy.pi) - numpy.pi * (other < 0)
             },
@@ -574,19 +536,6 @@ class LorentzVector(ThreeVector):
             with_name="LorentzVector",
         )
 
-    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
-    def divide(self, other):
-        """Divide this vector by a scalar elementwise using `x`, `y`, `z`, and `t` components"""
-        return awkward.zip(
-            {
-                "x": self.x / other,
-                "y": self.y / other,
-                "z": self.z / other,
-                "t": self.t / other,
-            },
-            with_name="LorentzVector",
-        )
-
     def delta_r2(self, other):
         """Squared `delta_r`"""
         return (self.eta - other.eta) ** 2 + self.delta_phi(other) ** 2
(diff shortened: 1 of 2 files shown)


coffea/nanoevents/methods/vector.py
```diff
@@ -136,13 +152,30 @@ def multiply(self, other):
             with_name="TwoVector",
         )
 
+    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
+    def divide(self, other):
```
nsmith-, line 156: Could this be implemented by calling multiply on the reciprocal?

coffea/nanoevents/methods/vector.py
```diff
@@ -189,6 +222,42 @@ def r2(self):
...
+        return awkward.zip(
+            {
+                "r": self.r * abs(other),
+                "phi": self.phi % (2 * numpy.pi) - (numpy.pi * (other < 0))
+            },
+            with_name="PolarTwoVector",
+        )
+
+    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
+    def divide(self, other):
```
nsmith-: Same question about multiply with reciprocal

coffea/nanoevents/methods/vector.py
```diff
@@ -271,6 +356,34 @@ def multiply(self, other):
             with_name="ThreeVector",
         )
 
+    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
+    def divide(self, other):
```
nsmith-: Reciprocal

coffea/nanoevents/methods/vector.py
```diff
@@ -322,6 +435,48 @@ def p2(self):
...
+            {
+                "rho": self.rho * abs(other),
+                "theta": (numpy.sign(other) * self.theta + numpy.pi) % numpy.pi,
+                "phi": self.phi % (2 * numpy.pi) - numpy.pi * (other < 0)
+            },
+            with_name="SphericalThreeVector",
+        )
+
+    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
+    def divide(self, other):
```
nsmith-: Reciprocal?

coffea/nanoevents/methods/vector.py
```diff
@@ -406,6 +574,19 @@ def multiply(self, other):
             with_name="LorentzVector",
         )
 
+    @awkward.mixin_class_method(numpy.divide, {numbers.Number})
+    def divide(self, other):
```
nsmith-: Reciprocal?

coffea/nanoevents/methods/vector.py
```diff
@@ -417,6 +598,75 @@ def delta_r(self, other):
...
+            {
+                "x": self.x,
+                "y": self.y,
+                "z": self.z
+            },
+            with_name="ThreeVector"
+        )
+
+    @property
+    def boostvec(self):
```
nsmith-, line 576: If `t <= rho` (or equivalently `mass <= 0`) do we want to make the output NaN since there is no physical way to get to that reference frame?

coffea/nanoevents/methods/vector.py
```diff
@@ -584,14 +834,43 @@ def mass2(self):
...
-        In reality, this multiplies `pt` and `mass` by the scalar quantity for performance
+        In reality, this directly adjusts `pt`, `eta`, `phi` and `mass` for performance
         """
         return awkward.zip(
             {
-                "pt": self.pt * other,
-                "eta": self.eta,
-                "phi": self.phi,
-                "mass": self.mass * other,
+                "pt": self.pt * abs(other),
```
nsmith-: Making `abs(other)` a temporary would be a slight performance boost

nsmith- (changes requested):
Thanks a lot for this! Also for the needed tests. Perhaps @henryiii might want to fold them into the scikit-hep/vector unit tests when the time comes.

coffea/nanoevents/methods/vector.py
jrueb: Would this not be a performance loss? `other` can be an array of numbers, thus it would first allocate an array for the reciprocal and then another one for the multiplication. And I could imagine reciprocal & multiplication also takes more cycles than just a division.

coffea/nanoevents/methods/vector.py
jrueb: I think this would make sense, but would you then also return NaN for zero vectors (there we have `0 == t == rho`)?

coffea/nanoevents/methods/vector.py
nsmith-: In general multiplication is much cheaper than division. So by computing the reciprocal and re-using it for each cartesian component multiplication should actually increase performance. The extra memory cost can be avoided by overwriting, e.g. `other = 1 / other`

nsmith-: I'm not sure how reference lifecycle behaves in a call stack, but python may do "the right thing" if you just implement this as `self.multiply(1 / other)`

jrueb: The only difference between any of the `divide` methods would then by their documentation string. As everything inherits from `TwoVector` it would be enough to implement it only there. Do you want to keep the individual implementations?

nsmith-: Sure, with a generic enough docstring just the TwoVector impl. is fine

coffea/nanoevents/methods/vector.py
nsmith-: As `t` approaches `rho`, `boostvec` approaches the unit vector (i.e. speed of light), so I think it would be consistent if we returned `self.pvec * numpy.minimum(1 / self.rho, 1 / self.t)`.

coffea/nanoevents/methods/vector.py
```diff
@@ -417,6 +598,75 @@ def delta_r(self, other):
...
+        return awkward.Array(out, behavior=self.behavior)
+
+    def boost(self, other):
+        """Apply a Lorentz boost given by the `ThreeVector` `other` and return it
+
+        Note that this follows the convention that, for example in order to boost
+        a vector into its own rest frame, one needs to use the negative of its `boostvec`
+        """
+        b2 = other.rho2
+        gamma = (1 - b2) ** (-0.5)
```
nsmith-, line 599: We should guard somehow against the case `b2 == 1` as well as `b2 == 0`.

coffea/nanoevents/methods/vector.py
nsmith-: I totally ignored your zero vector comment! Let me try again: `self.pvec * awkward.where(self.rho == 0, 0, numpy.minimum(1 / self.rho, 1 / self.t))`. I would make `self.rho` a temporary though as it may be expensive in some cases and is used twice.

lgray: Well - unphysical lorentz transformations still belong to SO(1,3), instead of SO+(1,3). So it's still a lorentz xform but it cannot be described by a boost and a rotation. 

It may be better, and more correct, to simply warn that a boost was calculated outside of the restricted lorentz group. Then also to protect against zeros in some reasonable way.

This may also suggest returning the the boost tensor rather than just the vec... Or extending what we do to include that notion. It then becomes much easier to impose notions of physicality since it's pretty easy to check the lorentz xform tensor for what type it is.

coffea/nanoevents/methods/vector.py
jrueb: Right now for `b2==1` there will be a warning and `inf` as vector components. Is there anything better result in this situation?

coffea/nanoevents/methods/vector.py
jrueb: I made it depend on `|t| <= rho`. It not returning a normal boost vector even if `|t| >> rho` sounds like a possible pitfall to me and one can easily get into this situation by `(-lorentzvec).boostvec`, which works fine in uproot_methods and ROOT. 
If the user tries to boost with a vector coming from `|t| <= rho` there will be a warning in the `boost` function.

coffea/nanoevents/methods/vector.py
nsmith-: That seems reasonable, I wasn't sure what the behavior was currently.

coffea/nanoevents/methods/vector.py
nsmith-: Not to wax too long on this but based on https://en.wikipedia.org/wiki/Lorentz_group#Connected_components I think actually improper transforms are outside SO(1, 3). But also |boostvec| > 1 would give complex-valued output in the current `boost` implementation so I'm not sure it makes any sense even as a parity-reversing transform.

nsmith- (approved):
Looks good

PR #449, opened 2021-02-10
Add support for reading NanoEvents from Parquet files stored in Skyhook through the Arrow Datasets API

JayjeetAtGithub:
Please refer to [this](https://github.com/uccross/coffea/blob/use-dataset-api/docker/coffea_rados_parquet/README.md) document for more details on this PR.


lgray:
@JayjeetAtGithub Please fix the linting errors! Thanks!

JayjeetAtGithub:
Hi @lgray, I  fixed the linting errors and ensured the tests run.  I see some installing dependencies on macOS are failing, and that is leading to other tests getting aborted. 

I want to add integration tests for the `run_rados_parquet_job` API introduced by this PR by booting up a single node ceph cluster in a docker container and then run the tests, so testing the end to end functionality. Do you have any suggestions on how to go ahead with this?  Is there something similar already in place for the `run_spark_job` API ? Thanks a lot.

lgray:
Can you please rebase this PR to the present master branch?

JayjeetAtGithub:
@lgray Done.

lgray:
Branch still shows as out of date. (and github won't present the button to merge in master myself???)

Please make sure you've gotten the latest master from CoffeaTeam/coffea.

JayjeetAtGithub:
Hi @lgray , sorry, It went out of date again somehow. Fixed it. Thanks.

lgray:
@JayjeetAtGithub going through this it looks like your ci definition didn't run?

Can you add it to the main ci definition so it's run regularly, similar to workqueue?

JayjeetAtGithub:
Hi @lgray, I have added the tests to the Github workflow along with some more improvements. Please have a look. Thanks.

coffea/processor/executor.py
```diff
@@ -96,15 +96,24 @@ def chunks(self, target_chunksize, align_clusters):
...
+        'dataset',
+        'filename',
+        'treename',
+        'entrystart',
+        'entrystop',
+        'fileuuid',
+        'extras'
+    ]
+
+    def __init__(self, dataset, filename, treename, entrystart, entrystop, fileuuid, extras={}):
```
nsmith-: Avoid mutable default arguments because they are essentially global variables and it is easy to take a reference to them (as you did just below!) Rather set it to `None` and set `self.extras = extras or {}`

coffea/processor/executor.py
```diff
@@ -96,15 +96,24 @@ def chunks(self, target_chunksize, align_clusters):
 
 
 class WorkItem(object):
-    __slots__ = ['dataset', 'filename', 'treename', 'entrystart', 'entrystop', 'fileuuid']
-
-    def __init__(self, dataset, filename, treename, entrystart, entrystop, fileuuid):
+    __slots__ = [
```
nsmith-, line 99: I'm a bit hesitant to allow the work item to grow in size, this is exactly why I used slots. These items are numerous and have to be shipped to the scheduler. It takes a non-trivial amount of time to serialize them.

coffea/processor/executor.py
```diff
@@ -1337,3 +1372,98 @@ def run_spark_job(fileset, processor_instance, executor, executor_args={},
         spark = None
 
     return output
+
+
+def run_parquet_job(fileset, treename, processor_instance, executor, executor_args={}):
```
nsmith-, line 1380: There's a lot of code duplication here. This is fine for now but we should try to find a way to refactor run_uproot_job to be able to accept parquet.

coffea/processor/executor.py
```diff
@@ -1337,3 +1372,98 @@ def run_spark_job(fileset, processor_instance, executor, executor_args={},
...
+    xrootdtimeout = executor_args.pop('xrootdtimeout', None)
+    ceph_config_path = executor_args.pop('ceph_config_path', None)
+    is_rados_parquet = False
+    if ceph_config_path:
+        is_rados_parquet = True
+
+    chunks = []
+    for dataset, filelist in dataset_filelist_map.items():
+        for filename in filelist:
+            chunks.append(WorkItem(dataset, filename, treename, 0, 0, '', {'ceph_config_path': ceph_config_path, 'is_rados_parquet': is_rados_parquet}))
```
nsmith-: Do I understand correctly that the current implementation cannot process chunks of data smaller than an entire parquet file? I guess the size can be controlled during the uproot root->parquet conversion step.

wf.yml
```diff
@@ -0,0 +1,8 @@
+steps:
```
nsmith-: Can this file be renamed something more specific?

.github/workflows/ci.yml
```diff
@@ -124,10 +124,31 @@ jobs:
...
+      shell: bash -l {0}
+      run: |
+        docker build -t coffea-rados-parquet-test \
+          --file docker/coffea_rados_parquet/Dockerfile \
+          .
+        docker run \
+        -v $(pwd):/w \
+        -w /w \
+        -e IS_CI=true \
+        --privileged \
```
nsmith-, line 144: This may mean trouble for certain environments with more restrictive permissions. What privileges does the container need? Can they be specified more exactly?

nsmith- (changes requested):
Thanks for the PR, clearly this is a big new feature, so we may need to iterate to fold it in properly (and payoff the increasingly large technical debt in `executor.py`)

coffea/processor/executor.py
lgray: Given that this is not reading an uproot file... `run_uproot_job` is a misnomer.
We should refactor our naming (and the code too, anyway) or go with the `from_root` / `from_parquet` setup we have in nanoevents.

coffea/processor/executor.py
```diff
@@ -843,9 +852,20 @@ def _get_cache(strategy):
     return cache
 
 
+class ParquetFileContext():
+    def __init__(self, filename):
+        self.filename = filename
+
+    def __enter__(self):
```
nsmith-, line 857: Here the file should be opened. The purpose of the context manager is to ensure we close the file and hopefully deallocate the arrays that were read from it regardless of how the processing went or if a reference to the NanoEvents objects persists.

coffea/nanoevents/mapping/parquet.py
```diff
@@ -6,18 +6,29 @@
...
             self.file = file
+            self.shim_options = shim_options
+            self.schema = shim_options['schema_arrow']
+            self.format = shim_options['format']
 
         def read(self, column_name):
             # make sure uproot is single-core since our calling context might not be
-            return self.file.read([column_name], use_threads=False)
+            if self.format == 'rados-parquet':
+                return ds.dataset(
```
nsmith-: If we're re-constructing the dataset each column access, doesn't this mean a whole connection setup and teardown? We should construct the parquet dataset object once in the constructor, or even back in the NanoEventsFactory

wf.yml
JayjeetAtGithub: Fixed

coffea/nanoevents/mapping/parquet.py
JayjeetAtGithub: Yeah, very sorry for that. I realize I was kind of ignorant of the performance and code quality and focussed more on getting the stuff to work. Will look for more such issues.

coffea/processor/executor.py
JayjeetAtGithub: I see. I think we can encode the extra information in the filename parameter then. For example, 
we can have the filename as `/path/to/conf:file1.parquet`. What do you think ?

JayjeetAtGithub: I can then parse it in the work_function.

coffea/processor/executor.py
JayjeetAtGithub: Removed it.

coffea/processor/executor.py
JayjeetAtGithub: The dataset API requires file paths as input. It does not accept open file handles. So, that's why I created a dummy ContextManager.

.github/workflows/ci.yml
JayjeetAtGithub: The container needs to have the `SYS_ADMIN` capability along with mounting the `/dev/fuse` device to be able to mount the Ceph file system using FUSE. Although it is user-space, but still, Docker needs high priviliges to be able to mount using FUSE according to [this](https://github.com/docker/for-linux/issues/321) open issue. Can we give `SYS_ADMIN` cap to the container for now?

coffea/processor/executor.py
JayjeetAtGithub: I am trying to understand the chunking case. Will let you know what I understand.

coffea/nanoevents/factory.py
```diff
@@ -205,7 +207,14 @@ def from_parquet(
         mapping = ParquetSourceMapping(
             TrivialParquetOpener(uuidpfn, parquet_options), access_log=access_log
         )
-        shim = TrivialParquetOpener.UprootLikeShim(table_file)
+
+        format_ = 'parquet'
+        if 'ceph_config_path' in rados_parquet_options:
+            format_ = ds.RadosParquetFileFormat(rados_parquet_options['ceph_config_path'].encode())
```
nsmith-, line 213: Might this be more flexible as `ds.RadosParquetFileFormat(**rados_parquet_options)`? Though maybe on your side the signature would need changing

coffea/processor/executor.py
nsmith-: I think this is a good approach for now. Eventually we need to abstract `filename` as any of several kinds of `source`s e.g. rados-parquet, uproot-root, plain parquet, etc. and the work function can dispatch the appropriate `NanoEventsFactor.from_X` according to the source type.

coffea/processor/executor.py
nsmith-: Ok. Do you think it is possible to create the Dataset object at this level? This would mean changing `NanoEventsFactory.from_parquet` to `NanoEventsFactory.from_arrowdataset` or something of a similar name. I think this is a better abstraction because then NanoEventsFactory can take any arrow dataset type in the future. @lgray

coffea/processor/executor.py
```diff
@@ -843,9 +850,20 @@ def _get_cache(strategy):
...
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        pass
+
+
 def _work_function(item, processor_instance, savemetrics=False,
                    mmap=False, schema=schemas.BaseSchema, cachestrategy=None, skipbadfiles=False,
-                   retries=0, xrootdtimeout=None, use_dataframes=False):
+                   retries=0, xrootdtimeout=None, use_dataframes=False, format="root"):
```
nsmith-, line 866: I mentioned earlier that eventually we should rework `filename` to be `source` and key on that to decide how to open the source. That way even a single job can process several different sources (e.g. half root files, half skyhook) This can wait for a different PR though.

coffea/processor/executor.py
nsmith-: Ok lets defer this to later

tests/test_rados_parquet_job.py
```diff
@@ -0,0 +1,78 @@
+import os
```
nsmith-, line 1: We should work to prepare a container to run these tests in github CI eventually. I noticed that the current container is extremely large (7.5GB) any chance to slim it down a bit?

nsmith- (approved):
I'm mostly happy with the implementation. I think there a few refactors that would be good to do now still but won't force them. There are a few more that definitely can wait for another PR though.

nsmith- (approved):
Looks like a force-push reset the approval, but the branch is the same as before.

nsmith-:
Can you update the branch from master (or allow edits from maintainers?)

JayjeetAtGithub:
Hi @nsmith-, Yes, there was a force push but that did not change the branch. I have updated the branch from master. Please take a look. Thanks.

PR #427, opened 2021-01-21
PFNano additions for NanoEvents

nsmith-:
Allows things like:
Screen Shot 2021-01-20 at 10 55 13 PM

Some cross-references aren't yet available in the PFNano file I was using. A complete set would need at least the following: https://gist.github.com/nsmith-/b29c4f4a13214ed65b93bf2d5a65f955

PR #450, opened 2021-02-11
fix: add matched_gen for FatJet

andrzejnovak:


nsmith-:
I had added this in #427 but its still waiting on some other items so we can merge this.

nsmith- (commented):
Note this does not appear to be present in stock NanoAOD for some reason.

PR #456, opened 2021-02-18
Fix Breaking WQ Test

dthain:
- Match slight change to Factory API.
- wq.py -> test_work_queue.py for consistency


lgray:
@dthain now pytest is picking up test_work_queue.py as a test to run and failing, because of the `test_` prefix.

Maybe just name it "work_queue.py" ?

dthain:
Trying too hard...

PR #460, opened 2021-02-20
Add checks to verify the validity of the JEC/JER correction tools

### LINKED ISSUE #455: Bug found with JEC/JER lookup tools (coffea.lookup_tools.jme_standard_function.py)
**Describe the bug**
There is a specific corner case which will cause a problem for the JEC/JER lookup tools. All of the jet tools rely on lookup tables, where there are binned and un-binned variables. This behavior is specified in `coffea.lookup_tools.jme_standard_function.py::_evaluate`.

When asked about a jet with properties outside of the range of validity for the un-binned variables, the correct behavior is to return the value of the corrections/SF/resolution at the boundary of the lowest/highest value in the range of validity. This behavior is correctly reproduced in Coffea (here using the CMSSW/C++ version of the lookup tools as golden references). However, when querying a lookup table for a jet with properties outside the specified bin ranges, the correct behavior is to return an identity (i.e. 1.0 for multiplicative corrections). This behavior is not being correctly reproduced. Instead, some non-identity value is being returned. As you can be, there is an intrinsic difference in the behavior of binned and un-binned variables.

First let me demonstrate an example of everything working. Assume we had a jet with `pt=36.9` and `eta=2.82`. We also have the correction lookup table (abbreviated):
```
{2 JetEta JetPt 1 JetPt max(0.0001,[0]+((x-[1])*([2]+((x-[1])*([3]+((x-[1])*[4])))))) Correction L2Relative}
-5.191         -4.889        5.65225        5.67149              7        5.65225        5.67149        1.68328        5.65225        -14.045        120.643        4896.33
-5.191         -4.889        5.67149        5.68974              7        5.67149        5.68974        1.49256        5.67149       -3.95898        113.642        745.105
...
 2.322            2.5        549.097           6500              7        549.097        728.873        1.08981        549.097   -2.88411e-05    2.88431e-07    9.03247e-11
   2.5           2.65        4.02981         4.2226              7        4.02981         4.2226       0.999657        4.02981        1.06597      -0.619679        -0.0494
...
  2.65          2.853        28.1263        35.8534              7        28.1263        35.8534        1.47843        28.1263     -0.0159995     0.00022154    2.39137e-05
  2.65          2.853        35.8534        48.3907              7        35.8534        48.3907        1.37906        35.8534    -0.00829227    7.96644e-05    5.18988e-06
  2.65          2.853        48.3907        63.8369              7        48.3907        63.8369        1.29785        48.3907    -0.00384741   -1.55835e-06    2.33578e-06
...
```
We would first find that the eta variable is in the range [2.65,2.853) and then we'd find that the pt variable is in the range [35.8534,48.3907). We further see that the jet pt is within the range of validity of the un-binned JetPt variable. Therefore, the correction should be `1.37038741364`, and it is.

Now let's look an example with an incorrect behavior. Assume you have a jet with `pt=3.54` and `eta=2.53`. While we can find the correct bin for eta, [2.5,2.65), the pt is below the minimum allowed binned value. There is no set of parameters what will return a valid correction for this jet. Although the lookup tools should return a correction of `1.0`, they instead return a correction of `0.9996569752693176`, which is outside of the `1e-6` tolerance we have been accepting.

**To Reproduce**
To produce the correct behavior using pyroot/CMSSW, you can run:
```python
import os
import ROOT as r
era = "Summer16_23Sep2016V3_MC"
vPar = r.vector(r.JetCorrectorParameters)()
vPar.push_back(r.JetCorrectorParameters(os.path.expandvars('$CMSSW_BASE/src/{}_L2Relative_AK4PFPuppi.jec.txt'.format(era))))
jec = r.FactorizedJetCorrector( vPar )
jec.setJetEta(2.82)
jec.setJetPt(36.9)
print("{0:.9f}".format(jec.getCorrection()))
jec.setJetEta(2.53)
jec.setJetPt(3.54)
print("{0:.9f}".format(jec.getCorrection()))
```
Here I'm assuming you have the `Summer16_23Sep2016V3_MC_L2Relative_AK4PFPuppi.jec.txt` JEC text file in the `src` directory of your CMSSW release. This file can be found at `coffea/test/samples/`.

The result will be:
```python
1.370474577
1.000000000
```

To reproduce what happens when using Coffea do:
```python
from coffea import lookup_tools
import uproot
import awkward as ak
import numpy as np
from coffea.jetmet_tools import FactorizedJetCorrector

extract = extractor()
extract.add_weight_sets(['* * tests/samples/Summer16_23Sep2016V3_MC_L2Relative_AK4PFPuppi.jec.txt.gz'])
extract.finalize()

jec_names = ['Summer16_23Sep2016V3_MC_L2Relative_AK4PFPuppi']
corrector = FactorizedJetCorrector(**{name: evaluator[name] for name in jec_names})

test_pt = np.array([36.9,3.54])
test_eta = np.array([2.82,2.53])
test_corrs = corrector.getCorrection(JetEta=test_eta, JetPt=test_pt)
print(test_corrs)
```

The result will be:
```python
[1.37047456 0.99965698]
```

 - OS: Scientific Linux
 - Version: 7.8 (Nitrogen)
 - Coffea version: `master` branch at commit `aa0b2fb`

aperloff:
This PR seeks to verify that the JEC and JER tools are working properly. For each test, the correction or scale factors being returned are also calculated by the CMSSW tools and/or by hand. The checks verify that the Coffea tools return the same values as the by-hand calculations, to within a precision of `1e-6`. `test_corrected_jets_factory` now also checks that the JECs and JERs are chained together correctly. While this is fairly robust confirmation, there isn't a by-hand check that the uncertainties (for JEC or JER) are also chained appropriately.

While many of the tests do pass, some of them do not as this PR relies upon #455 being fixed. The failing test should work once the aforementioned issue is fixed.

### DIFF
--- a/tests/test_jetmet_tools.py
+++ b/tests/test_jetmet_tools.py
@@ -6,7 +6,9 @@ import uproot
 import awkward as ak
 from coffea.util import numpy as np
 
-import pytest, time # , pyinstrument
+import pytest
+import time
+# import pyinstrument
 
 from dummy_distributions import dummy_jagged_eta_pt, dummy_four_momenta
 
@@ -32,6 +34,7 @@ def jetmet_evaluator():
 
     return extract.make_evaluator()
 
+
 evaluator = jetmet_evaluator()
 
 
@@ -42,6 +45,7 @@ def test_factorized_jet_corrector():
     test_Rho = np.full_like(test_eta, 100.)
     test_A = np.full_like(test_eta, 5.)
 
+    # Check that the FactorizedJetCorrector is functional
     jec_names = ['Summer16_23Sep2016V3_MC_L1FastJet_AK4PFPuppi',
                  'Summer16_23Sep2016V3_MC_L2Relative_AK4PFPuppi',
                  'Summer16_23Sep2016V3_MC_L2L3Residual_AK4PFPuppi',
@@ -52,6 +56,7 @@ def test_factorized_jet_corrector():
 
     pt_copy = np.copy(test_pt)
 
+    # Check that the corrector can be evaluated for flattened arrays
     corrs = corrector.getCorrection(JetEta=test_eta, Rho=test_Rho, JetPt=test_pt, JetA=test_A)
 
     assert((np.abs(pt_copy - test_pt) < 1e-6).all())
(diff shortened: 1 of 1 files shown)


lgray:
Hey @aperloff sorry for the delay here. I wasn't very happy with possible solutions here so I think we're going to first fix this in a very dumb way and then put the next iteration of fixes into correctionlib.

I'll try to get this done within the week.

PR #462, opened 2021-02-25
Add conversion routines for boost_histogram and hist

nsmith-:
Still needs some tests

### DIFF
--- a/coffea/hist/hist_tools.py
+++ b/coffea/hist/hist_tools.py
@@ -1327,45 +1327,43 @@ class Hist(AccumulatorABC):
 
         newaxes = []
         for axis in self.axes():
-            metadata = {"name": axis.name, "label": axis.label}
             if isinstance(axis, Bin) and axis._uniform:
-                newaxes.append(
-                    boost_histogram.axis.Regular(
-                        axis._bins,
-                        axis._lo,
-                        axis._hi,
-                        underflow=True,
-                        overflow=True,
-                        metadata=metadata,
-                    )
+                newaxis = boost_histogram.axis.Regular(
+                    axis._bins,
+                    axis._lo,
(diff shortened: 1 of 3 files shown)


coffea/hist/hist_tools.py
```diff
@@ -1318,3 +1318,81 @@ def identifiers(self, axis, overflow='none'):
...
+                newaxis.label = axis.label
+                newaxes.append(newaxis)
+            elif isinstance(axis, Cat):
+                identifiers = self.identifiers(axis)
+                newaxis = boost_histogram.axis.StrCategory(
+                    [x.name for x in identifiers],
+                )
+                newaxis.name = axis.name
+                newaxis.label = axis.label
+                newaxis.bin_labels = [x.label for x in identifiers]
```
nsmith-, line 1357: @andrzejnovak do you know if category custom labels (e.g. latex) are supported in hist/mplhep?

andrzejnovak: mpl will render that fine, but I don't think hist HTML repr can interpret latex (@henrii)

andrzejnovak: image

nsmith-: I noticed that issue as well, but here I'm asking about `StrCategory` axes. In coffea hists, when displaying a histogram with a `StrCategory` axis, by default it would use the category names as bin labels, but one could optionally override the category names with custom labels (so that the filling code could still use a simple ascii shorthand like `jetptcut` while the plot would show a more expressive label like `$p_{T,j} \geq 100 GeV$`)

nsmith-:
I'm going to merge as-is, perhaps later we'll find a convention for overriding bin labels.

PR #464, opened 2021-03-03
Take first branch if duplicate branch names exist in tree 

nsmith-:
To be consistent with uproot getitem behavior


coffea/nanoevents/mapping/uproot.py
```diff
@@ -34,6 +34,11 @@ def __init__(self, fileopener, cache=None, access_log=None):
     def _extract_base_form(cls, tree):
         branch_forms = {}
         for key, branch in tree.iteritems():
+            if key in branch_forms:
+                warnings.warn(
```
nsmith-, line 38: @andrzejnovak too onerous?

andrzejnovak: Nah, good to have that in. User can always filter

PR #470, opened 2021-03-19
Fixing attribute error when loading root file with TGraphErrors in add_weight_sets

### LINKED ISSUE #440: Attribute error when loading root file with TGraphErrors in add_weight_sets
**Describe the bug**
If the root file with scale factors contains TGraphErrors objects an Attribute error will be thrown because the object has no axes:

```
---> 22         edges = tuple(ax.edges() for ax in item.axes)
     23         if rootclass in histTypes:
     24             converted_file[(nicepath, "dense_lookup")] = item.values(), edges

AttributeError: 'Model_TGraphErrors_v3' object has no attribute 'axes'
```

This happens even if the TGraphErrors object is not used.

**To Reproduce**
```

ext = extractor()
ext.add_weight_sets(["ele_2016_loose EGamma_SF2D TnP_loose_ele_2016.root "])
```
TnP_loose_ele_2016.root  has to contain a TGraphErrors object.
EGamma_SF2D is just a usual TH2D.

**Expected behavior**
Just ignore the TGraphErrors objects.

The problem is this line:
https://github.com/CoffeaTeam/coffea/blob/master/coffea/lookup_tools/root_converters.py#L22

Can this line just be moved after the if statement? I tested this locally, and it works fine for my use cases so far. Let me know if I should just open a PR with the fix.

danbarto:
Just moving a line in coffea/lookup_tools/root_converters.py as proposed in #440:

### DIFF
--- a/coffea/lookup_tools/root_converters.py
+++ b/coffea/lookup_tools/root_converters.py
@@ -19,8 +19,8 @@ def convert_histo_root_file(file):
     for path, item in fin.iteritems(recursive=True):
         nicepath = killcycle(path, cycle)
         rootclass = item.classname
-        edges = tuple(ax.edges() for ax in item.axes)
         if rootclass in histTypes:
+            edges = tuple(ax.edges() for ax in item.axes)
             converted_file[(nicepath, "dense_lookup")] = item.values(), edges
             if hasattr(item, "variances"):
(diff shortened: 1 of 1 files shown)


nsmith- (approved):
Thanks

PR #472, opened 2021-03-25
New accumulator semantics

### LINKED ISSUE #180: Accumulator refactor
This is a project-tracking issue to refactor the coffea accumulators.  Goals:

- Migrate all accumulator types out of the `processor` submodule
- Remove requirement for an identity
- Enforce minimal copy by restricting add behavior to only in-place `__iadd__`

The second point has been brought up quite a while ago by @jpivarski but it seems much more straightforward now that I'm aware of this nice python behavior: for a class with the following arithmetic overrides,
```python
class accumulator:
    def __init__(self, value):
        self._value = value

    def __repr__(self):
        return "accumulator(%r)" % (self._value, )

    def __radd__(self, other):
        if other is None:
            return self
        return NotImplemented

    def __iadd__(self, other):
        if isinstance(other, accumulator):
            self._value += other._value
            return self
        elif isinstance(other, type(self._value)):
            self._value += other
            return self
        return NotImplemented

    @property
    def value(self):
        return self._value
```
this code will return `True`
```python
a = accumulator(0)
b = None
b += a
b is a
```
Unfortunately this means that the `ProcessorABC` idiom may need a rework: the user would need to provide an accumulator factory function rather than an accumulator object.

nsmith-:
Keeping the old accumulators intact but allowing new ones to be used in processor output

### DIFF
--- a/coffea/processor/accumulator.py
+++ b/coffea/processor/accumulator.py
@@ -1,11 +1,16 @@
-from abc import ABCMeta, abstractmethod
-from collections.abc import MutableSet, MutableMapping
-from collections import defaultdict
 import copy
 import operator
-from typing import Union, Optional, Protocol, TypeVar, Iterable
-import numpy
+from abc import ABCMeta, abstractmethod
+from collections import defaultdict
+from collections.abc import MutableMapping, MutableSet
+from typing import Iterable, Optional, TypeVar, Union
 
+try:
+    from typing import Protocol  # type: ignore
+except ImportError:
+    from typing_extensions import Protocol
+
+import numpy
 
 T = TypeVar("T")
 
--- a/coffea/processor/accumulator.py
+++ b/coffea/processor/accumulator.py
@@ -3,7 +3,7 @@ import operator
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 from collections.abc import MutableMapping, MutableSet
-from typing import Iterable, Optional, TypeVar, Union
+from typing import Dict, Iterable, Optional, Set, TypeVar, Union
 
 try:
     from typing import Protocol  # type: ignore
@@ -20,7 +20,7 @@ class Addable(Protocol):
         ...
 
 
-Accumulatable = Union[MutableSet, MutableMapping, Addable]
+Accumulatable = Union[Set, Dict, Addable]
 
 
 def add(a: Accumulatable, b: Accumulatable) -> Accumulatable:
--- a/coffea/processor/accumulator.py
+++ b/coffea/processor/accumulator.py
@@ -1,11 +1,77 @@
 from abc import ABCMeta, abstractmethod
-from collections.abc import Set, Mapping
+from collections.abc import MutableSet, MutableMapping
 from collections import defaultdict
+import copy
+import operator
+from typing import Union, Optional, Protocol, TypeVar, Iterable
 import numpy
 
 
+T = TypeVar("T")
+
+
+class Addable(Protocol):
+    def __add__(self: T, other: T) -> T:
+        ...
+
+
+Accumulatable = Union[MutableSet, MutableMapping, Addable]
+
+
+def add(a: Accumulatable, b: Accumulatable) -> Accumulatable:
+    """Add two accumulatables together, without altering inputs
+
+    This may make copies in certain situations
+    """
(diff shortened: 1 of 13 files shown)


coffea/processor/accumulator.py
```diff
@@ -1,11 +1,82 @@
...
+    """Add two accumulatables together, without altering inputs
+
+    This may make copies in certain situations
+    """
+    if isinstance(a, MutableSet) and isinstance(b, MutableSet):
+        return a | b
+    elif isinstance(a, MutableMapping) and isinstance(b, MutableMapping):
+        out = copy.copy(a)
+        out.clear()
+        for key in set(a) & set(b):
```
lgray: Not sure if it matters *that* much but minimizing the number of times we iterate over keys of a and b could be important for large mappings. The set operations are all very fast but the creation of `set(a)` and `set(b)` may not be.
```python3
a_keyset = set(a)
b_keyset = set(b)
for key in a_keyset & b_keyset:
    ...
```

If it's possible to non-hideously get rid of the walk over `a` from the instantiation of `out`, that would be good too.

coffea/processor/accumulator.py
```diff
@@ -1,11 +1,82 @@
...
+    else:
+        return operator.add(a, b)
+
+
+def iadd(a: Accumulatable, b: Accumulatable) -> Accumulatable:
+    """Add two accumulatables together, assuming the first is mutable"""
+    if isinstance(a, MutableSet) and isinstance(b, MutableSet):
+        return operator.ior(a, b)
+    elif isinstance(a, MutableMapping) and isinstance(b, MutableMapping):
+        for key in set(a) & set(b):
```
lgray: same comment as for `add` about re-instantiating sets that you use over and over.

coffea/processor/accumulator.py
```diff
@@ -1,11 +1,82 @@
...
+    elif isinstance(a, MutableMapping) and isinstance(b, MutableMapping):
+        out = copy.copy(a)
+        out.clear()
+        for key in set(a) & set(b):
+            out[key] = add(a[key], b[key])
+        for key in set(a) - set(b):
+            out[key] = copy.deepcopy(a[key])
+        for key in set(b) - set(a):
+            out[key] = copy.deepcopy(b[key])
+        return out
```
lgray, line 55: It makes more sense in iadd, but do we want add to always take the type of the first operand? I think the usual behavior would be to take the first common non-abstract type shared by a and b (counting back from the passed types)? 

I'm fine with this being the rule so long as we are clear about it. 

What about user defined accumulables which are somehow not compatible with each other? We should detect that and throw an error, to keep people from getting confusing answers.

lgray (changes requested):
One minor performance gripe and then one more important point on behavior. Otherwise the PR looks very good.

coffea/processor/accumulator.py
nsmith-: The eventual call to `a + b` would throw if they are not compatible

nsmith-: There is a loss of the type of the mapping `b` here but since I figure these are almost always just dictionaries no need to work too hard to resolve the common type of `a` and `b`. It would be a challenging problem. I had originally thought I would convert everything to plain dict in the output, but it might be nice for user to have some chance of receiving what they put in (e.g. defaultdict)

lgray: Yeah for sure the common case is covered well by this. I'm only concerned with not-to-farfetched uncommon cases from people who like tinkering.

It's also fine to simply say as a rule that for different input types the lhs operand's type is taken. That gives enough control to the user and just requires docs/comments.

nsmith-: Another option is to require `type(a) == type(b)` which in this context is how we would expect things to be used regardless. For some of our executors you can't really guarantee what ends up on lhs in a reproducible way since it depends on the order the jobs finish

nsmith-: I went for a bit of a softer type check, take a look at the tests to see how it plays out

lgray: That's also a fine solution. I think what we have right now is not confusing.

lgray (approved):

PR #482, opened 2021-04-02
Retry the coffea_rados_parquet job 3 times to overcome the rare OSD segfault issue

JayjeetAtGithub:


JayjeetAtGithub:
@nsmith- Looks like the error while bringing a Ceph cluster up is very transient and it's a little difficult to predict the actual cause with almost no good answers to that on the internet. 

So, I added a retry script to try the docker run 3 times or fail silently if the error happens all three times in a row, which I think will not be the case, the error being very rare.  But this patch will do the job for now. Thanks .

lgray:
I'd say if it fails three times in a row let it throw and error so we at least look at it. If there's a real problem and we can't see it, then why test?

JayjeetAtGithub:
Yeah, you are right. Let me fix

JayjeetAtGithub:
@nsmith- @lgray   fixed

PR #484, opened 2021-04-03
Setup speed and more deterministic behavior for jetmet tools

### LINKED ISSUE #454: Deterministic jet smearing
We'll need to find an approach to ensure that CorrectedJetsFactory builds JER-smeared jets in a repeatable way.
Options include:

- Seed once per chunk (will only be deterministic for the same chunk splitting)
- Seed once per event (too expensive in numpy, maybe a numba method could produce the RNG value quickly)
- Directly construct a random number by hashing some set of event and jet attributes

CMSSW and [NanoAODTools](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/25a793ec55b30fe7107af263c4523f20ff1a5fbd/python/postprocessing/modules/jme/jetSmearer.py#L85) uses once-per-event seeding.
Direct construction could be done by using an [integer hashing function](https://stackoverflow.com/questions/664014/what-integer-hash-function-are-good-that-accepts-an-integer-hash-key) and then apply [Box-Muller](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform)

### LINKED ISSUE #461: Output of FactorizedJetCorrector.getCorrection has too many dimensions
**Describe the bug**
When only events with 0 jets are selected and one tries to use `FactorizedJetCorrector` to get jet corrections for these events, there will be a dimension too much in the result of `getCorrection`. The corrections need to have `parms`.

**To Reproduce**
```python
import awkward as ak
import coffea.lookup_tools
import coffea.jetmet_tools
extract = coffea.lookup_tools.extractor()
extract.add_weight_sets(['* * Autumn18_V19_MC_L1FastJet_AK4PFchs.txt',])
extract.finalize()
corrector = coffea.jetmet_tools.FactorizedJetCorrector(**evaluator)
eta = ak.Array([[]])
rho = ak.Array([5])
pt = ak.Array([[]])
area = ak.Array([[]])
corr = corrector.getCorrection(JetPt=pt, JetEta=eta, JetA=area, Rho=rho)
print(repr(corr))
```
Using the correction file from here https://github.com/cms-jet/JECDatabase/blob/master/textFiles/Autumn18_V19_MC/Autumn18_V19_MC_L1FastJet_AK4PFchs.txt
Will print `<Array [[]] type='1 * var * 0 * float64'>`

**Expected behavior**
The number of dimensions should be the same as when there are jets present, meaning the result should be `<Array [[]] type='1 * var * float64'>`.

Using coffea version 0.7.1

**Additional context**
The additional dimension easily leads to broadcasting errors.
My guess it that the use of `squeeze` in `jme_standard_function._evaluate` causes this, for example here https://github.com/CoffeaTeam/coffea/blob/05af78bed2d890169842117d7ad647259dc40fa9/coffea/lookup_tools/jme_standard_function.py#L159
`squeeze` only acts on axes with length 1, but here we have length 0. Maybe using `flatten` instead solves this.

lgray:
- [x] make instantiating corrected jets faster at the cost of a longer evaluation time
- [x] make instantiating corrected met faster at the cost of a longer evaluation time
- [x] use per-batch deterministic random seed for jet smearing ( fix #454 )
- [x] investigate and fix reported memory leak in met corrector ( fixed in #487 )
- [x] investigate and fix #461


lgray:
@JayjeetAtGithub can you see what's going on with this failure of the rados tests? It looks like there's something screwey on the docker side of things. It's not failing in anything related to coffea.

lgray:
@kondratyevd if you have time - give this a quick look. The instantiation speed is an order of magnitude faster.

JayjeetAtGithub:
@lgray Yeah, I am trying to find the problem

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -240,9 +240,14 @@ def build(self, jets, lazy_cache):
                 **jersfargs, form=_JERSF_FORM, lazy_cache=lazy_cache
             )
 
+            seeds = numpy.array(out_dict[self.name_map["JetPt"] + "_orig"])[[0, -1]]
```
nsmith-: Can just do `seeds = numpy.array(out_dict[self.name_map["JetPt"] + "_orig"])[[0, -1]].view("i4")`

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -240,9 +240,14 @@ def build(self, jets, lazy_cache):
...
+            seeds = numpy.frombuffer(seeds.tostring(), dtype=numpy.int32)
             out_dict["jet_resolution_rand_gauss"] = awkward.virtual(
                 rand_gauss,
-                args=(out_dict[self.name_map["JetPt"] + "_orig"],),
+                args=(
+                    out_dict[self.name_map["JetPt"] + "_orig"],
+                    numpy.random.RandomState(seeds),
```
nsmith-: RandomState is listed as "legacy" in the numpy docs. How about:
```python
numpy.random.Generator(numpy.random.PCG64(seeds))
```

nsmith- (changes requested):

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -394,63 +404,82 @@ def build(self, jets, lazy_cache):
...
+                    args=(
+                        jets,
+                        out_dict[f"jet_energy_uncertainty_{name}"],
+                        self.name_map["JetPt"],
+                        out_dict[juncnames["JetPt"]],
+                        self.name_map["JetMass"],
+                        out_dict[juncnames["JetMass"]],
+                    ),
+                    length=len(out),
+                    cache=lazy_cache,
```
nsmith-: Putting a cache here makes me nervous, can you maybe add a check in the tests for proper cleanup if jets are made and then maybe just one thing is accessed?

coffea/jetmet_tools/CorrectedMETFactory.py
```diff
@@ -125,27 +149,26 @@ def make_variant(*args):
...
+                    unc,
+                    self.name_map["METpt"],
+                    self.name_map["METphi"],
+                    self.name_map["JetPt"],
+                    self.name_map["JetPhi"],
+                    self.name_map["ptRaw"],
+                    MET,
+                    corrected_jets,
+                ),
+                length=length,
```
nsmith-, line 162: Just a note, since no `cache` is supplied here, we should avoid cache reference cycles.

coffea/jetmet_tools/JetTransformer.py
```diff
@@ -29,11 +25,11 @@ def _update_jet_ptm(corr, jet, fromRaw=False):
...
+        jet._content._contents["__fast_pt"] = corr * jet.ptRaw.content
+        jet._content._contents["__fast_mass"] = corr * jet.massRaw.content
     else:
-        jet._content._contents['__fast_pt'] = corr * jet.pt.content
-        jet._content._contents['__fast_mass'] = corr * jet.mass.content
+        jet._content._contents["__fast_pt"] = corr * jet.pt.content
+        jet._content._contents["__fast_mass"] = corr * jet.mass.content
 
 
 # the class below does use hacks of JaggedCandidateArray to achieve the desired behavior
```
nsmith-: Is this file still in use now that JCA and ak0 are discontinued?

nsmith- (changes requested):
My only real worry is possible cycles, some extra tests or avoiding using the user-supplied cache for the more macroscopic virtual arrays (e.g. the delayed unc. builder) would be good

coffea/jetmet_tools/JetTransformer.py
lgray: deleted

coffea/jetmet_tools/CorrectedMETFactory.py
lgray: added non-user cache (fails the test with user cache so I suspect it is safe)

coffea/jetmet_tools/CorrectedJetsFactory.py
lgray: Replaced with non-user cache.

nsmith-: I think not specifying cache works too? More safe since `awkward._util` is semi-private

nsmith- (approved):

PR #486, opened 2021-04-07
use packaging to check parsl versions

lgray:


tests/test_parsl.py
```diff
@@ -4,8 +4,9 @@
 import multiprocessing
 import sys
 import pytest
+from packaging import version
 
-if sys.version_info.major == 3 and sys.version_info.minor >= 8 and sys.platform.startswith("darwin"):
+if version.parse(sys.version.split()[0]) >= version.parse('3.8') and sys.platform.startswith("darwin"):
```
nsmith-, line 9: What's up with the other method?

lgray: it's still running parsl on darwin?

lgray: checking if it's this or the 'startswith'

lgray: oh - that's weird... the CI that failed waiting for parsl to run now passes and doesn't run it... what....

lgray: however there is a ubuntu that stalled on parsl for py 3.8.... hmm.

lgray:
not needed.

PR #488, opened 2021-04-08
WIP: ServiceX processor

BenGalewsky:
This PR integrates the asynchronous serviceX processor into the _Coffea_ repo.

It represents a slightly different usage pattern since we have access to a streaming source of event data. 

This streaming interface is provided by the `DataSource` class. The constructor to this class accepts a func_adl query, a list of ServiceX datasets and optional metadata. It produces an aynchio stream of parquet files that are the result of the ServiceX transform on the provided dataset and selection statements. Users of this class can request a stream of paths to locally downloaded files or URLs to files on the ServiceX Minio object store.

We've created a simple `Analysis` base class that analyzers subclass to create the physics code to get their work done. We've boiled out almost all of the standard code to make an analysis tightly focused. This meant subclassing `dict_accumulator` to inject the standard setup.

Finally, we created a base class and a series of subclasses for an `Executor` - the contract of the executor is to consume files from the `DataSource` instance and invoke asynchronous calls to the analysis code on a backend of choice. The return value is an asynchio stream of histogram objects.  It performs some last-minute packaging to further reduce duplicated code that the analyzer might have to include. 

## Still to Come
~~Need to write unit tests~~
More documentation
Try this out on coffea-casa

## Outstanding Questions
1. Is this the right location for these classes?
2. My IDE includes the IRIS-HEP copyright on new files. I wonder if files contributed by IRIS-HEP should include that?
3. I put the example notebooks in the binder folder, but I don't know how these will run in binder. We have to provide a `.servicex` file somehow.


lgray:
@BenGalewsky can you rebase this on the latest master at some point? I can't seem to merge in today's additions.

lgray:
FYI we're fine with justified `# noqa: F401` lines in case you load stuff dynamically later.
Just write a comment to justify the exception. :-)

lgray:
FYI it looks like #495 may actually fix the random failures with the spark and parsl unit tests.

coffea/processor/servicex/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from .data_source import DataSource  # noqa F401
+from .sx_qastle import FuncAdlDataset, sx_event_stream  # noqa F401
+from .analysis import Analysis  # noqa F401
+from .accumulator import Accumulator  # noqa F401
+from .local_executor import LocalExecutor  # noqa F401
```
lgray: if you define 
```python3
__all__ = ["DataSource", ...]
```
it'll get rid of the need for the noqa's

lgray:
Hmm it has spent 1h installing in py39. What does SX drag along with it?

lgray:
It is also OK to not test in all versions on all platforms if it's significantly painful to do so.
Right now we don't test quite a few things on windows, parsl is very flakey depending on OSX and py version, spark 2.4 simply doesn't work on python >= 3.8, etc...

But, if you can make your stuff work everywhere - awesome. :-)

BenGalewsky:
> Hmm it has spent 1h installing in py39. What does SX drag along with it?

Not much at all, just some asynchronous I/O libraries and a little abstract syntax tree manipulation. Let me try building locally and see what the hangup could be

BenGalewsky:
Looks like some of the func_adl libraries don't offer support for 3.9 - rather than hack up the CI job to omit installing serviceX only in 3.9, I deactivated tests for serviceX in the CI build. I'll work with Gordon to get us working in 3.9 and update then

lgray:
@BenGalewsky I'd still recommend testing things and continuing to kick this down the field. You may go the route of SkyHook/Workqueue and add your own separate (non-required) tests in a python version that is suitably comfy.

coffea/processor/servicex/__init__.py
```diff
@@ -26,8 +26,17 @@
...
-from .accumulator import Accumulator  # noqa F401
-from .local_executor import LocalExecutor  # noqa F401
+from .data_source import *
+from .sx_qastle import *
+from .analysis import *
+from .accumulator import *
+from .local_executor import *
+
+__all__ = [
+    DataSource,
```
lgray: These names should be in quotes, `__all__` is expected to be a list of strings!

coffea/processor/servicex/accumulator.py
```diff
@@ -0,0 +1,42 @@
...
+
+from coffea.processor.accumulator import dict_accumulator, defaultdict_accumulator
+
+
+class Accumulator(dict_accumulator):
+    """
+    Simplified interface to Coffea accumulator. Sets up the most basic form automatically
+    """
+
+    def __init__(self, cats: Dict[str, Any]):
```
lgray: Migrate this to the new accumulator semantics.
https://github.com/CoffeaTeam/coffea/blob/master/tests/test_processor.py#L94-L111

coffea/processor/servicex/analysis.py
```diff
@@ -0,0 +1,54 @@
...
+class Analysis(ABC):
+    def __init__(self):
+        """
+        Concrete subclasses of this should construct the accumulator
+        """
+        self.accumulator = None
+
+    @staticmethod
+    @abstractmethod
+    def process(output: dict_accumulator, events: NanoEvents) -> dict_accumulator:
```
lgray: New accumulator semantics.

coffea/processor/servicex/dask_executor.py
```diff
@@ -0,0 +1,64 @@
...
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+from typing import Optional
+from dask.distributed import Client
+from .executor import run_coffea_processor, Executor
+
+
+class DaskExecutor(Executor):
```
lgray, line 33: What functionality/value is added here compared to the dask executor we already have, and what needs to be added there so we can share the executor implementation?

It would be good to maintain `run_uproot_job` as the most typical entrypoint.

lgray: Ah, right most of the async stuff.
That would require a large refactor, but we kinda wanted to do that anyway.

coffea/processor/servicex/schema.py
```diff
@@ -0,0 +1,78 @@
+from typing import Any, Dict
```
lgray: This should be inside of the nanoevents package, rather than servicex.

lgray:
I had a few minor comments, otherwise it looks fine for now. 
We could probably do with fewer tests while we're waiting for func_adl to get into py3.9... 
That's about it.

@nsmith- take a look as well?

coffea/processor/servicex/schema.py
```diff
@@ -0,0 +1,78 @@
...
+            record = _build_record_array(
+                c_name,
+                mapping,
+                contents,
+                "PtEtaPhiMCandidate" if is_4vector else "NanoCollection",
+            )
+            record["parameters"].update({"collection_name": c_name})
+            output[c_name] = record
+
+        self._form["contents"] = output
```
nsmith-: Here there should be a behavior specification for the schema. Since you use PtEtaPhiMCandidate and NanoCollection, I believe
```python
    @property
    def behavior(self):
        """Behaviors necessary to implement this schema"""
        from coffea.nanoevents.methods import base, candidate

        behavior = {}
        behavior.update(base.behavior)
        behavior.update(candidate.behavior)
        return behavior
```
should be sufficient

.github/workflows/ci.yml
```diff
@@ -140,6 +140,32 @@ jobs:
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install -q -e .[dev,servicex]
+        python -m pip list
+    - name: Test with pytest
+      env:
+        ARROW_PRE_0_15_IPC_FORMAT: 1
```
nsmith-, line 165: I've forgotten what this magic env var does, @lgray ?

coffea/processor/servicex/analysis.py
nsmith-: Why the different interface? Granted, its probably nicer in the long run to have a mechanism for the user code to add to an existing accumulatable object if it is available, but the current strategy seems to work for most everyone (the accumulators don't take much time to construct)

coffea/processor/servicex/executor.py
```diff
@@ -0,0 +1,197 @@
...
+            "This could be inline, but python 3.6"
+            x = await r
+            return x
+
+        finished_events = aiostream.stream.map(func_results, inline_wait, ordered=False)
+        # Finally, accumulate!
+        # There is an accumulate pattern in the aiostream lib
+        output = analysis.accumulator.identity()
+        async with finished_events.stream() as streamer:
+            async for results in streamer:
```
nsmith-: Since you want to yield the partial result each step, you'll have to lift https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/accumulator.py#L83-L97 and convert it to async (I think it would be enough to just change `next(gen)` to `await next(gen)` and then add a `yield` after line 94

coffea/processor/servicex/executor.py
```diff
@@ -0,0 +1,197 @@
...
+            result_file_stream, analysis.accumulator, analysis.process
+        )
+
+        # Wait for all the data to show up
+        async def inline_wait(r):
+            "This could be inline, but python 3.6"
+            x = await r
+            return x
+
+        finished_events = aiostream.stream.map(func_results, inline_wait, ordered=False)
```
nsmith-, line 76: Do you know the lifecycle of the futures here? Will they be cleaned up as soon as we're done adding them to `outpput`? If not, there will be a good chance of a memory explosion for large outputs. Further, since everything is async, you'll want to put some throttling/backpressure semaphores in this flow so that results don't pileup somewhere.

coffea/processor/servicex/executor.py
```diff
@@ -0,0 +1,197 @@
...
+            The analysis instance being accumulated
+        :param hist_stream:  AsyncGenerator[Accumulator]
+            Stream of histograms from each process task
+        :return:
+        """
+        # There is an accumulate pattern in the aiostream lib
+        output = analysis.accumulator.identity()
+        async with hist_stream.stream() as streamer:
+            async for results in streamer:
+                print("$$$ ", results)
```
nsmith-: 💰💰💰

coffea/processor/servicex/executor.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    # Since we execute remotely, explicitly include everything we need.
+    import awkward as ak
+    from coffea.nanoevents import NanoEventsFactory
+    from coffea.processor.servicex.schema import auto_schema
+
+    # This in is amazingly important - the invar mass will fail silently without it.
+    # And must be done in here as this function is shipped off to the funcx processor
+    # on a remote machine/remote python environment.
+    from coffea.nanoevents.methods import candidate
```
nsmith-: This can/should be removed once my comment on the schema is implemented. Though I'm surprised its "silent"--awkward has since at least a little while produced an error on addition of record arrays without an explicit overloadd.

coffea/processor/servicex/accumulator.py
nsmith-: I think once that is done this whole file can go away

nsmith- (changes requested):

.github/workflows/ci.yml
lgray: That's for spark, since it uses the arrow ipc format from before arrow v0.15

coffea/processor/servicex/accumulator.py
BenGalewsky: I'll delete this for now - I'd like to have a conversation about the trade-off of the flexibility of using a generic python interface (Accumulatable) vs the usability of a specific base class. Returning a dictionary of histograms is a bit harder to learn than returning an instance of a class with clear constructors.

nsmith-: With the new style it is also possible to return a single hist, and in general any nested dictionary of objects implementing add. This makes coffea maintenance easier because we don't have to create a wrapper/subclass/specialization/etc. for outside things like `hist` package objects. If it can be added, we'll add it up!

coffea/processor/servicex/analysis.py
nsmith-: Thanks for implementing, though looks like the example notebooks still use the old syntax?

gordonwatts:
> I had a few minor comments, otherwise it looks fine for now.
> We could probably do with fewer tests while we're waiting for func_adl to get into py3.9...
> That's about it.

Just seeing this. If you mean just the `func_adl` package, that is already at 3.9. The `servicex_frontend` isn't there yet because last time I attempted to run it I couldn't find `numpy` wheels - but this was before my teaching started. I can prioritize that - I suspect it will just work in 3.9. Let me know what you are specifically refering to here.

nsmith- (approved):
I'll be interested to see how the memory behavior plays out at scale with aiostream. If it is doing a good job of dropping references early that would be great news and maybe we can refactor the regular processor to use it.

Overall there's a lot of bifurcation here: 
* `processor.servicex.executor.run_coffea_processor` is akin to `processor.processor._work_function`
* `processor.servicex.executor.Executor` is akin to `run_uproot_job`

In the long run we'll need to harmonize (likely in the direction of objects representing analysis state, input state, output state, and executor state) but for now it can be considered an experimental interface

PR #489, opened 2021-04-08
Fix all instances of F401 and comment about exceptions

lgray:
As in title.


lgray:
@andrzejnovak can you check this PR for screwing things up with bokeh, just to make sure there's nothing magical going on there.

PR #502, opened 2021-04-13
Allow users to supply more metadata, and pass it to executors.

### LINKED ISSUE #479: Passing custom fileset metadata through run_uproot_job
The `fileset` argument in [`coffea.processor.run_uproot_job`](https://coffeateam.github.io/coffea/api/coffea.processor.run_uproot_job.html#coffea.processor.run_uproot_job) is used to specify dataset-specific information, like the dataset name and the name of the tree to use:
```python
fileset = {
    "signal": {treename": "signal", "files": ["mc.root"]}
}
```
This is available as metadata in the processor via e.g.
```python
def process(self, events):
    events.metadata["dataset"]
    events.metadata["treename"]
```
and allows dataset-dependent processing.

It would be convenient to allow passing additional key-value pairs through this structure. At the moment, additional data like
```python
"signal": {"treename": "signal", "files": ["mc.root"], "new_key": "value"}
```
will not be passed through. It is possible to pass additional custom metadata like this to the processor constructor to then have it available in `process()`, but this requires defining a second object with a very similar structure.

If feasible, passing custom fileset metadata through to the processor automatically so the metadata is available when processing.

**Describe alternatives you've considered**
Writing a modified version of `run_uproot_job` or passing the custom metadata to the processor constructor.

**Additional context**
related: #369

lgray:
Fixes #479 

Still a bit of a WIP.
At the very least this implementation does pass the metadata to the processors correctly.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -35,6 +35,19 @@ from collections.abc import Mapping, MutableMapping
 _PICKLE_PROTOCOL = pickle.HIGHEST_PROTOCOL
 DEFAULT_METADATA_CACHE: MutableMapping = LRUCache(100000)
 
+_PROTECTED_NAMES = {
+    "dataset",
+    "filename",
+    "treename",
+    "metadata",
+    "entrystart",
+    "entrystop",
+    "fileuuid",
+    "numentries",
+    "uuid",
+    "clusters",
+}
+
 
 class FileMeta(object):
     __slots__ = ["dataset", "filename", "treename", "metadata"]
@@ -74,7 +87,7 @@ class FileMeta(object):
     def chunks(self, target_chunksize, align_clusters):
         if not self.populated(clusters=align_clusters):
             raise RuntimeError
-        user_keys = set(self.metadata.keys()) - {"numentries", "uuid"}
+        user_keys = set(self.metadata.keys()) - _PROTECTED_NAMES
         user_meta = {k: self.metadata[k] for k in user_keys}
         if align_clusters:
             chunks = [0]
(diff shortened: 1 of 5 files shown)


lgray:
@nsmith- take a look. Lemme know if there's a better place to put things.

lgray:
also @alexander-held this is what you were more or less asking for right?

alexander-held:
Yes, this looks exactly like what I had in mind. Thanks for the work on this!

lgray:
Cool, could you give this branch a try and see if you immediately run into some problem?
I'd like to implement this more correctly, but if it works as it is now and doesn't break things... Well why not.

alexander-held:
I gave this a try and it worked as expected. I find this quite useful for bookkeeping, as it allows me to gather metadata in one central place as opposed to having to pass it through the processor constructor.

lgray:
@nsmith- you ok with the hacky version for now?

nsmith-:
Can we encapsulate the metadata instead of putting it at top level at least?
e.g.
```python
fileset = {
    "mydataset": {
        "files": ["blah.root"],
        "metadata": {"thing1": 1},
    }
}
```

lgray:
Sure, had been on the fence about it. Makes some things cleaner too.

alexander-held:
I guess the access would still be `events["metadata"]["thing1"]` instead of `events["metadata"]["metadata"]["thing1"]`?

nsmith-:
Right, in that sense it is a little confusing since dataset is set (presumably if "dataset" were specified in the metadata it would override the current value?)

lgray:
can just make dataset a protected name...

lgray:
@nsmith- this look ok-enough for you?
@alexander-held please give this a try real quick.

Now you do:
```python3
filelist = {
        "DummyBad": {"treename": "Events", "files": [osp.abspath("tests/samples/non_existent.root")]},
        "ZJets": {"treename": "Events", "files": [osp.abspath("tests/samples/nano_dy.root")],
                  "metadata": {"checkusermeta": True, "someusermeta": "hello"} },
        "Data" : {"treename": "Events", "files": [osp.abspath("tests/samples/nano_dimuon.root")],
                  "metadata": {"checkusermeta": True, "someusermeta2": "world"} }
    }
```

and it will complain at you if you try to use any reserved words.

alexander-held:
I gave this a try, it worked as expected. One minor suggestion: In the error message
```
ValueError: Reserved word "treename" in fileset dictionary, please rename this entry!
```
it might be useful to include "metadata"? Otherwise it may sound as if the top-level `"treename": "Events"` is to blame.

lgray:
@nsmith- good from your side?

nsmith- (approved):

PR #505, opened 2021-04-14
wq executor: do not transfer root files with absolute paths

btovar:
Such paths are assumed to be on a shared filesystem, thus they do not
need to be transfered to the execution site.


btovar:
@lgray Thanks Lindsey!

PR #519, opened 2021-04-28
wq executor: send and set X509 user proxy

btovar:
adds x509_proxy keyword to wq executor arguments to indicate the local
path of the X509 proxy to be sent along the tasks.

If set to None (the default), use the value of the X509_USER_PROXY
environment variable if set, or the file /tmp/x509up_u${UID} if exits.
If False, no proxy is sent.


lgray:
@btovar - the rados job is failing for other reasons. Would you like me to merge this or are you still workin' on it?

btovar:
It is good to go, thanks @lgray!

Issue #515, opened 2021-04-26
Add recoverable option implementation for dask


lgray:
This is a more general solution to #317 and #443.

@jrueb @andrzejnovak 

Add a new opt-in feature that lets us always finish jobs, and reports failed chunks in a way that can be resubmitted and easily bookkept.

* reports files/chunks of bad jobs, instead returning a list of chunks and reasons in the `metrics` accumulator, always completes
* has some basic quality metrics to let people know if it's worth rerunning 
  * i.e. easy to know if fail chunks / total chunks for a dataset is small

In the future: 
* returned list of failed chunks can be given to `run_uproot_job` to be rerun. (optional for now, may need more aggressive changes)

andrzejnovak:
Few points from recent experiences
- dask and parsl handle submission differently, which is not gonna be apparent to a new user and they have different caveats
  - dask submits reduce jobs, but for non-trivial accumulator sizes, these might need much more memory than the default processing job, so dask is stuck in last few percent just trying to resubmit them
  - parsl doesn't submit reduce jobs, instead, all chunks are added on the submit node which becomes a problem with more inputs files e.g. in my case 40k files/chunks the processing itself is a fraction of the total runtime (10%) and the rest is spent on the submit node just merging accumulators, the memory problem doesn't hit in this case since the submit node doesn't share resources, but it could be a problem if the submit node was weaker

andrzejnovak:
- executors could automatically save intermediate output if the accumulator grows too large for memory
  - maybe just warn by default, with an opt in flag

andrzejnovak:
This is now available in futures/parsl as `recoverable=True`

lgray:
This will be fixed in coffea 2023 with the report option that'll be available in uproot/dask-awkward.

PR #521, opened 2021-05-04
feat: batching for parsl/futures execs

andrzejnovak:
- As mentioned #515 doing only one reduce on submit node is a massive bottleneck for large (full analysis) runs, or in general when enough nodes are available to process worker functions faster than merging the outputs.
-  This implementation fixes that for `futures`/`parsl` by submitting batches of reduce jobs as the results become available
- Separate tracking (tqdm) for map/reduce jobs
- chunking is currently adaptive with `minsize`, `maxsize`, `nsplits` of finished jobs, this could/should be made an executor arg
- this implementation replaces the single node generator based merge from `_futures_handler` rendering the intermediate results on the submit node, this should however be in principle fine as the intermediate should be relatively small and not stay in mem on the submit node. Finally, the mem cost of rendering each batch should be tweakable by adjusting `maxsize`

### DIFF
--- a/coffea/processor/accumulator.py
+++ b/coffea/processor/accumulator.py
@@ -5,9 +5,6 @@ from collections import defaultdict
 from collections.abc import MutableMapping, MutableSet
 from typing import Iterable, Optional, TypeVar, Union
 
-import pickle
-import warnings
-
 try:
     from typing import Protocol, runtime_checkable  # type: ignore
 except ImportError:
@@ -95,9 +92,6 @@ def accumulate(
         while True:
             # subsequent additions can happen in-place, which may be more performant
             accum = iadd(accum, next(gen))
-            _size = len(pickle.dumps(accum)) / (1024 * 1024)
-            if _size > 4000:  # 2 GB
-                warnings.warn("Accumulator memory size > 4GB. Might cause hangups.")
     except StopIteration:
         pass
     return accum
--- a/coffea/processor/accumulator.py
+++ b/coffea/processor/accumulator.py
@@ -5,6 +5,9 @@ from collections import defaultdict
 from collections.abc import MutableMapping, MutableSet
 from typing import Iterable, Optional, TypeVar, Union
 
+import pickle
+import warnings
+
 try:
     from typing import Protocol, runtime_checkable  # type: ignore
 except ImportError:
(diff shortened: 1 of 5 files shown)


andrzejnovak:
@nsmith- @lgray can you take a look?

andrzejnovak:
I wanna say the last test is not my fault :)

nsmith-:
I'm suspicious that merging results does not typically take longer than transporting them between workers and client, except in certain cases. With that in mind, I'd rather not double the roundtrip time for futures and parsl executors.

Or put another way, for `futures_executor` we don't really expect the worker to client ratio to be much more than O(10), so as long as t_merge*10 < t_process the client process should be able to keep up with the output stream. In the parsl case, the ratio may be much larger, but there I think the solution is inter-worker communication to bring sets of results together in the the worker pool directly and do a tree reduction, as is done in dask.

I think we need to investigate in more detail why `hist` merging is slow and if can be sped up.

andrzejnovak:
I agree this would probably not be necessary for single machine futures, though at the same time it shouldn't hurt the performance given everything is at the same place.

Running a totally not scientific with the example `MyZPeak()` processor

andrzejnovak:
I agree this shouldn't really be needed for single machine futures, but I also don't expect it to noticeable change performance there (Also I expect most jobs via `future_executor` to be small-ish so it doesn't really matter either way.

Running a totally unscientific test with master and this branch with parsl scaling out to single machine futures, running ~30 files, ~500 chunks and the example `MyZPeak` processor and checking each twice I get:

 `1:20 | 1:30` for `master` and `1:20 | 1:40` for this branch. 

For a larger number of input chunks ~800. I get 

` 2:16 | 2:50` for `master` and for this branch `2:20 | 2:30`

It doesn't seem like there's a penalty to pay with ~short/medium jobs either.

I am not sure I understand the dask setup well enough, but I did try passing the completed futures to the parsl wrapped reduce job to only materialize them in the job itself. But it doesn't seem possible to pass the future objects into another future.

andrzejnovak:
Actually scaling out the SLURM starts to make a difference though. With ~170 files, ~2500 chunks and two nodes, it's `2:20` minutes on master, and only `1:20` for the new branch. <- this is reporting the tqdm times, not the cell times to not mix in waiting for node allocation

andrzejnovak:
Hmm. I am gonna close this for now. It seems that updating `parsl` to `1.1` which I did to fix some serialization problems while working on this PR, actually fixed the performance issues at scale that I saw in master.

While the slowdown/freeze originally did occur at the "merging histogram" step, it seems it was not actually due to histogram merges, but likely an underlying issue with the serialization which was changed from parsl `1.0` to `1.1`. 

For posterity, adding a large number of copies of non-trivially small (though not extremely large) accumulators by itself is not prohibitively time-consuming and neither is there any meaningful difference in speed between `hist` and `coffea/hist` based histograms in the accumulators. 

image

PR #524, opened 2021-05-06
fix: update parsl dependency for improved serialization

andrzejnovak:
See #521 for discussion.


nsmith-:
@lgray you tried already bumping parsl to the pre-release of 1.1, and saw some issue right?

lgray:
Ah, the pre-release of parsl was telling me more about the network adapter binding issues in GHA, other than that nothing special. I'm fine to bump us to the pre-release for an extra, it's only the strict dependencies I want to keep a rule for.

Issue #514, opened 2021-04-23
Follow up on #448 - DAOD Schema


lgray:
Now that #448 is done, implement a suitable `nanoevents` schema for DAOD_PHYSLITE files.

@nikoladze

lgray:
done.

PR #527, opened 2021-05-12
PHYSLITE schema

nikoladze:
This addresses #514. I started implementing a schema for `DAOD_PHYSLITE`. So far it's just zipping together the collections and assigning some basic LorentzVector behavior which i just copied over from my [prototype](https://gitlab.cern.ch/nihartma/physlite-experiments/-/blob/master/physlite_experiments/physlite_events.py).

So far the following works:

```python
from coffea.nanoevents.schemas import PHYSLITESchema
from coffea.nanoevents.factory import NanoEventsFactory

# file from https://cernbox.cern.ch/index.php/s/xTuPIvSJsP3QmXa/download
    "DAOD_PHYSLITE.art_split99.pool.root",
    treepath="CollectionTree",
    schemaclass=PHYSLITESchema
)
events = factory.events()
```

```pycon
>>> events.Electrons
<xAODParticleArray [[], [], ... truthType: 17}], []] type='100 * var * xAODParti...'>
>>> events.Electrons.pt
<Array [[], [], [9.46e+04, ... [6.91e+03], []] type='100 * var * float32[paramet...'>
>>> events.Electrons.px
<Array [[], [], [-9.39e+04, ... [6.81e+03], []] type='100 * var * float32'>
>>> events.GSFTrackParticles
<xAODTrackParticleArray [[], ... z0: 0.426}]] type='100 * var * xAODTrackParticl...'>
>>> events.GSFTrackParticles.pt
<Array [[], [9.46e+03, ... 3.11e+03, 2.44e+03]] type='100 * var * float32'>
```

Besides refining the behavior and adding other stuff like EventInfo the next big thing will be the cross references. From looking over the code for NanoAOD it looks like everything is there to do basic indexing from one collection into another, even with nested indices (that result in a double-jagged array). I just need to figure out how to best create the global indices. Maybe i'll have to add or modify some of the transforms to get them since i don't get the local indices from counts, but need to pop them out of already existing awkward arrays that uproot created.

Example for a local index, resulting in a single jagged array:

```pycon
>>> events.Muons["combinedTrackParticleLink.m_persIndex"]
<Array [[], [0], [], [], ... 0], [0], [0, 1]] type='100 * var * uint32[parameter...'>
```
which should link into `events.CombinedMuonTrackParticles`, e.g. without global indices that would be something like
```pycon
>>> muon_track_link = events.Muons["combinedTrackParticleLink.m_persIndex"]
>>> muon_track_key = events.Muons["combinedTrackParticleLink.m_persKey"]
>>> events.CombinedMuonTrackParticles[muon_track_link.mask[muon_track_key != 0]]
<xAODTrackParticleArray [[], ... z0: 12.5}]] type='100 * var * ?xAODTrackParticl...'>
```

Example for a local (nested) index, resulting in a double jagged array:

```pycon
>>> events.Electrons.trackParticleLinks.m_persIndex
<Array [[], [], [[0]], ... [], [[0, 0]], []] type='100 * var * var * uint32'>
```

which should link into `events.GSFTrackParticles`. I have [some example for that](https://gitlab.cern.ch/nihartma/physlite-experiments/-/blob/6f11b90cfb90216282467a09d0356d0a729421d5/physlite_experiments/behavior.py#L13), but i hope that this could also be bent somehow into the `nestedindex` stuff that already exists.

A bit more advanced would then be to link into multiple collections, resulting in a `UnionArray`. In principle, all these ElementLinks have an `m_persKey` field that contains hashes of the collections this index links into. In most cases (like the examples above) this is always the same, but in some cases it can be several, e.g.

```pycon
>>> import numpy as np
>>> import awkward as ak
>>> np.unique(ak.to_numpy(ak.flatten(events.TruthElectrons.parentLinks.m_persKey, axis=None)))
array([        0, 375408000, 614719239, 660928181, 779635413],
      dtype=uint32)
```

which are the hashes for `TruthTaus`, `TruthBoson`, `TruthTop`, `TruthBottom`. I think 0 should be masked out since they probably correspond to links to particles that don't exist anymore (maybe thinned out).

I'll try out a few more things, but let me know if you have ideas or suggestions how to do these things.


lgray:
This might be an interesting place to do unit conversions from MeV to GeV. ;-)

lgray:
Aside from the inter-experiment snark. This is awesome, I'll have a careful look through it. :-)

coffea/nanoevents/methods/physlite.py
```diff
@@ -0,0 +1,47 @@
+import awkward as ak
+import numpy as np
+from coffea.nanoevents.methods import vector
+
+
+behavior = {}
+behavior.update(vector.behavior)
+
+
+@ak.mixin_class(behavior)
```
lgray: Please use full package names in library code. `np -> numpy`, `ak -> awkward`.

coffea/nanoevents/schemas/physlite.py
```diff
@@ -0,0 +1,50 @@
+import warnings
+from collections import defaultdict
+from coffea.nanoevents.schemas.base import BaseSchema, zip_forms
+
+
+class PHYSLITESchema(BaseSchema):
+
+    mixins = {
+        "Electrons": "xAODParticle",
```
lgray: Are there no further specializations to electrons/muons/etc. that you'd like to do? It's still a draft so I suppose you may be working on that.

nikoladze: Yes, there are going to be more specializations - e.g. the cross references will be different for Electrons/Muons/Jets

coffea/nanoevents/transforms.py
```diff
@@ -125,7 +125,7 @@ def local2global(stack):
     index = stack.pop()
     index = index.mask[index >= 0] + target_offsets[:-1]
     index = index.mask[index < target_offsets[1:]]
-    out = numpy.array(awkward.flatten(awkward.fill_none(index, -1)))
+    out = numpy.array(awkward.flatten(awkward.fill_none(index, -1), axis=None))
```
nikoladze, line 128: added the `axis=None` here to support indices that are already more than single-jagged

coffea/nanoevents/transforms.py
```diff
@@ -335,3 +335,7 @@ def nestedindex(stack):
 def item(stack):
     field = stack.pop()
     stack.append(stack.pop()[field])
+
+
+def to_numpy(stack):
```
nikoladze: need this to put offsets that are directly popped of an existing `ListOffsetArray` into `local2global`

coffea/nanoevents/schemas/physlite.py
```diff
@@ -17,31 +19,87 @@ class PHYSLITESchema(BaseSchema):
...
+    def _create_global_index_form_elementlink(base_form, key, linkto_key):
+        form = copy.deepcopy(base_form)
+        record = form["content"]["content"]["contents"]
+        form["content"]["content"] = record["m_persIndex"]
+        form["content"]["content"]["form_key"] = quote(
+            f"{key},!load,m_persIndex,!item,{linkto_key},!load,!offsets,!to_numpy,!local2global"
+        )
+        form["content"]["form_key"] = quote(
+            # the !data is a hack to avoid having the same key as the actual elementlink
+            f"{key},!load,!content,!data"
```
nikoladze: i don't quite understand yet what's going on here - if i create a form where i e.g. have `events.Electrons.trackParticleLinks`  and `events.Electrons.trackParticleLinksG` which both have the same `form_key` on top level (which should be used only for the offsets), then the same array will come out, even if the `form_key` in the inner levels differs.

coffea/nanoevents/schemas/physlite.py
```diff
@@ -17,31 +19,87 @@ class PHYSLITESchema(BaseSchema):
...
+        form["content"]["primitive"] = "int64"
+        return form
+
+    @staticmethod
+    def _create_global_index_form_elementlink(base_form, key, linkto_key):
+        form = copy.deepcopy(base_form)
+        record = form["content"]["content"]["contents"]
+        form["content"]["content"] = record["m_persIndex"]
+        form["content"]["content"]["form_key"] = quote(
+            f"{key},!load,m_persIndex,!item,{linkto_key},!load,!offsets,!to_numpy,!local2global"
```
nikoladze: for the double-jagged  cross references i have to pop off the `m_persKey` since these are the branches that can't be split

coffea/nanoevents/schemas/physlite.py
```diff
@@ -17,31 +19,87 @@ class PHYSLITESchema(BaseSchema):
...
                 )
             except NotImplementedError:
                 warnings.warn(f"Can't zip collection {objname}")
         return contents
 
+    @staticmethod
+    def _create_global_index_form(base_form, key, linkto_key):
+        form = copy.deepcopy(base_form)
+        form["content"]["form_key"] = quote(
+            f"{key},!load,{linkto_key},!load,!offsets,!to_numpy,!local2global"
```
nikoladze: here i use the offsets from the original cross references array (that is already a `ListOffsetArray`), convert to numpy and the put them into `local2global`

lgray:
Do you have like a 40-event DAOD file that we can put into tests?

Then we can automate running test processors and such on it.

nikoladze:
I experimented a bit more to get cross references working (added some comments inline). I think it should work in principle - probably the way i implemented it can be improved. Let me know if you have ideas.

But there is another issue now - i get the wrong integer types for my ElementLink arrays. If i ask uproot for the akward form of this branch:

```python
# file from https://cernbox.cern.ch/index.php/s/xTuPIvSJsP3QmXa/download
tree = uproot.open("DAOD_PHYSLITE.art_split99.pool.root:CollectionTree")
tree["AnalysisElectronsAuxDyn.trackParticleLinks"].interpretation.awkward_form(None)
```

it will give me `uint32` for the record fields:

```
...
"content": {
            "class": "RecordArray",
            "contents": {
                "m_persIndex": "uint32",
                "m_persKey": "uint32"
            },
...
```

But when i actually retrieve that array with uproot it's suddenly `int64`:

```python
tree["AnalysisElectronsAuxDyn.trackParticleLinks"].array().layout.form
```
```
{
    "class": "ListOffsetArray64",
    "offsets": "i64",
    "content": {
        "class": "ListOffsetArray64",
        "offsets": "i64",
        "content": {
            "class": "RecordArray",
            "contents": {
                "m_persIndex": "int64",
                "m_persKey": "int64"
            },
            "parameters": {
                "__record__": "ElementLink<DataVector<xAOD::TrackParticle_v1>>"
            }
        }
    }
}
```

Any ideas how to circumvent this?

nikoladze:
> Do you have like a 40-event DAOD file that we can put into tests?
> 
> Then we can automate running test processors and such on it.

The file from https://cernbox.cern.ch/index.php/s/xTuPIvSJsP3QmXa/download has 100 events and 2.8MB. If that's still too large i could also make a smaller one.

lgray:
> > Do you have like a 40-event DAOD file that we can put into tests?
> > Then we can automate running test processors and such on it.
> 
> The file from https://cernbox.cern.ch/index.php/s/xTuPIvSJsP3QmXa/download has 100 events and 2.8MB. If that's still too large i could also make a smaller one.

That's fine we have other ~2MB files in `test/samples`.

nikoladze:
About the issue i mentioned before where the Link indices and keys become `int64` although the awkward form has `int32`: reading a bit through the [code of `uproot.lazy`](https://github.com/scikit-hep/uproot4/blob/main/src/uproot/behaviors/TBranch.py#L646) i noticed there is a special treatment for forms where the arrays are fetched via `ak.from_iter`, so that's probably something i would need as a temporary solution here as well until these branches are read directly, maybe with Awkward Forth in the future.

For now i just hardcoded the form for the element links to have `int64`. I also added a small file in `tests/samples` with 40 events.

Now i can have cross references using global indices

```python
import os
    os.path.abspath("tests/samples/DAOD_PHYSLITE_21.2.108.0.art.pool.root"),
    treepath="CollectionTree",
    schemaclass=PHYSLITESchema
)
events = factory.events()
```

```pycon
>>> events.Electrons.trackParticles
<xAODTrackParticleArray [[], [], ... vz: 0, z0: -0.0982}]]] type='40 * var * var...'>
>>> events[30:].Electrons.trackParticles.pt.tolist()
[[], [], [], [], [], [[100500.9765625]], [], [], [[36525.61328125]], [[27814.41015625, 18714.423828125]]]
>>> events[32:].Electrons.trackParticles.pt.tolist()
[[], [], [], [[100500.9765625]], [], [], [[36525.61328125]], [[27814.41015625, 18714.423828125]]]
```

I also experimented a bit with Links into multiple collections and made a function that creates a union array from them

```
>>> events.TruthBoson.children
<Array [[[], ... status: 1}]]] type='40 * var * var * ?union[?xAODTruthParticle[...'>
```

which creates a union array of all the collections linked into by evaluating `np.unique` on the `events.TruthBoson.childLinks.m_persKey` array. The key-to-branchname mapping is currently hardcoded, but can be read in principle from the `MetaData/EventFormat` tree. Also unfortunately i had to put all potential combinations of links into the form because the global indices have to be created from the form ...

So i'm not quite sure if this is the best way to do it - at the moment i see it more as a proof of concept. Maybe there is in general a better way to deal with these links ...

Also i don't quite know how to attach behavior to the union arrays.

Finally one thing i haven't quite understood yet: the way i use the `_apply_global_index` function it seems the virtual arrays i link to get materialized, e.g. if i do

```pycon
events.Electrons.trackParticles.layout
```

there is no VirtualArray anymore

nikoladze:
I figured out a few more things:

The behavior for union arrays works if i add `__record__` on the top level, so now i can do the famous "cyclic" references in NanoEvents

```pycon
>>> events.TruthBoson.children
<TruthParticleArray [[[], ... TruthParticle]]] type='40 * truthParticle'>
>>> events.TruthBoson.children.parents
<TruthParticleArray [[[], ... [TruthParticle]]]] type='40 * truthParticle'>
>>> events.TruthBoson.children.parents.children
<TruthParticleArray [[[], ... TruthParticle]]]]] type='40 * truthParticle'>
>>> events.TruthBoson.children.parents.children.ndim
5
```

Note: needs the fix from scikit-hep/awkward-1.0#870

However, as you can see the overridden typestring (with `_set_repr_name` which i copied from the nanoaod behavior) does not work for anymore here since i had to add the `__record__` on top level. But the arrays seem to be created correctly (although i haven't really tested it yet), so for now it's more an aesthetic issue.

Also i found that adding the `collection_name` parameters will avoid the index arrays to be materialized. I have not quite understood how that happens though ...

Note on performance: In current uproot the loading of the ElementLink branches is quite slow, so as a temporary solution for playing with larger ROOT files i wrote [a small hack to hijack the `extract_column` function of `UprootSourceMapping`](https://gitlab.cern.ch/nihartma/physlite-experiments/-/blob/ae980c5b91ad7b4f5fc4c329936e669ef4d38c5a/physlite_experiments/deserialization_hacks.py#L467) to use my custom Numba deserialization functions, to be used like

```python
# pip install git+https://gitlab.cern.ch/nihartma/physlite-experiments.git
from physlite_experiments.deserialization_hacks import patch_nanoevents
patch_nanoevents()
```

nikoladze:
While testing the element links i found scikit-hep/awkward-1.0#876 - which is already fixed. After the fix the tests should pass.

lgray:
Nice a new awkward is build built. Once that is done baking we should require `awkward >=1.3.0` with this branch.

lgray:
@nikoladze Can you add the awkward array version requirement to setup.py in this PR? Thanks!

lgray:
Are you ready to switch this from a draft PR, and have it reviewed?

coffea/nanoevents/methods/physlite.py
```diff
@@ -0,0 +1,178 @@
...
+            self._events().GSFTrackParticles,
+            self.trackParticleLinks__G__GSFTrackParticles,
+            self.trackParticleLinks.m_persKey,
+        )
+
+    @property
+    def trackParticle(self):
+        trackParticles = self.trackParticles
+        return self.trackParticles[
+            tuple([slice(None) for i in range(trackParticles.ndim - 1)] + [0])
```
nsmith-, line 148: Could be `self.trackParticles[..., 0]`?
By the way, nice catch to guard against accessing attributes in objects that may be nested further than the default 2 dimensions.

coffea/nanoevents/schemas/physlite.py
nsmith-: I would be fine also with just letting `!local2global` accept Index64 types as well as NumpyArray types for input.

coffea/nanoevents/transforms.py
nsmith-: I assume here the link target collection is still single-jagged? Otherwise I don't think this algorithm works properly.

coffea/nanoevents/transforms.py
nsmith-: As mentioned before, I think its fine to wrap the input of `local2global` with `numpy.array()` to reduce the need for this extra command

nsmith- (changes requested):
This looks really nice! I have some proposed simplifications, but for the most part it looks good to go.

coffea/nanoevents/schemas/physlite.py
nsmith-: Shouldn't the `trackParticleLinksG` form have a key that at least includes `!local2global` whereas the `trackParticleLinks` would not?

coffea/nanoevents/schemas/physlite.py
nsmith-: Sounds fine to have a record item accessor, certainly so once awkward-forth is ready to go.

coffea/nanoevents/methods/physlite.py
nikoladze: That did not wok - awkward complains with `ValueError: ellipsis (...) can't be used on a data structure of different depths`. Should that work? Then i can open an issue for awkward array.

Actually i noticed it for lower number of dimensions e.g. when i do `events[2].Electrons.trackParticle`

nsmith-: I would think that as long as `array.ndim` is well-defined, ellipsis is as well (provided no field indexers are used in the getitem). @jpivarski thoughts?

jpivarski: That error message about ellipsis is talking about an issue with unions (so, unless it's not what I'm thinking, it's not a bug).

```python
>>> array = ak.Array([[1], [2], [3]])
>>> array[..., 0]
<Array [1, 2, 3] type='3 * int64'>
>>> array = ak.Array([[1], [2], [3], [[4], [4], [4], [4]]])
>>> array[..., 0]
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward/highlevel.py", line 993, in __getitem__
    return ak._util.wrap(self.layout[where], self._behavior)
ValueError: ellipsis (...) can't be used on a data structure of different depths

(https://github.com/scikit-hep/awkward-1.0/blob/1.2.2/src/libawkward/Content.cpp#L1491)
```

Arrays always have an `ndim`, and this is the "purelist depth". A union like the above doesn't have an easily definable depth. I _had thought_ that I made this return -1 in cases like that, but apparently, it gives the depth of the UnionArray object:

```python
>>> array.ndim                    # high-level
0
>>> array.layout.purelist_depth   # low-level
0
```

That's not ideal: it doesn't give more information than if it had been a RecordArray at that level. But I don't know if this addresses your question.

nsmith-:
> Also i don't quite know how to attach behavior to the union arrays.

This might be a bug, perhaps related to https://github.com/scikit-hep/awkward-1.0/issues/770
In principle a union array should be able to have a behavior.

> Finally one thing i haven't quite understood yet: the way i use the _apply_global_index function it seems the virtual arrays i link to get materialized, e.g. if i do
> 
> events.Electrons.trackParticles.layout
> 
> there is no VirtualArray anymore

This might be related to the lack of a `collection_name` in the layout. I think what is happening is that in https://github.com/CoffeaTeam/coffea/blob/1963fc9371552b348a15084f5bde9390be1e6e1c/coffea/nanoevents/methods/base.py#L36-L50 if no `collection_name` is encountered then `awkward._util.recursively_apply(self.layout, descend)` just returns the original (but now materialized) array. Whereas if `collection_name` is encountered, it would return the flat collection RecordArray without materializing its constituent fields. The purpose of that function is to descend through possible extraneous array nodes like VirtualArray since in general VirtualArray nodes in layouts can disappear in certain circumstances.

lgray:
@nikoladze if we can wrap this up today or tomorrow we can have it in the next coffea release - which would be pretty snazzy.

coffea/nanoevents/schemas/physlite.py
nikoladze: sounds good - i moved the `numpy.asarray` conversion to `local2global`

coffea/nanoevents/transforms.py
nikoladze: yes, all target collections i'm considering so far are single-jagged

coffea/nanoevents/methods/physlite.py
nikoladze: I looked a bit what's going on for this particular array. Seems the issue here is not that it is a union, but that one field has a higher dimension. There is `definingParametersCovarianceMatrix` which has one dimension more than the other fields (but is still zipped on the outermost level of offsets). So a small example for that situation would be

```python
import awkward as ak
array = ak.zip(
    {"a" : [[1], [2, 3]], "b" : [[[10, 10]], [[20, 20], [30, 30]]]},
    depth_limit=2
)
```
The resulting array actually has `ndim=2` - which is what i use in that code for `trackParticle`.
```pycon
>>> array.ndim
2
```
I want this:
```pycon
>>> array[:, 0].tolist()
[{'a': 1, 'b': [10, 10]}, {'a': 2, 'b': [20, 20]}]
```
The ellipsis don't work, presumably because of the one dimension more in the field `b`
```
>>> array[..., 0]
  File "/home/nikolai/.local/lib/python3.9/site-packages/awkward/highlevel.py", line 993, in __getitem__
    return ak._util.wrap(self.layout[where], self._behavior)
ValueError: ellipsis (...) can't be used on a data structure of different depths

(https://github.com/scikit-hep/awkward-1.0/blob/1.3.0/src/libawkward/Content.cpp#L1491)
```

coffea/nanoevents/schemas/physlite.py
nikoladze: yes, but since they both have one dimension more than the other fields of `Electrons` there is also a `form_key` on the level of the `ListOffsetArray`. So currently i have for `events.Electrons.layout.form`

```
{
    "class": "ListOffsetArray64",
    "offsets": "i64",
    "content": {
        "class": "RecordArray",
        "contents": {
            --------------------------------------
            <lots-of-single-jagged-fields>,
            --------------------------------------
            "trackParticleLinks": {
                "class": "VirtualArray",
                "form": {
                    "class": "ListOffsetArray64",
                    "offsets": "i64",
                    "content": {
                        "class": "RecordArray",
                        "contents": {
                            "m_persIndex": {
                                "class": "VirtualArray",
                                "form": {
                                    "class": "NumpyArray",
                                    "itemsize": 4,
                                    "format": "I",
                                    "primitive": "uint32",
                                    "form_key": "AnalysisElectronsAuxDyn.trackParticleLinks%2C%21load%2C%21content%2C%21content%2Cm_persIndex%2C%21item"
                                },
                                "has_length": true
                            },
                            "m_persKey" : <similar>
                       [...]
                    },
--->                "form_key": "AnalysisElectronsAuxDyn.trackParticleLinks%2C%21load%2C%21content"
                },
                "has_length": true
            },
            "trackParticleLinks__G__GSFTrackParticles": {
                "class": "VirtualArray",
                "form": {
                    "class": "ListOffsetArray64",
                    "offsets": "i64",
                    "content": {
                        "class": "NumpyArray",
                        "itemsize": 8,
                        "format": "l",
                        "primitive": "int64",
                        "form_key": "AnalysisElectronsAuxDyn.trackParticleLinks%2C%21load%2Cm_persIndex%2C%21item%2CGSFTrackParticlesAuxDyn.z0%2C%21load%2C%21offsets%2C%21local2global"
                    },
                    [...]
--->                "form_key": "AnalysisElectronsAuxDyn.trackParticleLinks%2C%21load%2C%21content%2C%21skip%2CGSFTrackParticlesAuxDyn.z0"
                },
                "has_length": true
            },
```
The form keys marked with `--->` are the ones i'm confused about. I would think they are only used for the offsets, but i'm not really sure how this works (there is no `!offset` after all? On the other hand there is this `"{attribute}"` thing in `ak.from_buffers` that sort of tells awkward this is supposed to be an offset?)

Probably i also should zip the global indices and the local indices together, then i wouldn't have that problem anymore ...

nikoladze:
> @nikoladze if we can wrap this up today or tomorrow we can have it in the next coffea release - which would be pretty snazzy.

Fine for me. Of course this is still work in progress, but so is DAOD_PHYSLITE, so i think it won't hurt to have some basic functionality already included in coffea.

I tried to get the `PHYSLITESchema` included in the docs which seemed to work more or less. I just didn't manage yet for `nanovents.methods.physlite` - when i add it to `docs/source/reference.rst` and try to compile the documentation i get an error

```
Extension error (sphinx.ext.linkcode):
Handler <function doctree_read at 0x7fdd9b47e4c0> for event 'doctree-read' threw an exception (exception: could not find class definition)
```

So here is a TODO list for the future:

* Complete the lists of possible cross-reference mappings
* Complete the mixin classes, add all important methods
* Consider making Records of Unions instead of Unions of Records for the cases with links into multiple collections
* Write more tests

... and i'm sure more things will come up.

Today i won't be able to get more work on this done, so if you think there is still significant things to do until this can be merged it is of course also fine to just have it  for the next release then. Thanks a lot!

lgray:
> So here is a TODO list for the future:
>
> Complete the lists of possible cross-reference mappings
> Complete the mixin classes, add all important methods
> Consider making Records of Unions instead of Unions of Records for the cases with links into multiple collections
> Write more tests

Please make a tracking issue for these and use checkboxes or whatever you like to mark completion.

Unless @nsmith- has something further that needs addressing I'd like to go ahead and merge this, we can figure out the docs stuff in time (or I'll fix it before minting a release).

coffea/nanoevents/schemas/physlite.py
nsmith-: > On the other hand there is this "{attribute}" thing in ak.from_buffers that sort of tells awkward this is supposed to be an offset?)

Right, for the higher level keys, ak.from_buffers is adding the `!offset` part automagically. So in principle, since both of these collections have the same offsets, it should be ok (desirable, even) that they have the same offsets key.

nikoladze: maybe it's a bug - i opened scikit-hep/awkward-1.0#899

PR #528, opened 2021-05-13
Let item transformation get the item from stack

nikoladze:
I noticed the special treatment for the `!item` transformation was a bit going against the concept of that stack based language. So I modified it to get the item from the stack instead of the special treatment with the second function argument.


nsmith- (approved):
Thanks for this, I agree it is a lot more consistent.

PR #537, opened 2021-06-02
Gzip fixes

### LINKED ISSUE #534: Opening text files as .gz files
**Describe the bug**
This [if statement](https://github.com/CoffeaTeam/coffea/blob/2baae94028c38b59f0eb52127d8fb92840dbf23d/coffea/lookup_tools/txt_converters.py#L22-L30) in `txt_converters.py` evaluates to true when there is a `.gz` anywhere in the file path, even if the file is not a `.gz` file.

**To Reproduce**
To reproduce this behavior, one would need a path that includes `.gz`. For example, we encountered this issue when running with the work-queue executor, where the paths to the files can include e.g. `python-env.tar.gz-env/`. The error we see is `gzip.BadGzipFile: Not a gzipped file (b'{1')`, where `{1` are the first two charters of the JEC txt file that the code is trying to read. 

**Expected behavior**
We expected the code not to attempt to open the file (in our case a `.txt` file) as a `.gz` file.

kmohrman:
These updates would resolve [Issue #534](https://github.com/CoffeaTeam/coffea/issues/534) by checking whether or not a file is a gz file before attempting to open it as a gz file. The recent [PR](https://github.com/CoffeaTeam/coffea/pull/536/commits/82099e92d70b3e9fb188c7b79e8d802893056033) by @lgray already resolves the issue with an adjustment of the filename-based approach, but I wanted to provide this PR as well because it checks the file content instead of relying on the file name.

### DIFF
--- a/coffea/lookup_tools/txt_converters.py
+++ b/coffea/lookup_tools/txt_converters.py
@@ -12,9 +12,11 @@ except ImportError:
 # for later
 # func = numbaize(formula,['p%i'%i for i in range(nParms)]+[varnames[i] for i in range(nEvalVars)])
 
+
 def is_gz_file(filename):
-    with open(filename, 'rb') as f:
-        return f.read(2) == b'\x1f\x8b'
+    with open(filename, "rb") as f:
+        return f.read(2) == b"\x1f\x8b"
+
 
 def _parse_jme_formatted_file(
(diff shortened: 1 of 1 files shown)


nsmith- (approved):
I like this approach of checking the file magic bytes.

lgray:
Yep - much better than checking the file names, also makes it easy to add other compression formats later.

PR #541, opened 2021-06-07
Use eventindex instead of global index form for PHYSLITE schema

nikoladze:
This might be a way to avoid hardcoding all possible relations between collections that can link into each other: Instead of a global index i attach an event index to all collections (also via a lazy form). This eventindex can then be used to index into the offset array of an arbitrary target collection (even after the object has been sliced) which then can be used to create the global indices.

Then i only have to specify the targets for links into fixed collections in the corresponding mixin methods and for the links into arbitrary/multiple collections i can do it dynamically.

I think i would prefer that approach. What do @nsmith- , @lgray  think? Am i missing something obvious?

### DIFF
--- a/coffea/nanoevents/methods/physlite.py
+++ b/coffea/nanoevents/methods/physlite.py
@@ -58,6 +58,18 @@ def _element_link_multiple(events, obj, link_field, with_name=None):
     return out
 
 
+def _get_target_offsets(offsets, event_index):
+
+    if isinstance(event_index, int):
+        return offsets[event_index]
+
+    def descend(layout, depth):
+        if layout.purelist_depth == 1:
+            return lambda: awkward.layout.NumpyArray(offsets)[layout]
+
+    return awkward._util.recursively_apply(event_index.layout, descend)
+
+
 @awkward.mixin_class(behavior)
 class Particle(vector.PtEtaPhiMLorentzVector, base.NanoCollection):
     """Generic particle collection that has Lorentz vector properties"""
@@ -133,13 +145,21 @@ class Electron(Particle):
     <https://gitlab.cern.ch/atlas/athena/-/blob/21.2/Event/xAOD/xAODEgamma/Root/Electron_v1.cxx>`_.
     """
 
+    # @property
+    # def trackParticles(self):
+    #     return _element_link(
+    #         self._events().GSFTrackParticles,
+    #         self.trackParticleLinks__G__GSFTrackParticles,
+    #         self.trackParticleLinks.m_persKey,
+    #     )
+
     @property
     def trackParticles(self):
-        return _element_link(
-            self._events().GSFTrackParticles,
-            self.trackParticleLinks__G__GSFTrackParticles,
-            self.trackParticleLinks.m_persKey,
-        )
+        target = self._events().GSFTrackParticles
+        awkward.materialized(target.pt)  # need to load one column
+        target_offsets = _get_target_offsets(target.layout.offsets, self._eventindex)
+        global_index = target_offsets + self.trackParticleLinks.m_persIndex
+        return target._apply_global_index(global_index)
 
     @property
     def trackParticle(self):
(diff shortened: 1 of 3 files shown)


nsmith-:
Restating what you said for my own understanding: the idea here is to save the `parents` jagged representation inside the inner RecordArray of the ListArray for each highlevel collection, since unlike `counts` or `offsets`, `parents` can be transparently sliced alongside the other fields and still retains the ability to identify the record's position along the first axis in the original unsliced array. Then, if the collection references any other collections, one can reconstruct the global index via `target_offsets[source_parents] +  local_index`.

This is a nice idea, and I think also applies equally well to NanoAOD. Taking it one step further, it would be nice to defer the 
"indexed take from jagged array" operation to awkward, probably using something like what is discussed in https://github.com/scikit-hep/awkward-1.0/issues/492 but I wasn't able to figure out immediately what the appropriate syntax is for our case.

nikoladze:
> Restating what you said for my own understanding: the idea here is to save the `parents` jagged representation inside the inner RecordArray of the ListArray for each highlevel collection, since unlike `counts` or `offsets`, `parents` can be transparently sliced alongside the other fields and still retains the ability to identify the record's position along the first axis in the original unsliced array. Then, if the collection references any other collections, one can reconstruct the global index via `target_offsets[source_parents] + local_index`.

Yes, in case by `parents` you mean the indices of the parent node in the layout? So to give an example for a source collection with a local index into a target collection

```python
import awkward as ak
import numpy as np
source = ak.zip({"local_index" : [[0, 1], [0, 2], [], [1, 2, 3]]})
target = ak.zip({"x" : [[0, 10, 20], [0, 20, 30, 40], [10, 20], [10, 20, 30, 40]]})
```

i would an the eventindex as field (what you call `parent`?)

```pycon
>>> source["_eventindex"] = np.arange(len(source))
>>> source.local_index
<Array [[0, 1], [0, 2], [], [1, 2, 3]] type='4 * var * int64'>
>>> source._eventindex
<Array [[0, 0], [1, 1], [], [3, 3, 3]] type='4 * var * int64'>
```

from which i get the global index exactly as you say, by using the eventindex to index into the offsets of the target and then add the local index

```pycon
>>> global_index = np.asarray(target.layout.offsets)[np.asarray(ak.flatten(source._eventindex))] + ak.flatten(source.local_index)
>>> <Array [0, 1, 3, 5, 10, 11, 12] type='7 * int64'>
```

which is what we want, just need to do all the `_recursively_apply` stuff to get it in arbitrary shapes. For this special case one could also explicitly construct it:

```python
layout = ak.layout.ListOffsetArray64(
    source.layout.offsets,
    ak.layout.IndexedArray64(
        ak.layout.Index64(global_index),
        target.layout.content
    )
)
```

```pycon
>>> ak.Array(layout).x
<Array [[0, 10], [0, 30], [], [20, 30, 40]] type='4 * var * int64'>
```

Which we can cross check with the local index since we didn't slice here
```pycon
>>> target[source.local_index].x
<Array [[0, 10], [0, 30], [], [20, 30, 40]] type='4 * var * int64'>
```

> 
> This is a nice idea, and I think also applies equally well to NanoAOD. Taking it one step further, it would be nice to defer the
> "indexed take from jagged array" operation to awkward, probably using something like what is discussed in [scikit-hep/awkward-1.0#492](https://github.com/scikit-hep/awkward-1.0/issues/492) but I wasn't able to figure out immediately what the appropriate syntax is for our case.

So the resolution mentioned in that issue applies for applying local indices, but not global indices right? Not sure how to make an appropriate function for this. Maybe it will be the public interface to `_recursively_apply` you requested in scikit-hep/awkward-1.0#516 eventually?

lgray:
@nikoladze is this one ready to go?

nikoladze:
Yes, in case you don't have any more suggestions, please merge.

nsmith- (approved):
Indeed what you call `eventindex` here has in awkward0 been referred to as `parents`. Awkward1 doesn't seem to have any public parents function, probably because it only makes sense in the context of a jagged array node specifically.

PR #550, opened 2021-07-06
Collapse 1d and Nd lookup cases

### LINKED ISSUE #549: lookup_tools issue with 1D lookup
**Describe the bug**

When using lookup_tools for 1D histograms I get the following error:
`ValueError: object too deep for desired array`

**To Reproduce**

Take your favorite root file with a 1D histogram (called "histo_eff_data" here), and run the following snippet:
```

ext = extractor()
ext.add_weight_sets(["ele_pt histo_eff_data your_root_file.root"])
ext.finalize()
evaluator = ext.make_evaluator()

evaluator["ele_pt"](
   ak.Array([[45]]),
)
```

**Expected behavior**
Return the proper lookup value, similar to 2D lookups.

**Additional context**
The issue seems to come from this special case here: https://github.com/CoffeaTeam/coffea/blob/master/coffea/lookup_tools/dense_lookup.py#L34
Doing the same thing for 1D as for 2D fixes the issue.

lgray:
Fixes #549


lgray:
@danbarto If this fix works for you, can you provide a small test that fails for your issue and passes for this fix?

danbarto:
@lgray thanks, the fix works. This snippet works now (and failed before):

```
import awkward as ak

ext = extractor()
f_in = "https://github.com/danbarto/tW_scattering/raw/master/data/leptons/ttH/error/SFttbar_2016_ele_pt.root"
ext.add_weight_sets(["ele_pt histo_eff_data %s"%f_in])
ext.finalize()
evaluator = ext.make_evaluator()

evaluator["ele_pt"](
   ak.Array([[45]]),
)
```

lgray:
OK this fix breaks some other things - I'll see what's going on and fix up the rest. Will get back to you.

kondratyevd:
@lgray I've had issues with 1D lookups before, but I don't remember if it was exactly the same error. I was able to overcome it by doing this before evaluation (`lookup` is an instance of `dense_lookup`):
```python
lookup._axes = lookup._axes[0]
```
Maybe that helps.

lgray:
@kondratyevd thanks! I'll try this out later today. @danbarto can you help test when I'm back at a laptop (6-7 hours)?

danbarto:
Sure, happy to help. Time permitting I'll also check if I can use @kondratyevd's suggestion.

lgray:
@danbarto - thanks for the MRE I've added it to the coffea tests.
@kondratyevd 

This should handle both cases well. Please give it a try.

lgray:
@danbarto all good for you?

danbarto:
Thanks @lgray, works for me!

PR #551, opened 2021-07-07
Fix JER smearing hybrid method deltaPtRel threshold

bfis:
According to the [twiki](https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution), `deltaPtRel` should be smaller than 3 times the resolution, for the hybrid JER smearing method to be applicable.
This is corroborated by the [implementation of the `SmearedJetProducerT`](https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h), see [L76](https://github.com/cms-sw/cmssw/blob/a59283f45c8b5becca6cff5e643687c6241343a3/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h#L76) and [L243](https://github.com/cms-sw/cmssw/blob/a59283f45c8b5becca6cff5e643687c6241343a3/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h#L243) and [typical values for `dPtMaxFactor`](https://github.com/cms-sw/cmssw/search?q=dPtMaxFactor).

### DIFF
--- a/coffea/jetmet_tools/CorrectedJetsFactory.py
+++ b/coffea/jetmet_tools/CorrectedJetsFactory.py
@@ -71,7 +71,7 @@ def jer_smear(
     jersmear = jet_energy_resolution * jet_resolution_rand_gauss
     jersf = jet_energy_resolution_scale_factor[:, variation]
     deltaPtRel = (jetPt - pt_gen) / jetPt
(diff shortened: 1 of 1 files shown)


lgray:
Oops! Thank you.

PR #554, opened 2021-07-15
Fix `root_converters.py` for `uproot.ReadOnlyDirectories`

pfackeldey:
Dear coffea developers,

`uproot.ReadOnlyDirectories` do not have the attribute `.classname`. Currently the code can not handle ROOT files containing directories and thus breaks.
An alternative fix to this PR would be to directly iterate over the _wanted_ types, such as:

```python
for path, item in f.iteritems(recursive=True, filter_classname=set(histTypes) | set(graphTypes)):
    ...
```
However this would exclude: https://github.com/CoffeaTeam/coffea/blob/master/coffea/lookup_tools/root_converters.py#L33-L37
What is your opinion here?

Best, Peter


lgray:
The mentioned line is related to TGraphErrors anyway, so perhaps it's best to go with the 

> for path, item in f.iteritems(recursive=True, filter_classname=set(histTypes) | set(graphTypes)):
>  ...

Implementation anyway?

@nsmith- ?

nsmith-:
Maybe we are just missing `TGraphErrors` in addition to `TGraphAsymmErrors`?

lgray:
I think so. I've never heard of anyone using them for corrections though (no complaints)

pfackeldey:
To me it looks like that the code runs anyway in a [`continue`](https://github.com/CoffeaTeam/coffea/blob/master/coffea/lookup_tools/root_converters.py#L32) statement for the `TGraph`-types.  
What is your opinion on the implementation? I could update the PR and exclusively iterate over the `TH`-types, since `TGraph`-types are skipped currently anyway?  
I am happy for any feedback and will update the PR asap :)

lgray:
@pfackeldey Sorry for the delay here. 

We'll add the `iteritems` implementation and iterate only over the histogram types in a later pull request.

For now, we'll use the skip as you have submitted. Please add a sample file to `tests/samples` that causes the original code to fail and add a test against that new file.
Add a warning using `FutureWarning` that TGraph reading for the root file reader will be removed in a future release, and if someone needs TGraph reading they should post an issue on the coffea github.

pfackeldey:
@lgray I've updated this PR according to your wishes.

Sorry for the linting commit-noise... I was too used to python3.8's walrus operator and forgot that coffea also supports python<3.8. Let me fix it and I think then it is ready to go :)

lgray:
Thanks for adding these items! Merging. 🚀

PR #557, opened 2021-07-27
[Executors] Allow better interoperability with custom executors

pfackeldey:
Dear coffea developers,

(as announced in Gitter chat...)
This PR enables a better interoperability with custom user-defined executors. We run into this while optimising the `dask_executor` for our local computing cluster.

### New Usage

The code comes with a few usage-changes (old usage is still kept through backward-compatibility):

```python
from coffea.processor.executor import run_uproot_job

out = run_uproot_job(..., executor="Iterative", ...)
```
The available options are: `"Iterative", "Futures", "Dask", "Parsl", "WorkQueue"`. 

### New Feature

One can write now _properly_ custom executors, which e.g. are perfectly configured for a computing cluster. This is done by using a simple _decorator_:

```python
from coffea.processor.executor import run_uproot_job, DispatchExecutor, ExecutorKind

@DispatchExecutor(name="MyDask", kind=ExecutorKind.Dask)
def my_dask_executor(items, function, accumulator, **kwargs):
    # do whatever you want here
    

out = run_uproot_job(..., executor="MyDask", ...)
```

`DispatchExecutor` arguments:  
* `name`: registers the executor function under a name, which can then be used in the `run_uproot_job` function
* `kind`: ensures that the `run_uproot_job` function treats the executor function in a certain way, e.g.: set `retries=0` for all executors which use `kind=ExecutorKind.Dask`

I have updated the tests accordingly.

Let me know what you think about this proposal!
Thank you very much in advance for reviewing,
Peter


pfackeldey:
@lgray some CI-checks seem to be stuck, do you know why this happens?

lgray:
@pfackeldey Sorry for the delay here, this PR has caused some (possibly unnecessarily) deep thinking. I'm not entirely comfortable with the interface choices (particularly executor choice being driven by strings). We should discuss more how this evolves and long term support, etc.... We should be careful to understand when a specialized executor is really needed or if things should be incorporated into executor in coffea.

Nothing against you personally, of course!

pfackeldey:
@lgray Thanks for the reply and no worries!

I wanted to raise some awareness that the current code is very much tailored to the existing executors and makes it hard to write a custom executor without drilling into multiple places. 
Of course this is not the usual use-case. Most people will directly use the executors you offer, which work perfectly out of the box. 
We (my small analysis group) need to exploit a bit more the hardware setup our local cluster to get the best performance, which unfortunately needs us to write a custom tailored `dask_executor`, which is very likely not suited for other computing clusters. 

I would be happy to share our experiences and discuss how this evolves with you! 

This PR is meant to be a _proposal_, feel free to close it :) 
(In case you have an alternative implementation let me know and I'll give it a try :) )

Best, Peter

lgray:
@pfackeldey this is much more pleasant, thanks for experimenting! A remaining issue with merging this is that the names of standard executors change (this will make many people grumpy). Either we save this for coffea `0.8.0` or bring back the old executor naming for the existing executors if we want it in `0.7.x`.

@nsmith- your thoughts on this?

pfackeldey:
@lgray It is a lot more robust now and avoids not-so-nice mechanisms as introduced before (such as registries). 
I am not yet finished though, my plan is to make the `run_uproot_job` function accept a preconfigured instance of a executor and then do `isinstance`-checks instead of `issubclass`-checks. Also I want to move the `executor_args` into the constructor. Currently the `executor_args` hold kwargs for the `_work_function` like `mmap` or `retries`, which is a bit mixed up I think. I plan to clearly seperate them too within this PR.
Furthermore I would like to make the executors a python `dataclass`, which provide directly a nice `__repr__`. This might be handy once the other PR is ready, which introduces logging to coffea.
I imagine something like this:

```python
from coffea.processor import FuturesExecutor

executor = FuturesExecutor(status=True, compression=1, ...)

out = processor.run_uproot_job(
    filelist,
    "Events",
    some_processor_instance,
    executor,
    work_function_args, # <- only passed to: closure = partial(_work_function, **work_function_args)
    chunksize=chunksize,
)
```

Additionally my colleague @bfis wanted to add a few more improvements to the `DaskExecutor`, such as a working `tqdm`-progressbar. This work will be probably done within the next days. 

Once these updates are done, I would mark this PR as ready. 

Concerning the API and the now capitalized executor "names". I could put an alias into the `__init__` to expose it under the "old" name I assume. However, the usage will likely change a bit anyways to create a properly configured instance... So I wonder if this is useful?

pfackeldey:
Side note: I am not touching anything `spark`-related, since I have no experience with this package. How should this be migrated? Do you have any preference?

lgray:
Spark is special - don't worry about it.

nsmith-:
Well, amusingly enough, the original executor pattern was going to be
```python
executor = partial(processor.futures_executor, workers=4, ...)
processor.run_uproot_job(..., executor)
```
i.e. the idea was to be very functional. Then I realized people maybe aren't as into functional programming as I was at the time, so I added `executor_args` to allow the closure to be specified later. It seems what this has done is turn it into a class object, which is probably a cleaner way to avoid being functional than adding the ad-hoc dictionary. So I like it, just still a bit sad we didn't go full functional :p

coffea/processor/__init__.py
```diff
@@ -38,11 +38,11 @@
...
-    "iterative_executor",
-    "futures_executor",
-    "dask_executor",
-    "parsl_executor",
-    "work_queue_executor",
+    "IterativeExecutor",
+    "FuturesExecutor",
+    "DaskExecutor",
+    "ParslExecutor",
+    "WorkQueueExecutor",
```
nsmith-, line 132: I would prefer to drop the capitalization for the executor types, because if we make a small change to how the executor is initialized, we can make it so that no user changes are required.

coffea/processor/executor.py
```diff
@@ -1087,6 +1133,13 @@ def run_uproot_job(
 
     if pre_executor is None:
         pre_executor = executor
+
```
nsmith-, line 901: Here we could do:
```python
if isinstance(executor, type) and issubclass(executor, ExecutorABC):
    executor = executor()  # initialize with default arguments
elif not isinstance(executor, ExecutorABC):
    raise ValueError("...")
```
I think then we would eventually want to deprecate the first option and mandate users add `()` at the end of their executors to initialize them. This could come along with dropping `executor_args`.

coffea/processor/executor.py
```diff
@@ -259,7 +256,19 @@ def _futures_handler(futures, timeout):
             _cancel(futures.pop())
 
 
-def work_queue_executor(items, function, accumulator, **kwargs):
+class ExecutorABC(metaclass=ABCMeta):
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs or {}
```
nsmith-: This is a good opportunity to make the executor validate its options, e.g. the derived classes would implement properties with setters to adjust configurations.

pfackeldey:
Hi @lgray and @nsmith- ,

sorry @nsmith- I just saw your reviews after pushing, would you mind reviewing again?

Finally I arrived at an implementation, which improves the points I listed in my previous comments:

- customisation is now possible through inheritance
- `kwargs` are now explicit
- `kwargs` for executors and others (e.g.: `_work_function`) are now clearly splitted
- `run_uproot_job` and `run_parquet_job` share the same implementation
- multiple code simplifications/reductions due to class-based implementation
- nice `__repr__` through dataclasses for e.g. logging
- type annotations for all kwargs

_Side note_: Executors are required to be pickable (which is a reasonable requirement in my opinion).

The updated interface looks, for example, like this:

```python
# 1. fully configure an executor
executor = processor.DaskExecutor(client=client, use_dataframes=True)

# 2. fully configure a Runner instance + give it the preconfigured executor
run = processor.Runner(executor=executor, schema=nanoevents.NanoAODSchema)

# 3. run it with: files, tree_name and processor_instance
output = run(filelist, "Events", processor_instance=NanoTestProcessorPandas())
```

You can find more examples in the unittests.

I think 2 things still need some work and maybe input/help from you:

1. updated doc-strings for all kwargs in all executors/Runner
2. a review concerning the WorkQueueExecutor (I think I updated it correctly, but still would like to have some feedback here)

Let me know what you think!

Best, Peter

nsmith-:
This latest change looks like a really promising direction, and in fact is in line with my thoughts on a general `processor` submodule refactoring, so I'm very happy to see it! One of the main technical debts we have with the processor module is a mess of configuration options that aren't well classified. I see three classes of configuration options:
- Options that adjust how a execution resource is acquired, e.g. futures_executor `workers` or work_queue_executor `memory` (note that sometimes the resource is only given as-is, e.g. dask_executor `client`)
- Options that adjust how the input and output of the Processor is handled, e.g. `schema` or `use_dataframes` (some of these could also be specified in static Processor class attributes)
- Options that adjust performance and resilience of the map-reduce, e.g. `chunksize`, `retries`, `xrootdtimeout`, `treereduction`, `compression`, ...

It sounds like each class can be associated with one of `Executor`, `Processor`, or `Runner`, or some variation of this. One more ingredient that I wanted to implement is a better representation of the input, e.g. an InputSpec that would abstract `fileset` and `treename`, possibly allowing reading parquet or other input types (since NanoEvents supports this already)
I'm not specifically requesting that you implement all this, of course, but would be happy to hear input and ideas.

Regarding your comment:
> Executors are required to be pickable

I am not sure how this works when a futures_executor has been given a `pool` or a dask_executor is given a `client`, both of which are references to run-time resources that are not really serializable. Perhaps we need to separate the Executor into a serializable component (e.g. ExecutorConfig) and a resource handle component.

lgray:
OK - nice and we can make stand-in objects for the old-style executors and run_uproot_job using partial or similar tricks. Not bad.

pfackeldey:
Thanks for your replies,

I will try to comment in more detail tomorrow/in the next days.

Concerning the pickability...
Currently I do for the DaskExecutor to make it pickable (only the client is un-pickable):

```python
def __getstate__(self):
     return dict(self.__dict__, client=None)
```
This lives on the DaskExecutor directly. So my policy was that each Executor makes this sure by itself.

For the FuturesExecutor I did not yet have to do it, maybe I missed something?

(Update: I think I need to do the same for the FuturesExecutor if a pool _instance_ is given, will try & update)

pfackeldey:
@nsmith- I will try to comment inline,

> This latest change looks like a really promising direction, and in fact is in line with my thoughts on a general `processor` submodule refactoring, so I'm very happy to see it! One of the main technical debts we have with the processor module is a mess of configuration options that aren't well classified. I see three classes of configuration options:
> 
> * Options that adjust how a execution resource is acquired, e.g. futures_executor `workers` or work_queue_executor `memory` (note that sometimes the resource is only given as-is, e.g. dask_executor `client`)
> * Options that adjust how the input and output of the Processor is handled, e.g. `schema` or `use_dataframes` (some of these could also be specified in static Processor class attributes)
> * Options that adjust performance and resilience of the map-reduce, e.g. `chunksize`, `retries`, `xrootdtimeout`, `treereduction`, `compression`, ...
> 
> It sounds like each class can be associated with one of `Executor`, `Processor`, or `Runner`, or some variation of this. 

I fully agree with you. This sounds like a good plan, and actually what I was trying to do within this PR. For most options it is pretty clear where they belong. Some might require some thinking, e.g.:

- `use_dataframes`: Determines the input/output and thus makes sense to be defined on the `Processor`. However it can only be utilised by the DaskExecutor and uses Dask-specific data frames.
- `retries`: I think this and its mechanism makes more sense to be defined individually on each `Executor` instead of the `Runner` (as it is right now).
- `dynamic_chunksize`: Seems to be only supported by/for the `WorkQueueExecutor`. However I think this should be a general option for the `Runner`. 

There are also other options, which I am not really familiar with, mostly `WorkQueueExecutor`-specifics. So I tried to retain the original state as much as possible. The tests seem to pass 🎉 

> One more ingredient that I wanted to implement is a better representation of the input, e.g. an InputSpec that would abstract `fileset` and `treename`, possibly allowing reading parquet or other input types (since NanoEvents supports this already)

This sounds like a nice idea! Do you intend to have it as a simple data-container or with more sophisticated methods, like "parquet" and "root" specific file handling ? Options like `chunksize` or `dynamic_chunksize` may even better fit to this `InputSpec` than to the `Runner`.

> I'm not specifically requesting that you implement all this, of course, but would be happy to hear input and ideas.

As you like, I can try to add some more of your ideas. I think it makes sense to discuss at which point we would consider this PR as ready and maybe add other things later. What do you think, @lgray @nsmith- ?

> 
> Regarding your comment:
> 
> > Executors are required to be pickable
> 
> I am not sure how this works when a futures_executor has been given a `pool` or a dask_executor is given a `client`, both of which are references to run-time resources that are not really serializable. Perhaps we need to separate the Executor into a serializable component (e.g. ExecutorConfig) and a resource handle component.

nsmith-:
I'd like to invite @btovar to review the workqueue changes, and he may also have input on the overall strategy. In https://github.com/CoffeaTeam/coffea/pull/544 he indicated interest in having adaptive chunksizes also for other executors, so we should at least consider what needs to happen for that during this refactor.

pfackeldey:
Concerning the WorkQueue changes: The only difference I can see is that the defaults of the options are now directly in the definition of the Executor. So I think these defaults here: https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/work_queue_tools.py#L414-L455 can be removed.

coffea/processor/executor.py
```diff
@@ -259,7 +245,33 @@ def _futures_handler(futures, timeout):
...
+    # shared by all executors
+    status: bool = True
+    unit: str = "items"
+    desc: str = "Processing"
+    compression: Optional[int] = 1
+    function_name: Optional[str] = None
+
+    def __call__(
+        self,
+        items: Sequence,
```
btovar: `items`  could also be a generator or iterator, at least for WorkQueueExecutor. However, I'm not sure if that should be specified here, or just for Work Queue.  I think that that Sequence does not cover generators, as they lack __len__.

coffea/processor/executor.py
```diff
@@ -340,18 +349,69 @@ def work_queue_executor(items, function, accumulator, **kwargs):
...
+    wrapper: Optional[str] = shutil.which("python_package_run")
+    resource_monitor: bool = False
+    resources_mode: str = "fixed"
+    cores: Optional[int] = None
+    memory: Optional[int] = None
+    disk: Optional[int] = None
+    gpus: Optional[int] = None
+    chunks_per_accum: int = 10
+    chunks_accum_in_mem: int = 2
+    chunksize: int = 1024
```
btovar, line 395: I think we envisioned chunksize, dynamic_chunksize and dynamic_chunksize_ as parameters to ExecutorBase. (E.g., currently they are proper arguments to run_uproot_job.)

coffea/processor/executor.py
```diff
@@ -340,18 +349,69 @@ def work_queue_executor(items, function, accumulator, **kwargs):
...
+    dynamic_chunksize: bool = False
+    dynamic_chunksize_targets: Dict = field(default_factory=dict)
+
+    def __call__(
+        self,
+        items: Sequence,
```
btovar: (As mentioned above, this could be a generator or an iterator.)

btovar (changes requested):
@pfackeldey on first sight this looks very good, just a couple of minor comments. I have not tested it, but should be able to do that later today.

It is nice that the arguments to the runner, executor, and work_function are now more clearly divided.

Also, I wonder if we should rename work_queue_tools.py to something like work_queue_executor.py or some other standard name, as I imagine that some other executors may break away from executor.py.

coffea/processor/executor.py
pfackeldey: yep, it should be the `Iterable` protocol, which implements `__iter__`.

coffea/processor/executor.py
pfackeldey: This is where I am not sure about. So in my understanding they should be part of the `Runner` class. This one just generates pre-chunked `WorkItems` which any executor just digests, no matter their chunking. This is possible for all executor, except maybe the `WorkQueue` one? What do you think?

coffea/processor/executor.py
pfackeldey: will update :)

btovar:
> Concerning the WorkQueue changes: The only difference I can see is that the defaults of the options are now directly in the definition of the Executor. So I think these defaults here: https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/work_queue_tools.py#L414-L455 can be removed.

Yes, it is nice that now doc and defaults will be in the same place.

Another change is that the functions work_queue_tools._warn_unknown_kwords() and  _update_deprecated_kwords() are not needed anymore.

btovar:
The only thing I'm not a super fan of is the rename from xy_executor to XyExecutor, as that involves changing current user code.

But since run_uproot_job is changing to Runner, and we need to add the parenthesis to XyExecutor(...) anyway, the user code will need to change anyway...

coffea/processor/executor.py
```diff
@@ -1027,33 +800,13 @@ def run_uproot_job(
             treename can also be defined in fileset, which will override the passed treename
         processor_instance : ProcessorABC
             An instance of a class deriving from ProcessorABC
-        executor : callable
-            A function that takes 3 arguments: items, function, accumulator
+        executor : ExecutorBase instance
```
btovar, line 808: fileset, treename, and processor_instance should be removed from the doc.

btovar (approved):

btovar (approved):
Just a small change, as Runner arguments now "start" from executor.

I was able to run our main application topcoffea with the new changes, including with dynamic_chunksize.

coffea/processor/executor.py
btovar: Yes, this one is tricky, as both Runner and executor need to know dynamic_chunksize. 

One possible solution is to allow `items` to be a callback that Runner defines when it gets dynamic_chunksize=True. Then the executor calls the function with the current task reports as arguments when it needs the next chunk.

Probably it will be better to solve it after the current pr is merged. In the meanwhile, we just need to specify dynamic_chunksize as an arg to Runner and the executor.

coffea/processor/executor.py
pfackeldey: Good catch! I move it to the `__call__`, where it belongs to

coffea/processor/executor.py
pfackeldey: Sounds like a good plan. `dynamic_chunksize` is an arg to both, so this should be the current state.

btovar (approved):

coffea/processor/executor.py
btovar: As I thought about it last night, we don't even need the callback, as we already have the generator's .send method, so it will be just a matter to move some code to the chunk generator.

lgray:
@pfackeldey Given that @btovar is happy with the changes. I think a remaining pain point is to create an emulation of the old interface using the new one and to set deprecation markers.

pfackeldey:
@lgray I've reconstructed the old usage in the last commit with the new implementation. 
When using these old methods (`run_uproot_job` and `run_parquet_job`) the following deprecation warning is added:

```bash
/.../coffea/coffea/util.py:97: FutureWarning: In coffea version <unknown>, this will be an error.
(Set coffea.deprecations_as_errors = True to get a stack trace now.)
str: This method is deprecated, please use directly the new: <class 'coffea.processor.executor.DaskExecutor'> and <class 'coffea.processor.executor.Runner'> classes.

    Please use instead, e.g.:

        executor = IterativeExecutor()
        run = processor.Runner(
            executor=executor,
            schema=processor.NanoAODSchema,
        )
        hists = run(filelist, "Events", processor_instance=processor_instance)

  warnings.warn(message, FutureWarning)
[##########                              ] | 25% Completed |  4.4s
```
I think I could auto-generate the correct code one needs to implement depending on the used `executor` and `executor_args`, instead of this simple static one with `IterativeExecutor `. However I think this might be too much/over-engineered and this small snippet already gives the idea on the new usage.
Is that alright with you?

(I can also substitute the version where it will be gone if you let me know, currently I used `<unknown>` as a place holder.)

coffea/processor/__init__.py
```diff
@@ -33,18 +32,103 @@
...
+
+        executor = IterativeExecutor()
+        run = processor.Runner(
+            executor=executor,
+            schema=processor.NanoAODSchema,
+        )
+        hists = run(filelist, "Events", processor_instance=processor_instance)
+    """
+
+    deprecate(
```
nsmith-: I think we should hold off on the deprecation warning until v0.8, and say it will go away in v0.9, @lgray?
Meanwhile we can advertise the new interface and get some time to make sure it is robust.

coffea/processor/executor.py
```diff
@@ -11,22 +11,26 @@
...
-from .dataframe import (
-    LazyDataFrame,
-)
+from .accumulator import accumulate, set_accumulator, Accumulatable
+from .dataframe import LazyDataFrame
 from ..nanoevents import NanoEventsFactory, schemas
 from ..util import _hash
 
 from collections.abc import Mapping, MutableMapping
+from dataclasses import dataclass, field, asdict
```
nsmith-, line 26: `dataclasses` is py3.7+, but I guess the py3.6 tests must be passing because some existing dependency requires the py3.6 backport module. To be safe, can you add `dataclasses;python_version<"3.7"` to https://github.com/CoffeaTeam/coffea/blob/2197e71dacf329910373369dfbbb4895ef0977e1/setup.py#L58

coffea/processor/executor.py
nsmith-: Yeah, I like this plan: the `items` argument to the `__call__` method becomes type `Generator[WorkItem, RunMetrics, None]` where `RunMetrics` is whatever metrics is needed to be sent back to the chunk generator to optimize the size.
But that can be for a later PR

coffea/processor/executor.py
```diff
@@ -410,45 +481,53 @@ def futures_executor(items, function, accumulator, **kwargs):
...
-                tqdm(
-                    gen if clevel is None else map(_decompress, gen),
-                    disable=not status,
-                    unit=unit,
-                    total=len(items),
-                    desc=desc,
-                ),
-                accumulator,
+
+    pool: concurrent.futures.Executor = concurrent.futures.ProcessPoolExecutor
```
nsmith-: `pool: Union[Callable[..., concurrent.futures.Executor], concurrent.futures.Executor] = `
Since one can also pass a executor class.

coffea/processor/executor.py
```diff
@@ -483,96 +562,126 @@ def dask_executor(items, function, accumulator, **kwargs):
...
-    priority = kwargs.pop("priority", 0)
-    retries = kwargs.pop("retries", 3)
-    heavy_input = kwargs.pop("heavy_input", None)
-    function_name = kwargs.pop("function_name", None)
-    use_dataframes = kwargs.pop("use_dataframes", False)
+    client: Optional["dask.distributed.Client"] = None  # noqa
+    treereduction: int = 20
+    priority: int = 0
+    retries: int = 3
+    heavy_input: Optional[str] = None
```
nsmith-: Shouldn't this be `Optional[bytes]`? Should be a pickled item

coffea/processor/executor.py
```diff
@@ -483,96 +562,126 @@ def dask_executor(items, function, accumulator, **kwargs):
...
-    client = kwargs.pop("client")
-    ntree = kwargs.pop("treereduction", 20)
-    status = kwargs.pop("status", True)
-    clevel = kwargs.pop("compression", 1)
-    priority = kwargs.pop("priority", 0)
-    retries = kwargs.pop("retries", 3)
-    heavy_input = kwargs.pop("heavy_input", None)
-    function_name = kwargs.pop("function_name", None)
-    use_dataframes = kwargs.pop("use_dataframes", False)
+    client: Optional["dask.distributed.Client"] = None  # noqa
```
nsmith-, line 580: As it stands I don't think `client` is optional? Or is this more to do with the desire to have Executors be picklable? Maybe we have to return to the idea of separating the resource from the Executor.

coffea/processor/executor.py
```diff
@@ -483,96 +562,126 @@ def dask_executor(items, function, accumulator, **kwargs):
...
+        accumulator: Accumulatable,
+    ):
+        if len(items) == 0:
+            return accumulator
+
+        import dask.dataframe as dd
+        from dask.distributed import Client
+
+        if self.client is None:
+            self.client = Client()
```
nsmith-: A word of caution, the default dask client constructs a `LocalCluster` with typically more threads per core than we are able to use effectively due to awkward holding the GIL (this is something that should improve eventually) I would suggest to set `Client(threads_per_worker=1)` for now.

coffea/processor/executor.py
```diff
@@ -1073,157 +817,450 @@ def run_uproot_job(
...
+    maxchunks: Optional[int] = None
+    metadata_cache: Optional[MutableMapping] = None
+    dynamic_chunksize: bool = False
+    dynamic_chunksize_targets: Dict = field(default_factory=dict)
+    skipbadfiles: bool = False
+    xrootdtimeout: Optional[int] = None
+    align_clusters: bool = False
+    savemetrics: bool = False
+    mmap: bool = False
+    schema: schemas.BaseSchema = schemas.BaseSchema
```
nsmith-: `Optional[schemas.BaseSchema]` (at least until we deprecate LazyDataFrame)

coffea/processor/executor.py
```diff
@@ -1073,157 +817,450 @@ def run_uproot_job(
...
+    cachestrategy: Optional[str] = None
```
nsmith-: `Optional[Union[Literal["dask-worker"], Callable[..., MutableMapping]]]`

coffea/processor/executor.py
```diff
@@ -1073,157 +817,450 @@ def run_uproot_job(
...
+
+    def automatic_retries(self, func, *args, **kwargs):
+        """This should probably defined on Executor-level."""
+        import warnings
+
+        retry_count = 0
+        while retry_count <= self.retries:
+            try:
+                return func(*args, **kwargs)
+                break
```
nsmith-: superfluous break?

coffea/processor/executor.py
```diff
@@ -1073,157 +817,450 @@ def run_uproot_job(
...
+                    not self.skipbadfiles
+                    or "Auth failed" in str(e)
+                    or self.retries == retry_count
+                ):
+                    raise e
+                warnings.warn("Attempt %d of %d." % (retry_count + 1, self.retries + 1))
+            retry_count += 1
+
+    @staticmethod
+    def _normalize_fileset(fileset: Dict, treename: str) -> Generator:
```
nsmith-: `-> Generator[FileMeta, None, None]`

coffea/processor/executor.py
```diff
@@ -1073,157 +817,450 @@ def run_uproot_job(
...
+            else:
+                raise ValueError(
+                    "Expected schema to derive from nanoevents.BaseSchema, instead got %r"
+                    % self.schema
+                )
+            tic = time.time()
+            try:
+                out = processor_instance.process(events)
+            except Exception as e:
+                file_trace = f"\n\nFailed processing file: {item.filename} ({item.entrystart}-{item.entrystop})"
```
nsmith-: Now that `WorkItem` is a fancy dataclass, I think its repr can be used here

nsmith- (changes requested):
Overall it looks great! I mostly have some typing adjustments.
I'm curious, how does `mypy` look on this file now? Probably we are a long way from full typing but introducing the dataclasses is a big step in that direction.
One thing that is abundantly clear now is we need to convert the preprocessing and executing to a more stream-based interface, using a chain of generators to control the flow.

pfackeldey:
> Overall it looks great! I mostly have some typing adjustments.
> I'm curious, how does `mypy` look on this file now? Probably we are a long way from full typing but introducing the dataclasses is a big step in that direction.
> One thing that is abundantly clear now is we need to convert the preprocessing and executing to a more stream-based interface, using a chain of generators to control the flow.

Thanks for your feedback!
I will implement your wishes soon, I will be from _exactly now on_ on a 2 week vacation :D So I'll come back to this PR afterwards, hope this is fine with you :)

nsmith-:
Aha! Enjoy your vacation! I might make the changes myself in the meantime since they are small, if you are ok with that?

pfackeldey:
Sure, please go ahead :)

coffea/processor/__init__.py
lgray: Yeah I like this slower approach. 0.9 might be a while though.
I'll think about it over the weekend.

nsmith-: 0.9 can be as soon as we like! Anyway no harm in keeping deprecated interfaces around for a long time, as long as they are just shims like this one.

lgray:
@pfackeldey when you get back this will need a rebase!

coffea/processor/__init__.py
pfackeldey: I left it as a comment and updated the deprecated version to 0.9

coffea/processor/executor.py
pfackeldey: good catch, done

coffea/processor/executor.py
pfackeldey: done

coffea/processor/executor.py
pfackeldey: done

coffea/processor/executor.py
pfackeldey: Yes, I wanted to have a default client here, but without importing dask in the global scope. So I used this "None" as a workaround. However I can also make it a non-default argument if you wish.

coffea/processor/executor.py
pfackeldey: good point, done

coffea/processor/executor.py
pfackeldey: done

coffea/processor/executor.py
pfackeldey: done

coffea/processor/executor.py
pfackeldey: yep, done

coffea/processor/executor.py
pfackeldey: done

coffea/processor/executor.py
pfackeldey: done

pfackeldey:
Hi @lgray and @nsmith- ,
finally I am able to come back to this PR. I've merged the latest changes to the master and updated the PR regarding your wishes @nsmith- .
Best, Peter

coffea/processor/executor.py
pfackeldey: The pickling is no problem, since it is taken care of through: `__getstate__`.

pfackeldey:
Can you re-trigger the one failed test? I don't think it has anything to do with my code changes. (I can not find the button to rerun this test manually...)

nsmith- (approved):
Thanks!

PR #564, opened 2021-08-03
Make WorkItem a namedtuple

jrueb:
This allows proper comparison between instances while keeping them hashable.

### Issue
While implementing a custom executor, I noticed the `WorkItem` is really not flexible. While the `FileMeta` class has `__eq__` implemented, `WorkItem` doesn't. Not having `__eq__` turned out to be a big drawback because use cases like `item in somelist` are not possible.

### Solution
Inherit from `namedtuple`. This should not affect the current use of `WorkItem` in any way. However it adds several nice advantages, like `__eq__`, `__hash__` and `__str__`.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -12,7 +12,7 @@ import uproot
 import uuid
 import warnings
 from tqdm.auto import tqdm
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from cachetools import LRUCache
 import lz4.frame as lz4f
 from .processor import ProcessorABC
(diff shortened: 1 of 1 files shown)


jrueb:
Unfortunately I missed this test and in there the mutability of the previous `WorkItem` was made use of, which I thought to be irrelevant. Because of this, my assumption of this issue being just one simple change isn't valid anymore.

I still think an implementation of `__eq__` would be very useful here.
In case you agree with me I see two possible methods of solving this:

1. Make it unhashable, which would make it impossible to be used in sets or dict keys
2. Make it immutable (and create new instances if one wants to modify an attribute)

What from I see right now, immutability is possible with little change.

lgray:
Immutable is better for sure - please dig into why it's breaking the rados tests.

lgray:
@jrueb Let me know if you want to add anything more or if this is ready to go.

jrueb:
I do not have any other changes. For me, this is good to go.

PR #565, opened 2021-08-04
Fixed jersf_lookup for 0 jets 

bfis:
When passing arguments to `jersf_lookup` (both `__call__` and `_evaluate`) that have an effective shape of `(N...)`, where `N...` has an `0` in its shape i.e. `numpy` shape `(0,)`, the function returns an effective shape of `(N..., 3, 0)` instead of the expected `(N..., 3)`. This is caused by an [difference in behavior of fancy indexing on awkward arrays vs. numpy arrays](https://github.com/scikit-hep/awkward-1.0/discussions/1047). The proposed change uses the assumption that the inner most dimension of the arrays in `self._parms` is consistent, since it ought to be because it is actually not fed into any function to be evaluated but rather a constant.


lgray:
Excellent, thank you for this fix!

PR #566, opened 2021-08-06
Logging system

maxgalli:
After discussion with @lgray, here is a WIP PR to add a logging system in coffea.
Ideally, this would consist in two parts:
- [ ] in every module/submodule where we want information to be printed, we need to add 
```
import logging
logger = logging.getLogger(__name__)
```
and ```log.info(stuff)```, ```log.debug(stuff)``` depending on the level of information we want to print
- [x] provide a sort of coffea logger that can be imported with something like
```
from coffea.logger import get_logger
```
with customizable log level and output (description is provided in the function documentation)

Comments/questions:

- the logger uses ```rich``` if the Python version is >= 3.6.1, as ```rich``` does not support 3.6.0
- while trying this out with [hgg-coffea](https://github.com/maxgalli/hgg-coffea), I will try to identify spots in which dumping information could be useful; any hint in this sense is much appreciated
- better to implement a couple of tests for the ```get_logger``` function? 
- should ```rich``` be treated as a dependency?


andrzejnovak:
@henryiii had opinions on making rich a dependency in context of another project i think, but don't remember exactly what the problem was

lgray:
Yes I would value @henryiii's opinion here. 

However I think the discussion may be moot?

@maxgalli if you do:
```python3
try:
    import rich
except ImportError:
    rich = None
```

You can very easily make rich completely optional (not even a dependency!) while keeping the logging as you would like it in either case, regardless of python version.

Then all the logic later in `get_logger` becomes predicated on rich existing rather than the use of rich depending on what python version you are using (which could be confusing to users).

henryiii:
A few things - `rich` is fantastic, though you should be a little carful of using it in a _library_ - for example, a library should not replace the exception handler without a user specifically requesting it. Logging should be pretty safe, though, as long as it is built on the standard logger, which I think it was. It also is a Poetry project and follows Poetry's recommended bad practice of capping dependencies, so adding it to larger project could eventually cause version conflicts. But it's pretty safe, hopefully, it's updated regularly.

In this case, I'd either go with @lgray's suggestion of replacing `use_rich` with `rich is None`, or I'd make it a user option, say a keyword argument `rich=rich`.

maxgalli:
Makes sense, thanks. Changed accordingly.

andrzejnovak:
@maxgalli planning to revisit this? rich is now a dependency :)

maxgalli:
> @maxgalli planning to revisit this? rich is now a dependency :)

absolutely, I'm struggling a bit to find the time, but I'll do it soon

lgray:
@maxgalli after all this time - you win, closing this PR :-p

PR #568, opened 2021-08-10
Fix issue with unset parameters in NanoAODSchema

nsmith-:
As discovered upon opening any NanoAODv9 file with the new Proton collection﻿


nsmith-:
A workaround until this is in a release is:
```python

def patchv9(form):
    for key in list(form["contents"]):
        if key.startswith("Proton_"):
            form["contents"].pop(key)
    return NanoAODSchema(form)

events = NanoEventsFactory.from_root(filename, schemaclass=patchv9).events()
```
One can also pass `patchv9` as the `"schema"` executor argument of `run_uproot_job`.

lgray:
can you poke this PR? - github's instabilities yesterday were causing all sorts of CI troubles.

PR #570, opened 2021-08-17
Some performance improvements to nanoevents vectors

nsmith-:
20% overall improvement on ADL benchmarks


nsmith-:
@jrueb you might be interested in this

PR #571, opened 2021-08-19
wq: enable fast termination on workers

btovar:
Since Work Queue is build to run with heterogeneous resources, sometimes
some of those resources are significantly less powerful than the rest.
In such cases, it is better to terminate the slow workers and move the
tasks to other resources.

This pr activates the fast termination on workers of Work Queue. Tasks
that run for longer than task average time * fast_terminate_workers are
move to other workers, and the worker itself is released. Since there
are legitimately slow tasks, no task may trigger fast termination in two
different workers.

The default of fast_terminate_workers is None. Any value less than 1
disables it.

(This adds another parameter to the wq executor, which may produce a conflict with pr #557.)


btovar:
Thanks @lgray!

PR #572, opened 2021-08-20
[lumi-tools] Ensure pickability of `LumiData` and `LumiMask` classes

### LINKED ISSUE #569: not able to load LumiMask
**Describe the bug**
Cannot load LumiMask from a textfile.

**To Reproduce**
```
from coffea.lumi_tools import LumiMask
# load lumimask from e.g. this file https://github.com/cmantill/boostedhiggs/blob/main/boostedhiggs/data/Cert_294927-306462_13TeV_UL2017_Collisions17_GoldenJSON.txt
lumiMask = LumiMask("Cert_294927-306462_13TeV_UL2017_Collisions17_GoldenJSON.txt")
# call lumimask
lumiMask
```

Error:
```
/opt/conda/lib/python3.8/site-packages/coffea/processor/executor.py in run_uproot_job(fileset, treename, processor_instance, executor, executor_args, pre_executor, pre_args, chunksize, maxchunks, metadata_cache, dynamic_chunksize, dynamic_chunksize_targets)
   1619     else:
   1620         pi_to_send = lz4f.compress(
-> 1621             cloudpickle.dumps(processor_instance), compression_level=pi_compression
   1622         )
   1623     closure = partial(

/opt/conda/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py in dumps(obj, protocol, buffer_callback)
     71                 file, protocol=protocol, buffer_callback=buffer_callback
     72             )
---> 73             cp.dump(obj)
     74             return file.getvalue()
     75 

/opt/conda/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py in dump(self, obj)
    561     def dump(self, obj):
    562         try:
--> 563             return Pickler.dump(self, obj)
    564         except RuntimeError as e:
    565             if "recursion" in e.args[0]:

TypeError: cannot pickle '_nrt_python._MemInfo' object
```
**Expected behavior**
I would expect to be able to load the lumiMask when called as ```lumiMask(events.run, events.luminosityBlock)```.

pfackeldey:
Hi @nsmith- ,
 
 hope you don't mind that I took care of this.
 This PR closes #569 and ensures pickability of lumi-tools classes by moving `numba.typed.Dict` out of `__init__`.
 
 For performance the `Dict` calculations can be put additionally in e.g. `cached_property`. I am not sure how important this is though, since usually those functions are anyway only called once.
 
 Best, Peter

### DIFF
--- a/coffea/lumi_tools/lumi_tools.py
+++ b/coffea/lumi_tools/lumi_tools.py
@@ -34,28 +34,6 @@ class LumiData(object):
                 ],  # not sure what lumi:0 means, appears to be always zero (DAQ off before beam dump?)
             },
         )
-        self.index = Dict.empty(
-            key_type=types.Tuple([types.uint32, types.uint32]), value_type=types.float64
-        )
-        self.build_lumi_table()
-
-    def build_lumi_table(self):
-        """Build index for numba-compiled functions
-
-        This needs to be executed upon unpickling, it should be part of
-        a custom deserialize function.
-        """
-        runs = self._lumidata[:, 0].astype("u4")
-        lumis = self._lumidata[:, 1].astype("u4")
-        LumiData._build_lumi_table_kernel(runs, lumis, self._lumidata, self.index)
-
-    @staticmethod
-    @numba.njit(parallel=False, fastmath=False)
-    def _build_lumi_table_kernel(runs, lumis, lumidata, index):
-        for i in range(len(runs)):
-            run = runs[i]
-            lumi = lumis[i]
-            index[(run, lumi)] = float(lumidata[i, 2])
 
     def get_lumi(self, runlumis):
         """Calculate integrated lumi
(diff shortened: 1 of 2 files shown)


lgray:
My personal opinion - this looks fine and doesn't appear to break any user-facing interfaces.

Please add a test that ensures we do not break (un)pickling in the future!

coffea/lumi_tools/lumi_tools.py
```diff
@@ -123,18 +117,19 @@ def __call__(self, runs, lumis):
                 An array of dtype `bool` where valid (run, lumi) tuples
                 will have their corresponding entry set ``True``.
         """
+        # fill numba typed dict
+        _masks = Dict.empty(key_type=types.uint32, value_type=types.uint32[:])
+        for k, v in self._masks.items():
+            _masks[k] = v
```
nsmith-, line 123: I was at first a bit concerned that we re-build the numba typed dict on every evaluation (i.e. every chunk) but I decided to see how fast it was:
```
from coffea.lumi_tools import LumiMask
lumimask = LumiMask(
    "tests/samples/Cert_294927-306462_13TeV_EOY2017ReReco_Collisions17_JSON.txt"
)
from numba import types
from numba.typed import Dict
def test(m):
    _masks = Dict.empty(key_type=types.uint32, value_type=types.uint32[:])
    for k, v in m.items():
        _masks[k] = v

%timeit test(lumimask._masks)
```
and it only takes `822 µs ± 9.66 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)` so this is fine for a once-per-chunk operation. What I would have done is write `__getstate__` and `__setstate__` to convert the masks to numba dictionaries during the unpickling, but it doesn't make too much difference I guess.

nsmith- (approved):
Thanks for taking care of it!

pfackeldey:
@lgray hope this is sufficient as a test, I was not really sure how to add a `assert ...` test for this, however tests should fail now in case pickling is not possible anymore.

coffea/lumi_tools/lumi_tools.py
pfackeldey: Good point, I usually try to solve pickling things first by restructuring a bit and use `__getstate__` and `__setstate__` as a last rescue. 
Good test, that the performance is not impacted! :)
`numba.typed.Dict` is a _experimental_ feature according to the Numba docs, so maybe at some point this [should be/gets] on its own pickable in the future.

lgray:
@pfackeldey sorry to get nitpicky, but can you add a test for correctness after loading back in from a pickle dump? (i.e. make sure that normally constructed and constructed from pickle give same answers)

pfackeldey:
Yes sure, sorry for the sloppy previous test!

pfackeldey:
I hope this is better now :)

(I did not add more tests initially, since I have not modified any pickling-behaviour-affecting stuff like `__setstate__` and `__getstate__`. Thus I expected problems only to arise if `cloudpickle` doesn't do its job properly)

PR #573, opened 2021-08-20
[jersr-lookup] ensure no 0-dimensional arrays for `np.stack`

pfackeldey:
Hi @lgray @nsmith- ,

This PR fixes a very rare corner case, where only 1 jet enters the calculation.

The problem comes from the fact that `np.squeeze()` creates a 0-dimensional array if it has the length of 1:
```python
import numpy as np

np.array([[1, 2, 3]]).squeeze()
>> array([1, 2, 3]) # <- ndim == 1

np.array([[1]]).squeeze()
>> array(1)  # <- ndim == 0
```
In these cases `np.stack([...], axis=1)` failed.

Best, Peter

### DIFF
--- a/coffea/lookup_tools/jersf_lookup.py
+++ b/coffea/lookup_tools/jersf_lookup.py
@@ -98,7 +98,7 @@ class jersf_lookup(lookup_base):
 
         # get parameter values
         parm_values = numpy.stack(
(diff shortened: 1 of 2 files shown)


lgray:
thanks!

PR #575, opened 2021-08-20
move ipywidgets to spark spec. (reduce bloat)

lgray:
As in title.

I think @perilousapricot requested this for easy packaging in his cluster. 
I've set it up so that `pip install coffea[spark]` with spark already installed will do the same thing, that should be manageable?
Let me know if not, otherwise I'd just kinda prefer people `pip install coffea ipywidgets` if they need it like that.

This chops a little over 100MB off the install size for coffea and appropriately factorizes coffea from jupyter/ipython.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,14 +70,13 @@ INSTALL_REQUIRES = [
     "cloudpickle>=1.2.3",
     "mplhep>=0.1.18",
     "packaging",
-    "ipywidgets",
     "pandas",
     "hist>=2",
     'typing-extensions;python_version<"3.8"',
     "cachetools",
 ]
 EXTRAS_REQUIRE = {}
-EXTRAS_REQUIRE["spark"] = ["pyspark>=2.4.1,<3.0.0", "jinja2"]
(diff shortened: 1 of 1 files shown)


PerilousApricot:
:+1: thanks

nsmith-:
Added in #259, looks like it was something to do with `tqdm`?

lgray:
Yeah some ipy-kernel install business @PerilousApricot was doing triggered this.

Just to be clear: @PerilousApricot do you want `pip install coffea[spark]` to give you `ipywidgets` or do you wanna explicitly do `pip install coffea ipywidgets`?

PerilousApricot:
It's not a spark issue, per-se, (nick's comment reminded me). It's that
tqdm in a jupyter session will (was?) unconditionally try to use ipywidgets
and bombing even if it wasn't installed. I guess it's technically a bug in
TQDM? I'm pretty sure I put in an issue back when I first ran into it, but
I don't remember a resolution to it.

>

lgray:
I mean on our side we just stuck ipywidgets in coffea. :D

I'm happy to leave it in the `[spark]` specialization. Just remember to update your build scripts. If you change to `pip install coffea[spark]` now it'll be seamless when the new version comes on Sept. 3rd.

PR #578, opened 2021-08-27
Drop abseta assumption in btag tools

nsmith-:
Not entirely sure why this restriction was there before.


nsmith-:
Another example of why correctionlib will be easier to maintain

PR #579, opened 2021-09-03
wq: use a memory target to dynamically adapt the chunksize of processing tasks

btovar:
This pr allows to use the work queue executor as:

```python
run_uproot_job(..., executor=work_queue_executor, chunksize=base_chunksize, dynamic_chunksize={"memory": target_memory})
```
where `base_chunksize` is some intial chunksize, and `target_memory` is a desired memory usage per processing job. As tasks are run, data is collected to find some relation between the chunksize and the memory consumption. The data is used to compute new chunksizes for later tasks. The final computed chunksize can then be used as the base_chunksize for later runs.

This can be used in conjunction with the `"memory"` argument of executor arguments to ensure that processing jobs are executed in the available resources.  If a job goes above the cores, memory, or disk given in the execution arguments, the processing job is split in two and retried, until the maximum number of retries is reached.

Note that this pr simplifies `dynamic_chunksize_targets`. Instead it converts `dynamic_chunksize` into a dictionary with the targets. Before, `dynamic_chunksize` was simply a boolean to activate the dynamic chunksize computation.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -978,7 +978,7 @@ def _chunk_generator(
     chunksize,
     align_clusters,
     maxchunks,
-    dynamic_chunksize=False,
+    dynamic_chunksize=None,
 ):
     if maxchunks is None:
         for filemeta in fileset:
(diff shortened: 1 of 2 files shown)


lgray:
let me know if you have any more to add - if not I will merge this and get it into coffea 0.7.7.

btovar:
@lgray: it is ready, thanks!

PR #584, opened 2021-09-13
feat: NanoEvents Delphes

kratsg:
This PR will aim to add functionality for coffea to read in Delphes files.

Docstrings are added for common names. The following were more complicated branches that needed to be handled / fixed in uproot first initially before being handled correctly in coffea. The schema is modified in a custom way to drop in the `vector.LorentzVector` behavior for those that are TLVs.

- [x] `Area` (TLorentzVector)
- [x] `Constituents` (TRefArray)
- [x] `Edges[4]` (Float_t)
- [x] `FracPt[5]` (Float_t)
- [x] `Particle` (TRef)
- [x] `Particles` (TRefArray)
- [x] `PrunedP4[5]` (TLorentzVector)
- [x] `SoftDroppedJet` (TLorentzVector)
- [x] `SoftDroppedP4[5]` (TLorentzVector)
- [x] `SoftDroppedSubJet1` (TLorentzVector)
- [x] `SoftDroppedSubJet2` (TLorentzVector)
- [x] `Tau[5]` (Float_t)
- [x] `TrimmedP4[5]` (TLorentzVector)

We don't know why `Area` is a TLV.


coffea/nanoevents/mapping/uproot.py
```diff
@@ -82,6 +85,18 @@ def _extract_base_form(cls, tree, iteritems_options={}):
                         form["content"]["content"]["contents"][field][
                             "form_key"
                         ] = quote(f"{key},!load,!content,!content,{field},!item")
+            elif (
+                form["class"].startswith("ListOffset")
+                and form["content"]["class"] == "RecordArray"
+            ):
+                # TODO: fix me?
```
kratsg: @nsmith- the rest of the uninterpretable items are hitting this new conditional here. I have no idea what magic is going on with `form_key` so I couldn't get this working.

coffea/nanoevents/mapping/uproot.py
```diff
@@ -67,7 +67,10 @@ def _extract_base_form(cls, tree, iteritems_options={}):
                 form["content"]["parameters"] = {"__doc__": branch.title}
             elif (
                 form["class"].startswith("ListOffset")
-                and form["content"]["class"].startswith("ListOffset")
+                and (
+                    form["content"]["class"].startswith("ListOffset")
+                    or form["content"]["class"] == "RegularArray"
+                )
```
kratsg: This addition seems to allow for getting `Edges[4]`, `Tau[5]`, and `FracPt[5]` to work correctly(?) maybe? At least it seems to be ok with uproot.

coffea/nanoevents/schemas/delphes.py
```diff
@@ -0,0 +1,290 @@
...
+        for name in collections:
+            if f"{name}_size" in branch_forms:
+                branch_forms[f"o{name}"] = transforms.counts2offsets_form(
+                    branch_forms[f"{name}_size"]
+                )
+
+        for name in collections:
+            for key in self.tlorentzvectors:
+                if f"{name}/{name}.{key}" in branch_forms:
+                    ...  # do stuff
```
kratsg: I imagine we need to do something for any of the TLV cases, once they've been defined properly in `uproot/mapping.py` and "transform" them accordingly.

nsmith-: These are actually uproot fault, so I'll fix upstream

nsmith-:
With scikit-hep/uproot4#441 and scikit-hep/uproot4#442 I can now do:
```python
import awkward as ak
from coffea.nanoevents import NanoEventsFactory, DelphesSchema

    "tests/samples/delphes.root",
    "Delphes",
    schemaclass=DelphesSchema,
).events()
ak.to_list(events)
```
which confirms we can at least interpret all the branches successfully.

kratsg:
Using

```
pip install -U git+git://github.com/scikit-hep/uproot4.git@007b10014003a74d0e89e0166e21a6218de0d445
```

and an editable install at the current commit `baa2018`, i have the following

```
>>> ak.type(data.Jet[0].Area)
2 * struct[["fE", "fP"], [float64, TVector3["fX": float64, "fY": float64, "fZ": float64]], parameters={"__doc__": "Area[Jet_]"}]
>>> data.Jet[0].Area
<Array [{fE: 0, fP: {fX: 0, ... fZ: 0}}] type='2 * struct[["fE", "fP"], [float64...'>

>>> ak.type(data.Jet[0]['PrunedP4[5]'])
2 * [5 * struct[["fE", "fP"], [float64, TVector3["fX": float64, "fY": float64, "fZ": float64]], parameters={"__doc__": "PrunedP4[Jet_]"}], parameters={"__doc__": "first entry (i = 0) is the total Pruned Jet 4-momenta and from i = 1 to 4 are the pruned subjets 4-momenta"}]
>>> data.Jet[0]['PrunedP4[5]']
<Array [[{fE: 0, fP: {fX: 0, ... fZ: 0}}]] type='2 * [5 * struct[["fE", "fP"], [...'>
```

and the others look ok

```
>>> data.Jet[0].SoftDroppedSubJet1
<Array [{fE: 111, fP: {, ... fZ: 17.5}}] type='2 * struct[["fE", "fP"], [float64...'>
>>> data.Jet[0]['TrimmedP4[5]']
<Array [[{fE: 0, fP: {fX: 0, ... fZ: 0}}]] type='2 * [5 * struct[["fE", "fP"], [...'>
>>> data.Jet[0]['PrunedP4[5]']
<Array [[{fE: 0, fP: {fX: 0, ... fZ: 0}}]] type='2 * [5 * struct[["fE", "fP"], [...'>
>>> data.Jet[0]['SoftDroppedP4[5]']
<Array [[{fE: 126, fP: {, ... fZ: 0}}]] type='2 * [5 * struct[["fE", "fP"], [flo...'>
```

So I want to add quicker access for some of the branches that have brackets in the name, and it would be nice to automatically interpret some of these as TLVs -- and that requires wrapping it somehow? Or is it expected that `uproot` or `awkward` is supposed to recognize it automatically? I thought Jim mentioned something about that.

kratsg:
Can confirm it all looks ok without the TLV interpretations

```
>>> data.Jet.Area
<Array [[{fE: 0, fP: {fX: 0, ... fZ: 0}}]] type='25 * var * struct[["fE", "fP"],...'>
>>> data.Jet.Constituents
<Array [[{fName: '', fSize: 12, ... 5932]}]] type='25 * var * struct[["fName", "...'>
>>> data.Jet.Particles
<Array [[{fName: '', fSize: 12, ... 1929]}]] type='25 * var * struct[["fName", "...'>
>>> data.Jet.PrunedP4_5
<Array [[[{fE: 0, fP: {fX: 0, ... fZ: 0}}]]] type='25 * var * [5 * struct[["fE",...'>
>>> data.Jet.SoftDroppedJet
<Array [[{fE: 126, fP: {, ... fZ: 44.2}}]] type='25 * var * struct[["fE", "fP"],...'>
>>> data.Jet.SoftDroppedP4_5
<Array [[[{fE: 126, fP: {, ... fZ: 0}}]]] type='25 * var * [5 * struct[["fE", "f...'>
>>> data.Jet.SoftDroppedSubJet1
<Array [[{fE: 111, fP: {, ... fZ: 39.2}}]] type='25 * var * struct[["fE", "fP"],...'>
>>> data.Jet.SoftDroppedSubJet2
<Array [[{fE: 15.2, fP: {, ... fZ: 4.98}}]] type='25 * var * struct[["fE", "fP"]...'>
>>> data.Jet.Tau_5
<Array [[[0.197, 0.148, ... 0.0485, 0.0356]]] type='25 * var * [5 * float32[para...'>
>>> data.Jet.TrimmedP4_5
<Array [[[{fE: 0, fP: {fX: 0, ... fZ: 0}}]]] type='25 * var * [5 * struct[["fE",...'>
```

nsmith-:
Certainly the `vector` interpretation of TLorentzVector directly via uproot will come in some form eventually, but anyway we'll also have to move coffea vectors to `vector`. For now I think it would make more sense to consistently use coffea vectors and when we move we can just adjust these as appropriate. To do that I think the easiest is to just define a mixin for the TLV that provides properties `x y z t` that then access the objects `self["fP"]["fX"], ..., self["fE"]`

nsmith-:
Another option is to hack at the form to turn the record items into x y z t.

kratsg:
> Another option is to hack at the form to turn the record items into x y z t.

I went with this option for hacking at the form using a recursive function.

coffea/nanoevents/schemas/delphes.py
```diff
@@ -0,0 +1,296 @@
...
+                ] = self.docstrings.get(
+                    parameter,
+                    output[name]["content"]["contents"][parameter]["parameters"].get(
+                        "__doc__", "no docstring available"
+                    ),
+                )
+
+            # handle branches named like [4] and [5]
+            output[name]["content"]["contents"] = {
+                k.replace("[", "_").replace("]", ""): v
```
nsmith-, line 274: Since the extra dimension size can be inferred from the type, do you think it makes sense to drop also the number (i.e. `Thing` instead of `Thing_5` etc.) ? I'm ambivalent

nsmith- (approved):
One last check I would suggest is make sure `_ = ak.to_list(events)` runs. This makes sure all arrays materialize successfully. The output will be gigantic so best not to print it :)

coffea/nanoevents/schemas/delphes.py
kratsg: Initially I was thinking this, but I got worried that there might be a "PrunedP4" branch later and I just wanted to convert the non-python-dot-syntax-friendly characters to something more python-friendly. If this is something you're not necessarily concerned about, we could hard-code a regex replacement of `\[\d\]` instead.

nsmith-: That's a good point, there may be a future name clash if for some reason `PrunedP4` and also `PrunedP4[5]` are available.

kratsg: `Tau[4]` also might name-clash for (obvious) reasons. So i was stuck here :-/

lgray: What's the decision here? Shall we go ahead and merge as is?

PR #590, opened 2021-09-17
wq: use correct statistic when collecting task execute time

btovar:
execute_cmd_finish and execute_cmd_start measured from submission time, not execution time.


lgray:
@btovar for these minor changes and tweaks could you merge them together into single PRs in the future? Unless they're clobbering each other or affect tested functionality in a major way I don't think there much reason to have a dust of one-line PRs.

btovar:
> @btovar for these minor changes and tweaks could you merge them together into single PRs in the future? Unless they're clobbering each other or affect tested functionality in a major way I don't think there much reason to have a dust of one-line PRs.

@lgray sounds good!

PR #591, opened 2021-10-04
wq small fixes

btovar:
Mainly adds a measure only mode, high-throughput mode option for processing task, and captures output of running tasks to a file for debugging.


lgray:
Great - thanks for continuing to tweak and improve!

PR #593, opened 2021-10-06
Fixing 3-vector like behaviors for ROOT::Math::XYZPoint in TreeMaker ntuples

yimuchen:
- Instead of zipping XYZPoint coordinate leaves as "Points", default to using the `"ThreeVector"` behavior defined in `methods.vector`.
- Simple dictionary look up to allow for overriding of behaviors of an object collection.


lgray:
Looks good - thanks!

PR #600, opened 2021-10-26
wq: tee command was hiding exit code of task

btovar:
Now stdout is written to a file, which is read only if needed.

### DIFF
--- a/coffea/processor/work_queue_tools.py
+++ b/coffea/processor/work_queue_tools.py
@@ -96,6 +96,7 @@ class CoffeaWQTask(Task):
         self.itemid = itemid
 
         self.py_result = ResultUnavailable()
+        self._stdout = None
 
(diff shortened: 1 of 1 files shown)


lgray:
@maxgalli is working on integrating proper logging into coffea. We should make sure this is covered as well if you think its the right way to go @btovar .

btovar:
We are doing it this way because we had trouble with xrootd timing out and just blocking forever. In order to debug it and get the error print out we had to kill the tasks manually and wait for the stdout to be printed in the main coffea console. With this change, we can go to any execution site and see directly what the task is doing as it runs (e.g. with `export XRD_LOGLEVEL=Debug`).

I also think that having the stdout in a file may be beneficial later. Maybe we don't delete right away, and let a user see the debug outputs of all tasks after the run has ended. (This is something that parsl does, for example.)

Issue #601, opened 2021-10-26
DaskExecutor method doesn't support possibility to use an externally initialized Dask Client 


oshadura:
**Describe the bug**
DaskExecutor() doesn't support possibility to use an externally initialized Dask Client

Within DaskExecutor method Dask Client is initialized internally and we dont have any control on it except of  client_add (see related code https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/servicex/dask_executor.py#L41-L45)
```
executor = DaskExecutor(client_addr="tls://localhost:8786")
```

Ideally what we would like is to be able to use rich Dask Client API,  for example to be able upload files to the workers: 
```
client = Client("tls://localhost:8786")
import zipfile
zipfile.ZipFile("infofile.zip", "w").write("infofile.py")
client.upload_file("infofile.zip")
executor = DaskExecutor(client=client)
```

cc: @BenGalewsky 

Related to https://github.com/iris-hep/analysis-grand-challenge/issues/1

lgray:
This is only for the servicex flavor of the DaskExecutor.

https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L584 lets you pass in a client.

It would be a good idea to figure out how we can merge them!

oshadura:
@lgray yup, we are testing **Servicex** flavor of Dask executor for AGC Tools workshop and I wanted to report that we were not able to run any complex analysis with dependencies... (cc: @alexander-held )

And I totally agree, it will be nice to understand how to merge them :)

andrzejnovak:
Fixed in #602

PR #602, opened 2021-10-27
Allow pre-configured Dask client in DaskExecutor constructor

BenGalewsky:
# Problem
Users may want to prefigure their Dask client and not use the trivial setup provided by the DaskExecutor constructor.

Solution to #601 

# Approach
Added a second optional argument to the DaskExecutor constructor to allow a preconfigured client to be passed in.


oshadura:
cc: @lgray

PR #608, opened 2021-10-30
Only test latest python and 3.6

lgray:


lgray:
@gordonwatts @BenGalewsky funcadl needs python 3.10 wheels :-)

gordonwatts:
Ah- ok! I didn't realize 3.10 was out!

PR #616, opened 2021-11-11
Make Parsl stop serializing the entire Runner instance

jrueb:
Parsl serializes the callables provided to it and for each chunk keeps one copy of serialized buffer in memory locally. Currently in case of Coffea these are bound methods of Runner, which makes Parsl serialize the entire Runner instance. However the Runner instance contains things like caches and executors, which can be user-provided and very large. This makes the serialized buffer very large as well. In my tests I easily got buffer sizes of ~ MB and with 10000 chunks, which is a normal number for my applications, this amounts to ~ 10 GB. All this memory is occupied without any good reason. In some extreme cases I was even able to fill up all the memory of my machine.

This patch makes the methods that are serialized static, so that the serialization works independent of the Runner instance. Affected methods are `config`, `automatic_retries`, `metadata_fetcher` and `_work_function`.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -932,45 +932,45 @@ class Runner:
         else:
             return False
 
-    @cached_property
-    def cache(self):
+    @staticmethod
+    def get_cache(cachestrategy):
         cache = None
-        if self.cachestrategy == "dask-worker":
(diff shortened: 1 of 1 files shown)


lgray:
Thanks for catching this!

PR #621, opened 2021-11-30
Add bin span as xerr in plotrange.

kkrizka:
Adds an option to plot the size of the bin in `plotrange` as an x-axis error bar. This makes it easier to follow a curve, especially if variable binning is used.

If the new `xerr` (default is false) argument is true, then the `xerr` argument to `errorbar` that draw the ratio curve is set to half of each bin width.

Example of this style can be seen in [ROOT's TEfficiency](https://root.cern.ch/doc/master/classTEfficiency.html), [ATLAS](https://twiki.cern.ch/twiki/pub/AtlasPublic/JetTriggerPublicResults/L1J100_ATLASprelim.png) / [CMS](https://twiki.cern.ch/twiki/bin/view/CMSPublic/HLTplots2018BoostedHiggsbb) trigger efficiency plots.

### DIFF
--- a/coffea/hist/plot.py
+++ b/coffea/hist/plot.py
@@ -419,7 +419,7 @@ def plotratio(
         ax.set_ylabel(num.label)
         edges = axis.edges(overflow=overflow)
         centers = axis.centers(overflow=overflow)
-        ranges = (edges[1:] - edges[:-1])/2 if xerr else None
+        ranges = (edges[1:] - edges[:-1]) / 2 if xerr else None
 
         sumw_num, sumw2_num = num.values(sumw2=True, overflow=overflow)[()]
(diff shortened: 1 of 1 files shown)


lgray:
@kkrizka Thanks, your contribution is appreciated. However, could you please check at https://github.com/scikit-hep/mplhep and https://github.com/scikit-hep/hist that there is not a better place to put this addition which has longer term support?

Eventually we'd like to phase out coffea.hist since there are better packages, and coffea.hist is a prototype. I suppose I'm fine with merging it here, but take this as fair warning we prefer that people use other histogramming / plotting packages now.

lgray:
@kkrizka please fix the linting issues.

kkrizka:
Hi @lgray, thank you for pointing me to mplhep. I wasn't aware that this was the preferred way for long-term. I will definitely look into it.

Also thank you for fixing the formatting.

PR #622, opened 2021-12-05
Fix small bug in Runner

maxgalli:
A small bug was introduced in https://github.com/CoffeaTeam/coffea/commit/cf4dd421d37fe6118c576bbdb6668686b5681b39, not detected by the tests because the lines affected were never hit. This PR tweaks one of the tests to hit this case too and fixes the bug.

Tutorial https://github.com/CoffeaTeam/coffea/blob/master/binder/nanoevents.ipynb, which was affected by the change, is then fixed as well.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -1077,7 +1077,13 @@ class Runner:
                     continue
                 # fmt: on
                 if not filemeta.populated(clusters=self.align_clusters):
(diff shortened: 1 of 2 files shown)


nsmith- (approved):
Thanks!

lgray:
Just a sec before we make a release - this patch does not fix everything. I'll make another PR in a bit.

Issue #375, opened 2020-11-30
A bug was found in coffea.Release v0.6.46 skipbadfiles true


scarletnorberg:
**Describe the bug**
When using 'skipbadfiles':True, the code will skip over the current input files. The order of the code was changed recently in order to fix a different bug: https://github.com/CoffeaTeam/coffea/commit/97f26761e0e52d023c290f811e259edab9aa14de
This in turn then lead to the logic error.

**To Reproduce**
1. Set 'skipbadfiles':True, 

**Expected behavior**
https://github.com/CoffeaTeam/coffea/compare/f663640..384cc79#diff-2d9bea3caaa598242f68714de8fce6dce0f181d8f7462800ef62b6710bced905R1077-R1081
This is the source of the problem with the TreeMaker files, but I assume it will also be a problem with other files. The logic doesn't make sense:
1. if skipbadfiles=True then check for metadata and if it doesn't exist, continue.
2. Check for metadata and if it doesn't exist, add it.
This seems backwards.

**Output**
There is no error just nothing is output.

**Desktop (please complete the following information):**
This is the test code that we have been using to test: https://gitlab.cern.ch/cms-emj/EMJAnalysis/-/blob/dev-snorberg/notebooks/Mini_MakePlots.ipynb

**Additional context**
Currently we are just going to change the true to false. This seems to be the case for all files because the testing I think by default sets it to false for skip bad files.

nsmith-:
@andrzejnovak does this make sense to you? I think the original ordering with the addition of a check if the return value of the populate function is None is the fix.

andrzejnovak:
I don't see a reason the conditions should be together, might help readability. 

IIRC the reason for  [97f2676](https://github.com/CoffeaTeam/coffea/commit/97f26761e0e52d023c290f811e259edab9aa14de) is that [this](https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L1079) chunk was crashing if it ran on a bad file or was causing a crash further down the line. 

I am not sure what's different about the reported case and the test introduced in [97f2676](https://github.com/CoffeaTeam/coffea/commit/97f26761e0e52d023c290f811e259edab9aa14de), which seems to have run fine.

@scarletnorberg If you can separate the logic and the test added in the linked PR doesn't crash, the issue I was seeing should not come back.

lgray:
Did this reach any sort of resolution?

PR #623, opened 2021-12-06
More exe fixes

lgray:
Further failure modes covered.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -1034,52 +1034,38 @@ class Runner:
         return out
 
     def _preprocess_fileset(self, fileset: Dict) -> None:
-        if self.maxchunks is None:
-            # this is a bit of an abuse of map-reduce but ok
-            to_get = set(
-                filemeta
-                for filemeta in fileset
-                if not filemeta.populated(clusters=self.align_clusters)
+        # this is a bit of an abuse of map-reduce but ok
+        to_get = set(
+            filemeta
+            for filemeta in fileset
+            if not filemeta.populated(clusters=self.align_clusters)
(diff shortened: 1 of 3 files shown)


lgray:
@nsmith-

coffea/processor/executor.py
```diff
@@ -1071,11 +1071,6 @@ def _preprocess_fileset(self, fileset: Dict) -> None:
                     filemeta.maybe_populate(self.metadata_cache)
         else:
             for filemeta in fileset:
-                # not sure why need to check for bad files here... otherwise pop fails below with pytest.
-                # fmt: off
-                if self.skipbadfiles and not filemeta.populated(clusters=self.align_clusters):  # noqa  # fmt: skip <-- gets ignored, see: https://github.com/psf/black/issues/2421
```
nsmith-, line 1076: I guess this is no longer failing? It seems `_filter_badfiles` does this job later anyway?
Maybe #375 is fixed now as well?

lgray: It goes a little deeper than this. Fixing issues now.

lgray: Basically, there shouldn't be a if-block here for maxchunks.

nsmith-: Doesn't this make the pre-executor open all files even if only one chunk is desired then? The point of special handling when maxchunks=True was originally to speed up testing on one file #86

lgray: Well, we were doing it in a funny way if that was the case. This is making it more clear we want to be able to pass in pickled or otherwise cached preprocessing.

PR #627, opened 2021-12-13
wq-example.py not running out of the box

David-Simonetti-ND:
Taking the wq-example.py and running it straight off the documentation does not work because some of the executor arguments had a dash changed to an underscore. (ex master-name -> master_name). Also commented out resources_mode and resource_monitor because they also create errors. Now the wq-example.py is able to run without any changes.


btovar:
@David-Simonetti-ND could you compare with the example used for CI in `tests/wq.py`?

The `wq-example.py` in the docs looks a little out of date. We may want to simply use or adapt the example in `tests/wq.py` in the documentation.

lgray:
@btovar @David-Simonetti-ND is this ready for merge?

btovar:
@lgray Not quite. We are thinking on adapting the wq CI test as the documentation example. The current example in the documentation is a little out of date (e.g. it does not use the new Runner).

lgray:
ping!

btovar:
Sorry for the delay, we were on the holiday break. I should have a pr this week.

PR #629, opened 2021-12-16
Better error message for KilledWorker in dask executor

nsmith-:


nsmith-:
pypi is down? https://status.python.org/

lgray:
@nsmith- 
```
TypeError: Dask no longer supports mapping over Iterators or Queues.Consider using a normal for loop and Client.submit
```

New dask failure.

nsmith-:
ah dumb mistake, good thing we test worker_affinity

PR #630, opened 2021-12-20
Add reaction to KeyboardInterrupt when using Parsl

jrueb:
Currently when is running a Processor using the ParslExecutor and trying to stop it by pressing Ctrl+C, the only thing that happens is that a warning is printed. The content of this warning is incorrect. This PR changes the behavior when using Parsl to stopping the script, like it normally would. For cases where one is not using Parsl, nothing changes.

Background:
Parsl has not implemented canceling jobs. Coffea tries to cancel jobs on `KeyboardInterrupt` but ignores the `NotImplementedError`. However, even if using Parsl a warning is still printed by Coffea telling the user that all jobs are canceled.
So effectively nothing happens when Ctrl+C is pressed together with the ParslExecutor, which is not helpful at all, because Ctrl+C is the first method the user tries (and sometimes even the only method the user knows) to end the script.

Not gracefully canceling is jobs is not optimal, but the only thing available in Parsl.


jrueb:
Instead of re-raising the `KeyboardInterrupt`, one could also have a `break`, similarly to the case in which the `timeout` is hit. This could save possible output but takes of course longer to do.

PR #632, opened 2021-12-26
Adding ProtoDUNE schema

tejinc:
Working copy of pdune schema. Hacked physlite and base.py because pduneana schema needs a modification of zip_form


tejinc:
I couldn't understand why this new iteration of a check failed:
- testing on python 3.9 succeeded on my local system
- a previous CI/CD check was successful, why did it fail now?

lgray:
@tejinc it's not your fault - there's a new version of dask out that removed some patterns we were using.

lgray:
@tejinc can you please add a small ProtoDune file (< 5MB please) to `tests/samples/` and make a test for your schema?

tejinc:
> @tejinc can you please add a small ProtoDune file (< 5MB please) to `tests/samples/` and make a test for your schema?

@lgray The commit does contain a small data file with very basic file opening/close test. The data file is called: pduneana.root and the test is test_nanoevents_pdune.py

lgray:
ah - missed it - thanks!

PR #633, opened 2022-01-05
updates docs/source/wq-example.py to Runner()

btovar:
Supersedes #627 

- Updates to Runner() from run_uproot_job
- Updates keywords from - to_
- Uses a wq.Factory to launch local workers


lgray:
I'll figure out these weird parsl errors. Thanks for the update!

btovar:
Thanks @lgray!

Issue #634, opened 2022-01-18
[Skyhook] In NanoEventsFactory `from_parquet` method: `skyhook_options` doesn't support different cluster pool names


oshadura:
Trying to run https://github.com/CoffeaTeam/coffea/blob/master/binder/nanoevents_pq.ipynb on coffea-casa AF facility with Skyhook deployment and running next cell:
```
events_skyhook = NanoEventsFactory.from_parquet("/mnt/cephfs/nanoevents/ZJets/nano_dy.0.parquet", skyhook_options = {"ceph_config_path": "/home/cms-jovyan/.ceph/ceph.conf", "cephfs-data0"}).events()
events_skyhook.Muon
```
we got an invalid syntax error: 
```
ile "/tmp/ipykernel_1373/3224886486.py", line 1
    events_skyhook = NanoEventsFactory.from_parquet("/mnt/cephfs/nanoevents/ZJets/nano_dy.0.parquet", skyhook_options = {"ceph_config_path": "/home/cms-jovyan/.ceph/ceph.conf", "cephfs-data0"}).events()
                                                                                                                                                                                               ^
SyntaxError: invalid syntax
```
Looks like `skyhook_options` doesn't support different cluster pool names.

cc: @JayjeetAtGithub

PR #635, opened 2022-01-18
Support a different ceph pool in skyhook and refactor skyhook args

JayjeetAtGithub:
To switch to using Skyhook, set the `use_skyhook` flag to True in the executor args. By default, this will use `/etc/ceph/ceph.conf` and `cephfs_data` for the config file and data pool respectively. To use custom options, use the `skyhook_options` dict in executor args along with the `use_skyhook` flag to provide the values.

```python
executor_args={"use_skyhook": True, "skyhook_options": {"ceph_config_path": "/somepath/ceph.conf", "ceph_data_pool": "mypool"}}
```

@oshadura, this is a temporary fix for now. We can change this later after we figure out the best way to provide custom options.


JayjeetAtGithub:
Resolves #634

Issue #637, opened 2022-02-01
Skyhook tests fail CI


lgray:
@JayjeetAtGithub 

Docker build fails as:
```
Status: Downloaded newer image for uccross/skyhookdm-arrow:v0.4.0
 ---> c1a67b127273
Step 2/2 : RUN yum -y update &&     yum -y install python3                    python3-devel                    python3-pip                    llvm-devel &&     pip3 install pip --upgrade
 ---> Running in f450c5826367
CentOS Linux 8 - AppStream                      347  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
The command '/bin/sh -c yum -y update &&     yum -y install python3                    python3-devel                    python3-pip                    llvm-devel &&     pip3 install pip --upgrade' returned a non-zero code: 1
Unable to find image 'coffea-skyhook-test:latest' locally
docker: Error response from daemon: pull access denied for coffea-skyhook-test, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```

JayjeetAtGithub:
Let me try to fix this.

PR #638, opened 2022-02-02
Fix the skyhook failed builds

JayjeetAtGithub:


lgray:
thanks a lot!

lgray:
Fixes #637

PR #643, opened 2022-02-23
try awkward 1.8rc

lgray:
temporarily alter ci to use pre-release awkward


lgray:
(the most recent change will not work until rc6 is in pypi)

Issue #97, opened 2019-06-07
make progress of futures based on callbacks


aminnj:
Would there be opposition to changing
https://github.com/CoffeaTeam/coffea/blob/1b0b8fac4ce0d774e4c8785a2c43188b95c12667/coffea/processor/executor.py#L43-L51
into
```python
for job in tqdm(concurrent.futures.as_completed(futures),disable=not status, unit=unit, total=len(futures)):
    accumulator += job.result()
```
which relies on callbacks instead of polling every 1s (I guess this saves 1/2 sec on average). Also, `as_completed` can take a timeout parameter.

If you think it's useful, I can make a small PR later today.

lgray:
@aminnj yeah I use as_completed for spark and parsl. It would be nice to uniformize the code a bit. 

If you wanna give it a go, please do. If you'd rather I roll it into the PR I'll take care of for #96, once I'm off this bus, that's fine too.

aminnj:
Turned out it only took a couple mins so I put in a PR (I thought I had screwed up my setup earlier because of the repo name change :) ). If you want to lump it with the other change, go ahead.

lgray:
Sure, I'll grab the commit from your PR in a bit.

Thanks!

nsmith-:
Please don't, there's a good reason not to use as_completed.  It queues up the incoming results *in the order of execution* so if one job stalls and the others continue, the queue will fill up with results.  If the result package is large the driver will run out of memory.  I originally used as_completed but moved away from it due to running into this problem.

nsmith-:
That being said, I would prefer a callback but unless one exists that doesn't guarantee the order of results matches the order of executor submission, I would highly advise against it.

lgray:
Thanks for the heads up. Will leave as is. I'll try out what you're doing with spark / parsl.

aminnj:
Sorry I'm a bit confused. I thought the iterator returns futures for things that complete first without regard to the order in which they were `executor.submit`ted based on
```python
import concurrent.futures
import time
import random

def work(i):
    time.sleep(random.random())
    return i

executor = concurrent.futures.ProcessPoolExecutor(4)
futures = [executor.submit(work,i) for i in range(20)]
for future in concurrent.futures.as_completed(futures):
    print future.result()
```
so that if one future takes forever, the others will have been aggregated by the time that one finishes

nsmith-:
[...]

PR #646, opened 2022-03-07
feat: recoverable processing + merging

andrzejnovak:
We discussed in the past how it's bad when a big job crashes and many hours and core hours can be lost. Here's a draft of how we could go about solving that (ignore some of the merge logic, I used the `FutureHolder` class from my job merging PR because this needs a bit finer control than the `_futures_handler` generator, but it's not necessarily connected)

Workflow can look like this. 

```
run_instance = processor.Runner(....
)

filemeta = run_instance(samples, "Events", processor_instance=MyZPeak(), prepro_only=True)` 
# save it in a pickle or sth for reference

# passing existing pre-made chunks to the Runner instance will skip the prepro step
output, processed, metrics = run_instance(filemeta , "Events", processor_instance=MyZPeak())` 
# WorkItem is attached to the accumulator like metrics are
# if exception is raised, report it, but return futures that already finished

missing = set(filemeta) - set(processed)
# resubmit
output, processed, metrics = run_instance(list(missing), "Events", processor_instance=MyZPeak())` 
```

@nsmith-

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -1734,12 +1734,18 @@ class Runner:
         else:
             processor_instance.postprocess(wrapped_out["out"])
 
+        def make_j_serializable(metrics):
+            for k, v in metrics.items():
+                if isinstance(v, set):
+                    metrics[k] = list(v)
+            return metrics
+
         _return = (wrapped_out["out"],)
         if hasattr(self.executor, "recoverable") and self.executor.recoverable:
             _return = *_return, list(wrapped_out["processed"])
         if self.savemetrics and not self.use_dataframes:
             wrapped_out["metrics"]["chunks"] = len(chunks)
-            _return = *_return, wrapped_out["metrics"]
+            _return = *_return, make_j_serializable(wrapped_out["metrics"])
         return _return if len(_return) > 1 else _return[0]
 
 
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -587,7 +587,7 @@ class FuturesExecutor(ExecutorBase):
         merging : bool | tuple(int, int, int), optional
             Enables submitting intermediate merge jobs to the executor. Format is
             (n_batches, min_batch_size, max_batch_size). Passing ``True`` will use default: (5, 4, 100).
-            Default is ``None`` - results get merged as they finish in the main process.
+            Default is ``False`` - results get merged as they finish in the main process.
         mergepool : concurrent.futures.Executor class or instance | int, optional
             Supply an additional executor to process merge jobs indepedently.
             An ``int`` will be interpretted as ``ProcessPoolExecutor(max_workers=int)``.
@@ -606,7 +606,7 @@ class FuturesExecutor(ExecutorBase):
             bool,
         ]
     ] = None
-    merging: Optional[Union[bool, Tuple[int, int, int]]] = False
+    merging: Union[bool, Tuple[int, int, int]] = False
     workers: int = 1
     recoverable: bool = False
     tailtimeout: Optional[int] = None
@@ -631,47 +631,49 @@ class FuturesExecutor(ExecutorBase):
                 set(pool.submit(function, item) for item in items), refresh=2
             )
 
-            if isinstance(self.merging, tuple) and len(self.merging) == 3:
-                nparts, minred, maxred = self.merging
-            else:
-                nparts, minred, maxred = 5, 4, 100
+            nparts, minred, maxred = 5, 4, 100
+            if isinstance(self.merging, tuple):
+                if len(self.merging) == 3:
+                    nparts, minred, maxred = self.merging
 
             try:
                 with rich_bar() as progress:
-                    prog_id = progress.add_task(
+                    p_id = progress.add_task(
                         self.desc, total=len(items), unit=self.unit
                     )
                     _mdesc = "Merging" if self.merging else "Merging (local)"
-                    prog_id_merge = progress.add_task(_mdesc, total=0, unit="merges")
+                    p_id_m = progress.add_task(_mdesc, total=0, unit="merges")
 
                     merged = None
                     while len(FH.futures) + len(FH.merges) > 0:
                         FH.update()
-                        reduce = min(
+                        merge_size = min(
                             maxred, max(len(FH.completed) // nparts + 1, minred)
                         )
                         progress.update(
-                            prog_id,
+                            p_id,
                             advance=FH.done["futures"]
-                            - progress._tasks[prog_id].completed,
+                            - progress._tasks[p_id].completed,
                         )
 
                         if self.merging:
                             progress.update(
-                                prog_id_merge,
+                                p_id_m,
                                 advance=FH.done["merges"]
-                                - progress._tasks[prog_id_merge].completed,
+                                - progress._tasks[p_id_m].completed,
                                 refresh=True,
                             )
                             while len(FH.completed) > 1:
-                                batch = [b for b in FH.fetch(reduce)]
+                                if len(FH.futures) + len(FH.merges) > 0 and len(FH.completed) < minred:
+                                    break
+                                batch = [b for b in FH.fetch(merge_size)]
                                 if mergepool is None:
                                     FH.merges.add(pool.submit(reducer, batch))
                                 else:
                                     FH.merges.add(mergepool.submit(reducer, batch))
                                 progress.update(
-                                    prog_id_merge,
-                                    total=progress._tasks[prog_id_merge].total + 1,
+                                    p_id_m,
+                                    total=progress._tasks[p_id_m].total + 1,
                                     refresh=True,
                                 )
                         else:  # Merge within process
@@ -680,8 +682,8 @@ class FuturesExecutor(ExecutorBase):
                                 accumulate(
                                     progress.track(
                                         map(_decompress, (c for c in batch)),
-                                        task_id=prog_id_merge,
-                                        total=progress._tasks[prog_id_merge].total
+                                        task_id=p_id_m,
+                                        total=progress._tasks[p_id_m].total
                                         + len(batch),
                                     ),
                                     _decompress(merged),
@@ -733,14 +735,14 @@ class FuturesExecutor(ExecutorBase):
                             for future in FH.completed
                             if is_good(future)
                         ]
-                        prog_id_merge = progress.add_task(
+                        p_id_m = progress.add_task(
                             "Merging finished jobs", unit="merges"
                         )
                         merged = _compress(
                             accumulate(
                                 progress.track(
                                     map(_decompress, (c for c in recovered)),
-                                    task_id=prog_id_merge,
+                                    task_id=p_id_m,
                                     total=len(recovered),
                                 )
                             ),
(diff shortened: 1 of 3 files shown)


lgray:
image

coffea/processor/executor.py
```diff
@@ -216,6 +216,42 @@ def __call__(self, items):
         return accumulate(items)
 
 
+class FuturesHolder:
+    def __init__(self, futures, refresh=2):
+        self.futures = set(futures)
+        self.merges = set()
+        self.completed = set()
```
nsmith-, line 246: Won't this hold a reference to all task futures until the processing has finished? In which case all the intermediate output stays alive longer than it needs to. This would be really bad for local processing.

coffea/processor/executor.py
```diff
@@ -553,23 +596,94 @@ def __call__(
...
-                        disable=not self.status,
-                        unit=self.unit,
-                        total=len(items),
-                        desc=self.desc,
-                    ),
-                    accumulator,
-                )
-            finally:
-                gen.close()
+                pbar = tqdm(disable=not self.status, unit=self.unit, total=len(items),
```
nsmith-: Add context managers for the progress bars (I should have done that before)

coffea/processor/executor.py
```diff
@@ -216,6 +216,42 @@ def __call__(self, items):
...
+                self.merges,
+                timeout=refresh,
+                return_when=concurrent.futures.FIRST_COMPLETED,
+            )
+            self.completed.update(completed)
+            self.done["merges"] += len(completed)
+
+    def fetch(self, N):
+        return [
+            self.completed.pop().result() for _ in range(N) if len(self.completed) > 0
```
nsmith-: `range(max(N, len(self.completed))`

nsmith- (commented):
Some early comments, mostly I'm concerned about object lifetime. One of the first problems we had was holding onto futures longer than necessary and blowing up our memory, as discussed in #97

coffea/processor/executor.py
andrzejnovak: that's there just to make the merging optional, nesting with with the conditional looks weird

coffea/processor/executor.py
andrzejnovak: Well technically with merging the intermediate/finished futures should get cleaned up. OTOH the merge futures get queue only after the standard jobs, not sure if there's a way around that. 

The future itself should add negligible overhead, not sure how bad we are doing with multiple hists vs. merged hist.

andrzejnovak: Ok, added intermediate merging on the submit node when merge jobs are turned off. Should be functionally equivalent to current implementation

nsmith-: Futures hold internally a ref to the result, which is large (histograms)
But if the goal here is to only hold `completed` that are pending a merge job and `merges` only holds finished merges pending a final merge, then I think that's ok. We might be in danger of having too deep a queue here though, but maybe lets wait and see if it becomes a problem

andrzejnovak:
The magic sauce seems to be a secondary executor (either spawned automatically or passed in) to run the merge jobs in parallel. 

I am not sure where we landed on having rich as a dependency, but rich progress bars make the code with multiple bars a bit neater https://github.com/andrzejnovak/coffea/pull/1/files

NJManganelli:
pip made rich a vendored dependency for 22.0, so was it transitively via that?

andrzejnovak:
Actually seems quite a bit faster this time around, two runs each, merge pool with 2 workers, which can probably run anywhere.
image
image

image
image

lgray:
Can you replicate the work to be done a little bit and see how the memory scales?

andrzejnovak:
image
 Same files though, so I am guessing the files are cached somehow so the time doesn't scale x10. Looking at `htop` while that happens, the mem consumption looks consistent throughout. I am running a slightly modded MyZPeak example with 4x60bins axes, so should be not entirely negligible.

lgray:
Be careful with tests in windows - forking and multiprocessing is strange there.

andrzejnovak:
Alright, I guess that makes it reviewable now. Specifically, I'd like to hear your thoughts on the added API and particularly for how the error state/processed items are returned back up by the executor(s). This is currently passed as a tuple, but it's not very elegant. I assume the reason the returns are formed like eg. `(accumulator, metrics)` as opposed to just returning the `wrapped_out` dict, is that somewhere it has to be immutable?

lgray:
We should test this on a many-core (64+) machine and check the scaling in extremum as well.

lgray:
We'll still run into some nasty issues with big histograms... hmm.

andrzejnovak:
> We should test this on a many-core machine and check the scaling in extremum as well.

I can test this on a node up to 80 cores, but there's a lot of RAM available, so it's not necessarily stressing the code.

The next scale up is obv the parsl executor (either with Futures or another parsl dfk doing the merging), which should maybe be a separate PR.

lgray:
Sure - the point is to observe trends rather than really stress anything.

andrzejnovak:
Alright, this crept up in scope a bit. 
- New features (for futures/parsl executors)
  - `recoverable` - Optionally return currently processed chunks along with the error in case of failure, instead of raising directly
  - `merging` - Implement merging in batches (either in main process/in main executor/in separate executor)
    - On small test jobs at least as fast as master or faster 
    - Will particularly improve when many small chunks are being processed or when many workers are available
  - rich progress bars
- API changes - breaking
  - convert sets to list in metrics (it's annoying they are not be default json dumpable)
  - Executors return `(out, 0)` or `(out, error)` if `recoverable` is implemented
- API changes
  - Calling Runner type standardized - instead of a variable-length tuple, return dict
    - isolated into Runner().run(...)
    - current `Runner()()` remain unchanged - under the hood calls `run()` 
    - deprecate and switch in the future?

Ready to review. We can expand the merge logic to dask and overall use of rich progress bars in future PRs.

coffea/processor/executor.py
```diff
@@ -1296,26 +1612,64 @@ def __call__(
...
+        ----------
+            fileset : dict
+                A dictionary ``{dataset: [file, file], }``
+                Optionally, if some files' tree name differ, the dictionary can be specified:
+                ``{dataset: {'treename': 'name', 'files': [file, file]}, }``
+            treename : str
+                name of tree inside each root file, can be ``None``;
+                treename can also be defined in fileset, which will override the passed treename
+            processor_instance : ProcessorABC
+                An instance of a class deriving from ProcessorABC
```
andrzejnovak, line 1653: fix docstring

coffea/processor/executor.py
```diff
@@ -208,7 +215,7 @@ def __str__(self):
         return "reduce"
 
     def __call__(self, items):
-        items = list(items)
+        items = list(it for it in items if it is not None)
```
nsmith-, line 232: Now that this list is filtered, is there a case where `len(items)==0`? Since it would trigger the exception immediately following

coffea/processor/executor.py
```diff
@@ -555,30 +623,152 @@ def __call__(
...
+        def processwith(pool, mergepool):
+            reducer = _reduce(self.compression)
+            FH = FuturesHolder(
+                set(pool.submit(function, item) for item in items), refresh=2
             )
+
+            nparts, minred, maxred = 5, 4, 100
+            if isinstance(self.merging, tuple):
+                if len(self.merging) == 3:
+                    nparts, minred, maxred = self.merging
```
nsmith-: Move above `pool.submit` and return an error if the format isn't valid

nsmith-: Best would be a dataclass `__post_init__` hook https://docs.python.org/3/library/dataclasses.html#post-init-processing

coffea/processor/executor.py
```diff
@@ -555,30 +623,152 @@ def __call__(
...
+                    merged = None
+                    while len(FH.futures) + len(FH.merges) > 0:
+                        FH.update()
+                        merge_size = min(
+                            maxred, max(len(FH.completed) // nparts + 1, minred)
+                        )
+                        progress.update(
+                            p_id,
+                            advance=FH.done["futures"]
+                            - progress._tasks[p_id].completed,
```
nsmith-: Is there a public method on the rich progress bar?

coffea/processor/executor.py
```diff
@@ -218,6 +225,41 @@ def __call__(self, items):
...
+                self.merges,
+                timeout=refresh,
+                return_when=concurrent.futures.FIRST_COMPLETED,
+            )
+            self.completed.update(completed)
+            self.done["merges"] += len(completed)
+
+    def fetch(self, N):
+        return [
+            self.completed.pop().result() for _ in range(min(N, len(self.completed)))
```
nsmith-: Since a future's exception would be raised from here (`f.result()`) then any other completed futures that were popped before getting to the broken one would get dropped from the recovery process, right? Can we let `fetch` return the future and later access its result, perhaps after an explicit check if it has an `f.exception()`?

coffea/processor/executor.py
```diff
@@ -817,28 +1029,143 @@ def __call__(
...
-                    disable=not self.status,
-                    unit=self.unit,
-                    total=len(items),
-                    desc=self.desc,
-                ),
-                accumulator,
-            )
-        finally:
-            gen.close()
+            with rich_bar() as progress:
```
nsmith-: How much of this section is different from the `FuturesExecutor` impl? It looks quite similar, can we deduplicate? (maybe by putting the future update loop into FuturesHolder and passing it a callback to submit the processing and reducing task?)

coffea/processor/executor.py
```diff
@@ -817,28 +1029,143 @@ def __call__(
...
+                                total=len(recovered),
+                            )
+                        ),
+                        self.compression,
+                    )
 
-        return accumulator
+                    return accumulate([_decompress(merged), accumulator]), e
+            else:
+                raise type(e)(str(e)).with_traceback(sys.exc_info()[2]) from None
```
nsmith-: Based on #650 this should just be `raise Exception(...) from e`

coffea/processor/executor.py
```diff
@@ -901,20 +1228,14 @@ class Runner:
...
+    cachestrategy: Optional[
+        Union[Literal["dask-worker"], Callable[..., MutableMapping]]
+    ] = None  # fmt: skip
     processor_compression: int = 1
     use_skyhook: Optional[bool] = False
+    skyhook_options: Optional[Dict] = field(default_factory=dict)
     format: str = "root"
 
-    @staticmethod
-    def read_coffea_config():
```
nsmith-: Why is this deleted?

tests/test_processor.py
```diff
@@ -117,7 +117,7 @@ def postprocess(self, accumulator):
         proc.process,
         None,
     )
-    assert out == {"itemsum": 45}
+    assert out == ({"itemsum": 45}, 0)
```
nsmith-, line 120: Based on the fact that `__call__` preserves the old behavior, why did this end up changing?

nsmith- (changes requested):
As an overall comment, if you can sprinkle some typing in some of the new function signatures that would be helpful.

tests/test_processor.py
andrzejnovak: Runner.__call__ preservers old behaviour, this is invoking the exec directly

coffea/processor/executor.py
andrzejnovak: Hmm, something weird happened when I updated the branch

coffea/processor/executor.py
andrzejnovak: 👍

coffea/processor/executor.py
andrzejnovak: refactored

coffea/processor/executor.py
andrzejnovak: there isn't but there's a different way to advance the progress

coffea/processor/executor.py
andrzejnovak: I suppose not, but if `len(items)==0` then the original would throw just below anyway no? This is in place if a default new accumulator is being added to the merge, but can just be initiated as None

andrzejnovak:
Instead of passing "prepro" to run, I tried to factor it out into a separate fcn, which makes more sense, but having to comply with the dynamic chunking there makes it a bit awkward (relies on both filemeta and chunk generator, so `run` needs to be able to take `fileset` as input.

andrzejnovak:
@nsmith- Cleaned up the commits and also added rich.progress for Iterative. I'd say this is ready

lgray:
Since we can't capture it very well in CI have we tried this PR at scale with all the executors?

andrzejnovak:
> Since we can't capture it very well in CI have we tried this PR at scale with all the executors?

Reasonable scale for futures, full 2016 production of my analysis with parsl about 3 times. The speed there is dependent on how busy the disk, but in each it was faster or as fast as master

coffea/processor/executor.py
```diff
@@ -1315,7 +1622,47 @@ def __call__(
             # v0.7.4. This fixes tests using maxchunks.
             fileset.reverse()
 
-        chunks = self._chunk_generator(fileset, treename)
+        return self._chunk_generator(fileset, treename)
+
+    def run(
+        self,
+        fileset: Union[Dict, List, str],
```
nsmith-: Generator now too?

coffea/processor/executor.py
```diff
@@ -525,25 +695,65 @@ class FuturesExecutor(ExecutorBase):
...
+            Callable[..., concurrent.futures.Executor],
+            concurrent.futures.Executor,
+            bool,
+        ]
+    ] = None
+    recoverable: bool = False
+    merging: Union[bool, Tuple[int, int, int]] = False
+    nparts: int = 5
+    minred: int = 4
+    maxred: int = 100
```
nsmith-: I guess I was thinking to still have the merging as one arg. This is a bit of duplication of parameters but probably ok since its validated after construction

lgray:
@nsmith-  you happy here?

nsmith- (approved):
Just a small type thing, not making it required but would be nice.

coffea/processor/executor.py
andrzejnovak: seems requisite to making the parsing in the _post_init

andrzejnovak:
@nsmith- seems we're good to go

PR #647, opened 2022-03-08
Set default xrootdtimeout to 60s

lgray:


nsmith-:
Of course now that I set this I get `OSError: XRootD error: [ERROR] Operation expired` all the time now. I guess we should let FuturesExecutor perform retries. I'm actually not sure why it doesn't. Looks like a small change

PR #648, opened 2022-03-09
Read Coffea config from TOML file

JayjeetAtGithub:


JayjeetAtGithub:
@oshadura Here I have created a concept PR to create a global coffea config file which contains skyhook parameters. The coffea config file is located in the `$HOME/.coffea.toml`.  We read this config file inside the `Runner` and read out skyhook config out of it. The user just needs to define `use_skyhook=True` now. Please let me know if this is in the right direction to make Coffea/Skyhook user-friendly.

lgray:
Please do not make this configuration information for skyhook a requirement for every single executor. That's a big change for our users.

JayjeetAtGithub:
@lgray Thanks a lot for looking into this. For the very same reason, I have kept the `.coffea.toml` file as optional. If there exists a .coffea.toml, then it will be read, otherwise it is not required. This config file currently serves only skyhook specific config, but serves as a placeholder to add other backend specific config in the future.

lgray:
@JayjeetAtGithub are you done with this PR?

JayjeetAtGithub:
@lgray Yes, I am done

Issue #438, opened 2021-02-02
skipbadfiles to pass AuthError through


andrzejnovak:
`skipbadfiles=True` (often defaulted to) when the files are inaccessible due to Auth will not display any errors and return empty accumulator, which is very intransparent for new users to debug. Not ignoring auth error should not affect functionality

Issue #370, opened 2020-11-20
File/entries error annotation printed twice


andrzejnovak:
@pfackeldey There is some duplication in the error message after your changes

<details>

```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/afs/desy.de/user/a/anovak/dust/miniconda3/lib/python3.7/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/afs/desy.de/user/a/anovak/coffea/coffea/processor/executor.py", line 129, in __call__
    out = self.function(*args, **kwargs)
  File "/afs/desy.de/user/a/anovak/coffea/coffea/processor/executor.py", line 877, in _work_function
    raise e
  File "/afs/desy.de/user/a/anovak/coffea/coffea/processor/executor.py", line 836, in _work_function
    raise Exception(f"Failed processing file: {item.filename} ({item.entrystart}-{item.entrystop})") from e
Exception: Failed processing file: root://xrootd-cms.infn.it///store/user/anovak/PFNano/106X_vA01/TTToSemiLeptonic_TuneCP5_13TeV-powheg-pythia8/RunIIFall17PFNanoAODv2-PU2017_12Apr2018_new_pmx_94X_mc2017_realistic_v14-v1/201118_010348/0000/nano106X_on_mini106X_2017_mc_NANO_10.root (0-259880)
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "runner.py", line 99, in <module>
    chunksize=args.chunk, maxchunks=args.max
  File "/afs/desy.de/user/a/anovak/coffea/coffea/processor/executor.py", line 1130, in run_uproot_job
    executor(chunks, closure, wrapped_out, **exe_args)
  File "/afs/desy.de/user/a/anovak/coffea/coffea/processor/executor.py", line 559, in futures_executor
    _futures_handler(futures, accumulator, status, unit, desc, add_fn, tailtimeout)
  File "/afs/desy.de/user/a/anovak/coffea/coffea/processor/executor.py", line 187, in _futures_handler
    add_fn(output, finished.pop().result())
  File "/afs/desy.de/user/a/anovak/dust/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 428, in result
    return self.__get_result()
[...]

PR #650, opened 2022-03-10
fix Processor.process exception handling

### LINKED ISSUE #649: [Bug] Exception raising `processor_instance.process(events)` fails for custom Exception signatures
Dear coffea developer,

The following lines fail in case the catched Exception has a custom signature: https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L1234-L1240

This happened recently to us, where it catched a `DeserializationError` from uproot (see: https://github.com/scikit-hep/uproot4/blob/22407793f9eab02ce73d6843bb3395bedacab50b/src/uproot/deserialization.py#L322-L347)

Some time ago I have opened a PR to avoid this: https://github.com/CoffeaTeam/coffea/pull/360
However this resulted (for whatever reason, which I still do not really understand...) in the double printing of the error message: https://github.com/CoffeaTeam/coffea/issues/370 and thus the Exception handling was reverted to its old behaviour in https://github.com/CoffeaTeam/coffea/pull/439.

I think it is a bug when coffea completely hides Exceptions as the `type(e)(...)` call fails and thus does not even output the catched exception at all. On the other hand I understand that double printing of error messages is also annoying.

What do you think about this @lgray @nsmith- @andrzejnovak (tagging here the relevant people from the last issue and PR)?

Best, Peter

bfis:
Fixes the exception handling behavior for the Processor.process call by correctly chaining exceptions.

Reintroduces the benefits of #360.
Reintroduces the behavior of #370, since this is working as intended in python's `concurrent/futures`.
Retrains the fix for #438.
Fixes #649.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -20,7 +20,7 @@ from .processor import ProcessorABC
 from .accumulator import accumulate, set_accumulator, Accumulatable
 from .dataframe import LazyDataFrame
 from ..nanoevents import NanoEventsFactory, schemas
-from ..util import _hash
+from ..util import _hash, _exception_chain
 
 from collections.abc import Mapping, MutableMapping
 from dataclasses import dataclass, field, asdict
@@ -979,12 +979,13 @@ class Runner:
             # catch xrootd errors and optionally skip
             # or retry to read the file
             except Exception as e:
-                if skipbadfiles and isinstance(e, FileNotFoundError):
+                chain = _exception_chain(e)
+                if skipbadfiles and any(isinstance(c, FileNotFoundError) for c in chain):
                     warnings.warn(str(e))
                     break
                 if (
                     not skipbadfiles
-                    or "Auth failed" in str(e)
+                    or any("Auth failed" in str(c) for c in chain)
                     or retries == retry_count
                 ):
                     raise e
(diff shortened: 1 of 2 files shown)


nsmith-:
So the key is to add a new exception to the chain rather than attempt to modify the last exception's message?
We're approaching java levels of traceback length sadly, but it is too hazardous to try to edit them I think.

andrzejnovak:
Filtering the traceback for a duplicate substring over some length would be too fragile?

bfis:
> We're approaching java levels of traceback length sadly, [...].

Yes, surprisingly, there is no tooling/help from python's side with handling/catching exceptions that are down the chain - it seems to me the original [PEP-3134](https://peps.python.org/pep-3134/) did not concern itself too much with the handling of exception chains it introduced.


> Filtering the traceback for a duplicate substring over some length would be too fragile?

Messing around with exception in such a away is a recipe for disaster since they have no fixed format and can do just about anything (e.g. tracebacks can be formatted very differently etc.). The underlying "duplication"/formatting issue lies in the internals of python`s [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor), so should be remedied there.

PR #657, opened 2022-04-05
wq: adds log of accumulated chunks

btovar:
As chunks get accumulated, they are written to a log amenable to pandas
processing. The log has information about the resources used, the
dataset involved, and the source file.

The current motivation is to debug processors in topcoffea for which
certain datasets/filenames seem to be taking much longer than others.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -553,6 +553,8 @@ class WorkQueueExecutor(ExecutorBase):
             Filename for tasks statistics output
         transactions_log : str
             Filename for tasks lifetime reports output
+        tasks_accum_log : str
+            Filename for the log of tasks that have been processed and accumulated.
         print_stdout : bool
             If true (default), print the standard output of work queue task on completion.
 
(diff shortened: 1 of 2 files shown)


nsmith-:
We have the existing `metrics` accumulatable returned by the `_work_function`, do you think that could be amended to return the CPU info you are interested in?
Alternatively, as part of #646, some plumbing was added to optionally return a list of processed `WorkItem` objects, for future use with a partial re-try mechanism. I wonder if that could be leveraged to track stats per WorkItem?

btovar:
@nsmith-  One important difference is that this log is created as the workflow is executed and it applies to preprocessing, processing and accumulating tasks (not just processing as with _work_function). This is important for our use case as we do not have to wait for all the chunks to be processed and accumulated before figuring out something is not as it should be. (E.g., an IO bottleneck vs a particular dataset that is taking much longer than others for the processor being developed, or accumulating tasks using much more memory than expected.) This is mostly a tool to help fix buggy processor while they are being developed.

I think in theory you could reconstruct failed WorkItem's, as they would be listed as processing tasks together with their dataset name, filename, and event range. However, that would be only for work items that were actually generated. Since WorkItem's are created on demand using a generator, then we would either have to look at which event ranges are missing from a preprocessed file, or we would have to generate all the WorkItem's even if the run hits some exception.

PR #660, opened 2022-04-14
Ensure behavior sticks to vector functions

nsmith-:
By adding an explicit behavior to all `ak.zip` calls. For most, it would have been implied but this is a bit safer


nsmith-:
Motivated by the fact that before this change, `events.Jet[0].sum().mass` did not work out of the box because the sum lost its behavior.

PR #663, opened 2022-04-20
Return scalar for massless particles if Record type

### LINKED ISSUE #640: Bug in nanoevents.methods.nanoaod.Photon
Here is a snippet that describes the problem: it uses a nanoAOD that I have locally but it can be reproduced with any nanoAOD.

```python
In [20]: from coffea import nanoevents

In [21]:

In [21]: events = nanoevents.NanoEventsFactory.from_root('/pnfs/psi.ch/cms/trivcat/store/user/gallim/HggNano_UL17_DY-TnPCompare/DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8/crab_HggNA
    ...: NO_UL17_DY-TnPCompare/220202_072553/0000/DYJetsToLL--UL17_v6-v2_1.root').events()

In [22]:

In [22]: photons = events.Photon

In [23]: ak.num(photons)
Out[23]: <Array [2, 2, 0, 2, 0, 0, ... 0, 2, 1, 1, 0, 0] type='63795 * int64'>

In [24]: # accessing the first photon of the first event

In [25]: photons[0][0].mass
TypeError                                 Traceback (most recent call last)
<ipython-input-25-5d5b0fca44dc> in <module>
----> 1 photons[0][0].mass

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/coffea/nanoevents/methods/nanoaod.py in mass(self)
    216     @property
    217     def mass(self):
--> 218         return awkward.without_parameters(awkward.zeros_like(self.pt))
    219
    220     @property

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/awkward/operations/structure.py in zeros_like(array, highlevel, behavior, dtype)
   1062     if dtype is not None:
   1063         return full_like(array, 0, highlevel=highlevel, behavior=behavior, dtype=dtype)
-> 1064     return full_like(array, _ZEROS, highlevel=highlevel, behavior=behavior, dtype=dtype)
   1065
   1066

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/awkward/operations/structure.py in full_like(array, fill_value, highlevel, behavior, dtype)
   1160             fill_value = int(fill_value)
   1161
-> 1162     layout = ak.operations.convert.to_layout(
   1163         array, allow_record=True, allow_other=False
   1164     )

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/awkward/operations/convert.py in to_layout(array, allow_record, allow_other, numpytype)
   1936
   1937     elif not allow_other:
-> 1938         raise TypeError(
   1939             "{0} cannot be converted into an Awkward Array".format(array)
   1940             + ak._util.exception_suffix(__file__)

TypeError: 47.68001174926758 cannot be converted into an Awkward Array

(https://github.com/scikit-hep/awkward-1.0/blob/1.7.0/src/awkward/operations/convert.py#L1940)
```

I think the reason why this happens is that ```awkward.zeros_like``` does not have the same behaviour of ```numpy.zeros_like``` when a float is passed:

```python
In [26]: import numpy as np

In [27]: import awkward as ak

In [28]: np.zeros_like(1.0)
Out[28]: array(0.)

In [29]: ak.zeros_like(1.0)
TypeError                                 Traceback (most recent call last)
<ipython-input-29-95cb45da8df1> in <module>
----> 1 ak.zeros_like(1.0)

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/awkward/operations/structure.py in zeros_like(array, highlevel, behavior, dtype)
   1062     if dtype is not None:
   1063         return full_like(array, 0, highlevel=highlevel, behavior=behavior, dtype=dtype)
-> 1064     return full_like(array, _ZEROS, highlevel=highlevel, behavior=behavior, dtype=dtype)
   1065
   1066

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/awkward/operations/structure.py in full_like(array, fill_value, highlevel, behavior, dtype)
   1160             fill_value = int(fill_value)
   1161
-> 1162     layout = ak.operations.convert.to_layout(
   1163         array, allow_record=True, allow_other=False
   1164     )

/work/gallim/anaconda3/envs/higgs-dna/lib/python3.9/site-packages/awkward/operations/convert.py in to_layout(array, allow_record, allow_other, numpytype)
   1936
   1937     elif not allow_other:
-> 1938         raise TypeError(
   1939             "{0} cannot be converted into an Awkward Array".format(array)
   1940             + ak._util.exception_suffix(__file__)

TypeError: 1.0 cannot be converted into an Awkward Array

(https://github.com/scikit-hep/awkward-1.0/blob/1.7.0/src/awkward/operations/convert.py#L1940)
```

Should I maybe open an issue for this in Awkward? The docs say that ```ak.zeros_like``` is the equivalent of ```np.zeros_like```, but from this example it doesn't seem to be exactly the case.

nsmith-:
Fixes #640

### DIFF
--- a/coffea/nanoevents/methods/delphes.py
+++ b/coffea/nanoevents/methods/delphes.py
@@ -182,7 +182,9 @@ _set_repr_name("Particle")
 class MasslessParticle(Particle, base.NanoCollection):
     @property
     def mass(self):
-        return awkward.without_parameters(awkward.zeros_like(self.pt))
+        if hasattr(self.pt, "__len__"):
+            return awkward.without_parameters(awkward.zeros_like(self.pt))
+        return 0.0
 
 
 _set_repr_name("MasslessParticle")
--- a/coffea/nanoevents/methods/delphes.py
+++ b/coffea/nanoevents/methods/delphes.py
@@ -182,9 +182,7 @@ _set_repr_name("Particle")
 class MasslessParticle(Particle, base.NanoCollection):
     @property
     def mass(self):
-        if hasattr(self.pt, "__len__"):
-            return awkward.without_parameters(awkward.zeros_like(self.pt))
-        return 0.0
+        return 0.0 * self.pt
 
 
 _set_repr_name("MasslessParticle")
--- a/coffea/nanoevents/methods/nanoaod.py
+++ b/coffea/nanoevents/methods/nanoaod.py
@@ -215,9 +215,7 @@ class Photon(candidate.PtEtaPhiMCandidate, base.NanoCollection, base.Systematic)
 
     @property
     def mass(self):
-        if hasattr(self.pt, "__len__"):
-            return awkward.without_parameters(awkward.zeros_like(self.pt))
-        return 0.0
+        return 0.0 * self.pt
 
     @property
     def isLoose(self):
--- a/coffea/nanoevents/methods/nanoaod.py
+++ b/coffea/nanoevents/methods/nanoaod.py
@@ -215,7 +215,9 @@ class Photon(candidate.PtEtaPhiMCandidate, base.NanoCollection, base.Systematic)
 
     @property
     def mass(self):
-        return awkward.without_parameters(awkward.zeros_like(self.pt))
+        if hasattr(self.pt, "__len__"):
+            return awkward.without_parameters(awkward.zeros_like(self.pt))
+        return 0.0
 
     @property
     def isLoose(self):


nsmith-:
@jpivarski is this general enough for long-term, given the `wont-fix` for https://github.com/scikit-hep/awkward-1.0/issues/1278?

jpivarski:
Well, Awkward v1 returns a Python scalar, like the `0.0` here, but v2 returns whatever the library returns, which is `np.float64(0.0)` from NumPy and may be a zero-dimensional array from other libraries, such as CuPy and JAX. (The "backend" is a user-visible attribute of an ak._v2.Array.) CuPy has the very good reasoning that it should remain an array so that a scalar can be used as part of a calculation without forcing a GPU → CPU → GPU copy in the middle of the calculation. Even though the scalar is small, it synchronizes what would otherwise be an asynchronous calculation. That's the motivation for letting the backend library do its thing, whatever that is.

NumPy scalars don't have a `"__len__"`, but they do have a `"shape"`, so maybe that's a better thing to check for? Or to be even more hands-off, multiply whatever `pt` is by `0.0`? Multiplication by a Python scalar drops parameters and keeps the type of NumPy scalars, CuPy zero-dimensional arrays, and JAX zero-dimensional arrays, so all the types that come out of a v1 or a v2 Awkward Array would be maintained.

nsmith-:
Ok, especially in view of backends like jax that will be tracing, it seems best to just multiply by zero.

PR #681, opened 2022-06-02
Some lookup fixes

slehti:
Two problems
1: get mu (pileup) for data. For MC one can use events.Pileup.nTrueInt, but that is not possible for data. For data the pileup have to be taken from the pileup json, like 
/afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions16/13TeV/PileUp/UltraLegacy/pileup_latest.txt
/afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions17/13TeV/PileUp/UltraLegacy/pileup_latest.txt
/afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions18/13TeV/PileUp/UltraLegacy/pileup_latest.txt
as a function of run and lumisection. Added support for lookup from pileup jsons to be able to get the mu for data.
2: The jet resolution scale factor files from the JRDatabase are not working to get the scale factor, as the code looks for *.jersf.* in the filename. The JER SF files in the JRDatabase are named as *_SF_*, so made a change in the extractor.py function import_file to look for _SF_ and if found set thetype as 'jersf'. 
For some reason thetype = "default" is not working for these files, although it is working for the other JEC files. Either the thetype = 'default' should be made working for JER SF's, or use this fix looking for the filenames.

### DIFF
--- a/coffea/lookup_tools/json_lookup.py
+++ b/coffea/lookup_tools/json_lookup.py
@@ -1,0 +1,14 @@
+class json_lookup:
+    def __init__(self, wrapped_values):
+        self.values = wrapped_values
+
+    def __call__(self,run,lumi):
+        mu = []
+        for i in range(len(run)):
+            run_ = str(run[i])
+            lumi_ = lumi[i]
+            if not run_ in self.values.keys() or not lumi_ in self.values[run_].keys():
+                print("    \033[91m run:lumi %s:%s not found \033[00m"%(run_,lumi_))
+            else:
+                mu.append(self.values[run_][lumi_])
+        return mu
--- a/coffea/lookup_tools/json_lookup.py
+++ b/coffea/lookup_tools/json_lookup.py
@@ -7,8 +7,8 @@ class json_lookup:
         for i in range(len(run)):
             run_ = str(run[i])
             lumi_ = lumi[i]
-            if not run_ in self.values.keys() or not lumi_ in self.values[run_].keys():
-                print("    \033[91m run:lumi %s:%s not found \033[00m"%(run_, lumi_))
+            if run_ not in self.values.keys() or lumi_ not in self.values[run_].keys():
+                print("    \033[91m run:lumi %s:%s not found \033[00m" % (run_, lumi_))
             else:
                 mu.append(self.values[run_][lumi_])
         return mu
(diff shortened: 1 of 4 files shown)


lgray:
@slehti you need to fix linting / code format errors.

slehti:
Fixed format errors.

lgray:
@slehti There are still a few errors from flake8 rules (and possibly black too).
Please run flake8 and black to make sure your code will pass this.
You can run it yourself by doing the following:
```bash
cd coffea
pip install flake8 black
black tests setup.py coffea
flake8 tests setup.py coffea
```

lgray:
So the reason the parser doesn't work for the jersf files is because their structure and meaning is actually different from the rest of the JEC inputs. I personally think it is better to make this apparent than to automagically deal with it, since that implies you stop to think about it at least once (which reduces error rate in the long run).

lgray:
Could you please add a test for `json_lookup`, since undoubtedly other people will use it and so we must ensure its consistent functioning.

lgray:
Ah - now reading through this a bit more.

Here's the little philosophy behind the extra file postfixes.
`_SF_` or `pileup` can very easily appear in the names of files in one way or another. This makes searching for these markers in the bulk of the filename *extremely* unreliable in the general case since the same word can have many different meanings depending on context and that context may not be clearly communicated in all cases.

If you turn that meaning into part of the file's postfix like `ZZZZZ.jersf.txt` or `YYYYY.pileup.json` then the purpose becomes consistently clear and there is never any guess work on the part of any maintainer or user.

lgray:
ping @slehti

lgray:
I'll let it in for now - if this causes any problems you'll get pinged.

coffea/lookup_tools/json_lookup.py
```diff
@@ -0,0 +1,14 @@
+class json_lookup:
+    def __init__(self, wrapped_values):
+        self.values = wrapped_values
+
+    def __call__(self, run, lumi):
+        mu = []
+        for i in range(len(run)):
```
nsmith-, line 7: Do we really want an interpreted loop here?

nsmith-:
This pileup json format would be a very good candidate to convert to correctionlib, it would evaluate much faster there.

lgray:
@slehti #727 is why I protested against your changes. `_SF_` is far too common. I'll make a fix, the real solution is correctionlib + json or using additional file postfixes.

PR #684, opened 2022-06-10
Added `E` alias for 4-vectors

ssrothman:
The fastjet library (https://github.com/scikit-hep/fastjet) expects inputs with an attribute named "E" with the particle energy. This small PR adds this alias to the PtEtaPhiM and PtEtaPhiE Lorentz Vectors to make coffea compatible with fastjet.


lgray:
@ssrothman please format this code with black!

PR #685, opened 2022-06-20
Adding file sizes to work queue tasks log

andrewhennessee:
This update to `work_queue_tools.py` in `coffea/processor` includes the input and output file sizes of tasks in the task accumulation log (tasks.log). Output file sizes are measured using the `getsize()` method from the `os.path` library. Input sizes of accumulating tasks are measured by summing the output sizes of the tasks to be accumulated. Currently, input sizes of preprocessing and processing tasks are set to zero (not measured).

### DIFF
--- a/coffea/processor/work_queue_tools.py
+++ b/coffea/processor/work_queue_tools.py
@@ -6,7 +6,7 @@ import tempfile
 import textwrap
 import signal
 
-from os.path import basename, join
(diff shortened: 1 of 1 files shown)


lgray:
Do you need to work any more on this PR or is it done?

andrewhennessee:
@lgray I'm done working on it unless you need me to make a change. Thanks!

PR #688, opened 2022-06-22
do not use uproot 4.2.4

lgray:


lgray:
better covered in #689

Issue #686, opened 2022-06-21
nanoevents uproot mapping in coffea v0.7.15 needs update for changes in uproot v4.2.4  (.awkward_form method)


btovar:
**Describe the bug**

As seen in uproot issue https://github.com/scikit-hep/uproot5/issues/614, the arguments of the `.awkward_form` method used in https://github.com/CoffeaTeam/coffea/blob/532c69afc8f837c6886542727d493b274d2e2c78/coffea/nanoevents/mapping/uproot.py#L94 have changed.   The method now takes a second required argument `context`. To preserve the current default behavior, this argument should be set to:

```python
{"index_format": "i64", "header": False, "tobject_header": True, "breadcrumbs": ()}
```

I have not checked if this method is used somewhere else in coffea. I can make this pr if you want.

jpivarski:
Well, actually, that's our fault; Uproot will change back. The method didn't have a leading underscore, so it's public and we (I) shouldn't be changing the interface without warning like that.

@aryan26roy is making the `Interpretation.awkward_form` methods accept the same arguments that they used to (which then immediately create the new `context` and pass that down through a bunch of internal `awkward_form` functions, separate from the `Interpretation.awkward_form`). Coffea will not need to change.

This will be deployed as Uproot 4.3.0, first in a line of bug-fix releases for Uproot 4, tagged from the new `main-v4` branch. It will also be a test of this new v4 bug-fix workflow.

jpivarski:
See https://github.com/scikit-hep/uproot5/issues/614#issuecomment-1162434040.

PR #689, opened 2022-06-22
Avoid uproot release with API change

### LINKED ISSUE #687: Pin some dependencies to specific versions
Hi!

This "issue" is something between "feature request", "bug report" and a suggestion :)

When installing coffea and its requirements, I've noticed that some version checks are rather loose, e.g. here:

https://github.com/CoffeaTeam/coffea/blob/532c69afc8f837c6886542727d493b274d2e2c78/setup.py#L60

This can cause situations where updates of upstream dependencies, that changed parts of their API, are propagated to the user. This happens, for instance, at the moment:

Signatures of some core functions in `uproot` were changed between patch releases ([from 4.2.3 to 4.2.4](https://github.com/scikit-hep/uproot5/commit/04a775267574ab1687df7a03a2bb1e2fb2790aa4#diff-bf61590d05b6dde9ae4aab8ffeccacf52c8d5379997354ebe59cca60df2d12b3R409), released yesterday), leading to

```python
TypeError: awkward_form() missing 1 required positional argument: 'context'
```

being raised here:

https://github.com/CoffeaTeam/coffea/blob/532c69afc8f837c6886542727d493b274d2e2c78/coffea/nanoevents/mapping/uproot.py#L94

It think in most situations it should be sufficient to pin dependencies to a certain minor release (e.g. `scipy==1.1.*`), or, if necessary, to patch releases for packages not strictly following semantic versioning where these API changes are known to happen from time to time (e.g. `uproot==4.2.3`). I think this would help to be more robust against breaking changes in upstream deps.

Fyi: I'm trying to create a PR in the meantime to address the above mentioned issue, but I might need some time to dive in :)

nsmith-:
See #686
Closes #687

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ def get_description():
 
 INSTALL_REQUIRES = [
     "awkward>=1.5.1,<2",
-    "uproot>=4.1.6",
+    "uproot>=4.1.6,==4.*,!=4.2.4",
     "uproot3-methods>=0.10.0",
     "uproot3>=3.14.1",
     "correctionlib>=2.0.0",
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ def get_description():
 
 INSTALL_REQUIRES = [
     "awkward>=1.5.1,<2",
-    "uproot>=4.1.6,==4.*,!=4.2.4",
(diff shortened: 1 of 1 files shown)


nsmith-:
uproot5 will be rc-only for quite a while but just to be safe we put an upper limit

PR #691, opened 2022-06-24
Make vertex position explicit

nsmith-:
@andrzejnovak thoughts?

### DIFF
--- a/coffea/nanoevents/methods/nanoaod.py
+++ b/coffea/nanoevents/methods/nanoaod.py
@@ -365,10 +365,21 @@ _set_repr_name("MissingET")
 
 
 @awkward.mixin_class(behavior)
-class Vertex(vector.ThreeVector, base.NanoCollection):
+class Vertex(base.NanoCollection):
     """NanoAOD vertex object"""
 
-    pass
+    @property
+    def pos(self):
+        """Vertex position as a three vector"""
+        return awkward.zip(
+            {
+                "x": self["x"],
+                "y": self["y"],
+                "z": self["z"],
+            },
+            with_name="ThreeVector",
+            behavior=self.behavior,
+        )
 
 
 _set_repr_name("Vertex")


andrzejnovak:
Do I understand this right? With this change the 4-vector props for adding/dR/nearest get read from pt/phi/eta/mass, while the x/y/z info is accessible via the newly added `.pos`? I think that's good.

Hard to say if anyone else was using this, but probably should be marked as breaking

nsmith-:
Rather, there are no `.nearest`, `.deltaR`, etc. for the base object but you have to use explicitly `.pos` or `.p4`

andrzejnovak:
So `ak.first(events.Jet).nearest(events.SV)` will throw an error?

nsmith-:
Hmm I guess not because the base object has all the relevant quantities:
```
>>> ak.firsts(events.Jet).nearest(events.SV)
<SecondaryVertexArray [[None], [None], ... [None], [None]] type='40 * option[var...'>
>>> ak.firsts(events.Jet).nearest(events.SV.pos)
  File "/Users/ncsmith/src/coffea/coffea/nanoevents/methods/vector.py", line 709, in nearest
    mval, (a, b) = self.metric_table(other, axis, metric, return_combinations=True)
  File "/Users/ncsmith/src/coffea/coffea/nanoevents/methods/vector.py", line 675, in metric_table
    mval = metric(a, b)
  File "/Users/ncsmith/src/coffea/coffea/nanoevents/methods/vector.py", line 684, in <lambda>
    metric=lambda a, b: a.delta_r(b),
  File "/Users/ncsmith/src/coffea/coffea/nanoevents/methods/vector.py", line 585, in delta_r
    return numpy.hypot(self.eta - other.eta, self.delta_phi(other))
  File "/Users/ncsmith/src/coffea/.env/lib/python3.9/site-packages/awkward/highlevel.py", line 1123, in __getattr__
    raise AttributeError(
AttributeError: no field named 'eta'

(https://github.com/scikit-hep/awkward-1.0/blob/1.8.0/src/awkward/highlevel.py#L1124)
>>> ak.firsts(events.Jet).nearest(events.SV.p4)
<PtEtaPhiMLorentzVectorArray [[None], [None], ... [None], [None]] type='40 * opt...'>
```
I think I would have to adjust the form to remove those fields from the base object if we wanted it to throw, since `eta` and `phi` are real data members of SV. On the other hand, it at least gives the correct result since those are what are used to make `.p4`

andrzejnovak:
Imo that's fine. It could be nice to force explicitness when both are available, but i don't think anyone actually wants the xyz pos, so as proposed is not gonna be a surprise

Issue #690, opened 2022-06-23
Missing $HOME raises exception


DryRun:
**Describe the bug**
If `$HOME` doesn't exist (which is the case for condor jobs on cmslpc), an exception is raised at https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L1203. 

```
/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/util.py:154: FutureWarning: In coffea version v0.8.0 (target date: 31 Dec 2022), this will be an error.
(Set coffea.deprecations_as_errors = True to get a stack trace now.)
ImportError: coffea.hist is deprecated
  warnings.warn(message, FutureWarning)
Traceback (most recent call last):
  File "runFullDataset.py", line 38, in <module>
    maxchunks          = args.maxchunks
  File "/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/processor/__init__.py", line 107, in _run_x_job
    processor_instance=processor_instance,
  File "/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/processor/executor.py", line 1591, in __call__
    wrapped_out = self.run(fileset, processor_instance, treename)
  File "/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/processor/executor.py", line 1715, in run
    chunks = [c for c in chunks]
  File "/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/processor/executor.py", line 1715, in <listcomp>
    chunks = [c for c in chunks]
  File "/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/processor/executor.py", line 1403, in _chunk_generator
    config = Runner.read_coffea_config()
  File "/pool/condor/dir_13517/ttgenv/lib/python3.6/site-packages/coffea/processor/executor.py", line 1203, in read_coffea_config
    config_path = os.path.join(os.environ["HOME"], ".coffea.toml")
  File "/cvmfs/sft.cern.ch/lcg/views/LCG_96python3/x86_64-centos7-gcc8-opt/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'HOME'
```

The exception comes from this function:
```
    def read_coffea_config():
        config_path = os.path.join(os.environ["HOME"], ".coffea.toml")
```

**To Reproduce**
[...]

PR #692, opened 2022-07-01
Fix more skyhook

lgray:
Fixed more instances of the pyarrow dataset interface being assumed. Deal with instances of environment being assumed.


lgray:
Fixes #690

PR #698, opened 2022-07-16
Add deep children lookup to GenParticle

riga:
Hi!

This PR adds an additional accessor `distinctChildrenDeep` to the `GenParticle` that is similar to `distinctChildren` but recursively jumps over child particles with the same pdg id, resulting in a flat list of children with a different pdg id. This can be useful in scenarios with longish-decay chains and possible intermediate radiations. A simplified case is this top decay:

```mermaid
graph TD
    classDef target stroke: purple, stroke-width: 3px
    classDef selected stroke: green, stroke-width: 2px

    t1a(t)
    b1a(b)
    W1a(W1)
    q1a(q)
    q2a(q')

    class W1a target
    class q1a selected
    class q2a selected

    t1b(t)
    b1b(b)
    W1b(W1)
    W2b(W2)
    g1b(gamma)
    q1b(q)
    q2b(q')

    class W1b target
    class g1b selected

    t1c(t)
    b1c(b)
    W1c(W1)
    W2c(W2)
    g1c(gamma)
    q1c(q)
    q2c(q')

    class W1c target
    class g1c selected
    class q1c selected
    class q2c selected

    subgraph "<span style='color:purple;'>W1</span>.distinctChildrenDeep, case 2 <span style='color:green;'>✔︎</span>"
        t1c --> b1c
        t1c --> W1c
        W1c -. skipped .-> W2c
        W1c -- selected --> g1c
        W2c -- selected --> q1c
        W2c -- selected --> q2c
        
    end

    subgraph "<span style='color:purple;'>W1</span>.distinctChildren, case 2 <span style='color:red;'>✗</span>"
        t1b --> b1b
        t1b --> W1b
        W1b --> W2b
        W1b -- selected --> g1b
        W2b -- missing --> q1b
        W2b -- missing --> q2b
    end

    subgraph "<span style='color:purple;'>W1</span>.distinctChildren, case 1 <span style='color:green;'>✔︎</span>"
        t1a --> b1a
        t1a --> W1a
        W1a -- selected ---> q1a
        W1a -- selected ---> q2a 
    end
```

If one is interested in all W decay products (whose 4-vector sum should match that of the W again), `distinctChildren` does the job in case 1. However, in case 2, this would only yield the gamma. `disctinctChildrenDeep` provides the gamma as well as the q's, skipping any intermediate particles with the same pdg id.

### Changes in detail

- New `distinctChildrenDeep` property in `GenParticle`.
- New entry `GenPart_distinctChildrenDeepIdxG` in the `NanoAODSchema.special_items`.
- New kernel (similar to `_children_kernel`) and form (similar to `children_form`) in transforms.py.
- I used the `_children_kernel` as a starting point, but I was quite confused about the initial value of `offset0 = 0` (since `offsets1_out` is already filled with one element), and the order in which the "out of bounds" checks are performed. If I understand correctly, the checks should be performed right before the index is used rather than afterwards. I did it this way  in `_distinctChildrenDeep_kernel` and I hope it's okay that I changed it in `_children_kernel ` as well.
- Docstrings for all `GenParticle` accessors.
- Additional test.


lgray:
Please lint (latest black and flake8, `pip install .[dev]`). If you could expand the test to specifically check for the physics results you desire compared to the standard children it would be good as well. Otherwise this looks very nice, thank you!

lgray:
Given you're changing the indexing a little bit are we absolutely sure this does not change the behavior of the current distinctChildren indices?

riga:
I think the behavior doesn't change, but maybe @nsmith- can confirm.
Btw: I couldn't find a `distinctChildren` test case. I can add that one to be 100% sure.

lgray:
Yeah I noticed we don't have one. :cringe: please add it!

riga:
While working on the tests, I ran into a few edge cases where the `distinctChildrenDeep` accessor did not return what I initially had in mind. What I was going for is a lookup for decay products, **skipping intermediate decay chains** with same pdg ids, and the 4-vector sums of decay products should again add up to that of the parent particle (provided that the generator didn't mess things up at vertices). However, this requires that child particles should be considered as well even if they  **do** have the same pdg id but that themselves **do not** have children. Here is one example:

```mermaid
graph TD
    classDef target stroke: purple, stroke-width: 3px
    classDef selected stroke: green, stroke-width: 2px

    n0(1)
    n1(21)
    n2(21)
    n3(...)
    n4(21)
    n5(21)
    n6(21)
    n7(2)
    n8("-2")

    class n2 target
    class n4 selected
    class n5 selected
    class n7 selected
    class n8 selected

    subgraph "<span style='color:purple;'>target</span>.discintChildrenDeep -> <span style='color:green;'>selected</span>"
        n0 --> n1
        n1 --> n3

        n0 --> n2
        n2 --> n4
        n2 --> n5
        n2 --> n6
        n6 --> n7
        n6 --> n8
    end
```

With the initial implementation, only the 2 u's (-2/2) would have been selected. With the last commit the two child-less gluons (21) are included as well. With that change, I don't think that the name `distinctChildrenDeep` actually fits. Maybe `distinctOrLastChildren`? Up to you :)

Sorry for changing the objective mid-PR. If you agree with the change, I can adjust the PR description for future reference. One could toggle this behavior also with some flag, but I guess that could be implemented when there's the need for that.

<hr />

One other thing that I observed while testing this on CMS NanoAOD v9: in the global parent index it can happen in some rare cases that parent indices are listed after the indices of their children, which would not be detected by the current `children` accessor, nor by the lookup in this PR.

Do you know if this intended by the NanoAOD devs? I honestly don't hope so since that would require a different approach to finding children with N^2 complexity, compared to the N*logN that is implemented right now.
(I can also open a separate discussion on this)

lgray:
@kdlong Any thoughts on this w.r.t. expected behavior in NanoAODs?

riga:
> Given you're changing the indexing a little bit are we absolutely sure this does not change the behavior of the current distinctChildren indices?

I forgot that above: I added the `distinctChildren` test cases for the unchanged `_children_kernel`, and they still pass with the changes applied.

lgray:
@riga Isn't the `distictOrLastChildren` vs. `distictChildrenDeep` solved if you consider color flow as part of the identifying information of the particle? I don't recall off the top of my head if that's saved in the Nano gen info, but it would definitely resolve this issue in a very clean way.

kdlong:
So I would be very surprised if this changed between NanoV8 and NanoV9, I don't know of any changes to the gen particles happening here. I also don't think any special sorting of the genparticle collection happens for the NanoAOD. I'm not 100% sure that the order parent -> child is always respected in the original gen particle list, but I think it is. My first hypothesis for where the order gets mixed up is in the GenParticlePruner: https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/HepMCCandAlgos/plugins/GenParticlePruner.cc

A quick look and it seems to loop first over the selection criteria and then over the particles in an inner loop. Should look closer to understand exactly how this works, or maybe @bendavid can confirm.

To be more sure, I would take an example of a file where this happens, and read the genParticles collection from the AOD to see if the order you expect is respected. Then I would look at the prunedGenParticles in the Mini, and finally the GenPart collection in the Nano to see where this gets mixed up.

coffea/nanoevents/transforms.py
```diff
@@ -226,22 +226,22 @@ def _children_kernel(offsets_in, parentidx):
     content1_out = numpy.empty(len(parentidx), dtype=numpy.int64)
     offsets1_out[0] = 0
 
-    offset0 = 0
+    offset0 = 1
```
nsmith-, line 229: Regarding this change, if you look at L241 in the previous code, it was `offsets1_out[offset0 + 1] = offset1` so the +1 was handling the shift. By my reading, the new code works the same so I'm fine with the change. It does read a bit more clearly now as well, which is nice.

coffea/nanoevents/transforms.py
```diff
@@ -288,6 +288,138 @@ def children(stack):
...
+    offset1 = 0
+    for record_index in range(len(offsets_in) - 1):
+        start_src, stop_src = offsets_in[record_index], offsets_in[record_index + 1]
+
+        for index in range(start_src, stop_src):
+            this_pdg = global_pdgs[index]
+
+            # only perform the deep lookup when this particle is not already part of a decay chain
+            # otherwise, the same child indices would be repeated for every parent in the chain
+            # which would require content_out to have a length that isa-priori unknown
```
nsmith-, line 307: `is a-priori`

coffea/nanoevents/transforms.py
```diff
@@ -288,6 +288,138 @@ def children(stack):
...
+            if (
+                global_parents[index] >= 0
+                and this_pdg != global_pdgs[global_parents[index]]
+            ):
+                # keep an index of parents with same pdg id
+                parents = numpy.empty(stop_src - index, dtype=numpy.int64)
```
nsmith-, line 313: I'm a bit worried about the allocation time here. Is it possible the upper bound is known and it can be allocated outside the event loop?

riga:
@kdlong Thanks for looking into this. An example is [/store/mc/RunIISummer20UL18NanoAODv9/ST_t-channel_top_4f_InclusiveDecays_TuneCP5_13TeV-powheg-madspin-pythia8/NANOAODSIM/106X_upgrade2018_realistic_v16_L1v1-v1/2550000/2325C825-4095-D74F-A98A-5B42318F8DC4.root](https://cmsweb.cern.ch/das/request?instance=prod/global&input=file=/store/mc/RunIISummer20UL18NanoAODv9/ST_t-channel_top_4f_InclusiveDecays_TuneCP5_13TeV-powheg-madspin-pythia8/NANOAODSIM/106X_upgrade2018_realistic_v16_L1v1-v1/2550000/2325C825-4095-D74F-A98A-5B42318F8DC4.root) at event (index) 53.

For gen particles 6-8, I see

| Gen part index | Mother index | Gen part PDG |
| -------------: | -----------: | -----------: |
|            ... |              |              |
|              6 |            7 |            4 |
|              7 |            8 |            4 |
|              8 |           -1 |            4 |
|            ... |              |              |

It looks like a reversed decay chain of the same particle type. Gen part 6 has a bunch of other non-c children, but they are listed correctly below that. No other gen particles have index 7 or 8 as their mother, so the connection from the first particle in the chain (8) down to chain (7 to 6 to the other non-c particles) would be lost if the lookup for children starts at index 8.

This happens quite rare through (328 out of 50k), but it caused a test case to fail.

coffea/nanoevents/transforms.py
```diff
@@ -288,6 +288,138 @@ def children(stack):
...
+                # keep an additional index with parents that have at least one child
+                parents_with_children = numpy.empty(stop_src - index, dtype=numpy.int64)
+                offset3 = 0
+
+                for possible_child in range(index, stop_src):
+                    possible_parent = global_parents[possible_child]
+                    possibe_child_pdg = global_pdgs[possible_child]
+
+                    # compare with seen parents
+                    for parent_index in range(offset2):
```
nsmith-, line 326: I can't say I understand this section yet but I somehow expected a tail recursion.

coffea/nanoevents/transforms.py
riga: That might be possible. The upper bound should be the length of particles itself, and then one could search between 2 offsets. I'll give it a try.

riga:
> @riga Isn't the `distictOrLastChildren` vs. `distictChildrenDeep` solved if you consider color flow as part of the identifying information of the particle? I don't recall off the top of my head if that's saved in the Nano gen info, but it would definitely resolve this issue in a very clean way.

Looks like this info isn't stored, but I also don't think that having it could identify all cases (or maybe I'm not seeing sth ...).

The question whether one wants `distictChildrenDeep` (initial PR) or `distictOrLastChildren` (updated PR)  is not that easy to answer. I'd prefer the latter, since users could simply add another filter for the pdg id in case they want to former.

coffea/nanoevents/transforms.py
riga: Thinking about it, the `parents_with_children` array should be moved out of the loops as well, but I'm anyway not 100% happy with how children with same pdg id but with no further grandchildren (\*) are identified. Do you think it's a good idea to pass the children index to the kernel to identify (\*)? This could be used here:

```python
# then, depending on the pdg id, add to parents or content
if possibe_child_pdg == this_pdg and has_children(possible_child):
    # has the same pdg id and further children, add to parents
    ...
```

kdlong:
Hi @riga, thanks for the example. I'm about to be away on vacation so it's very unlikely that I'll do this check myself in the coming weeks. I'm also not terribly invested in the performance of this function :) I would just point out that there is no requirement in the HepMC format that requires the ordering the you are counting on, so the safest thing to do would be to have some fallback option that doesn't assume ordering.

Now it may be that Pythia tends to respect this ordering anyway, I'm not 100% sure. As I said the thing to do is to check the same event at Gen/AOD level, and then at MiniAOD level, to see if the ordering is not respected already at the HepMC or if the gen particle pruning etc mixes it.

bendavid:
In general you should assume that the details of the generator truth are maximally generator dependent and follow very few fixed conventions.  (There are a small number of fixed rules about status 1 and status 2 particles).  Any checks based on what is "typical" in NanoAOD likely just reflects the mix of generators being used for a given production.  A new generator or generator version, or change of setup for "afterburners" like EVTGEN/Tauola/Photos may change/break these assumptions at any moment.

riga:
Thanks for your inputs @kdlong and @bendavid . Do you think it would make sense to enforce this order specifically during the central NanoAOD production? Or is there some kind of constraint that would require the order to match exactly that in upstream data tiers?

bendavid:
No, it's in general a terrible idea to try and enforce behaviour/modify the generator truth information beyond simply information reduction.  Any attempt at this is likely to be fragile/prone to break with different generators/versions.  (Even if the pruning/slimming itself is already fragile in this respect.)

riga:
In general I agree with you. But in this case it would just mean that when writing the gen particles to Nano, daughter particles are always listed after their parents, i.e., the generator information itself (esp. parent-child connections) would not be changed by that.

coffea/nanoevents/transforms.py
nsmith-: As far as I know, numba should be able to compile functions that have local functions defined inside. So yeah that would work and probably improve readability

lgray:
Hey folks so what's the plan here?

lgray:
What do we want to do with this? @nsmith- @riga

lgray:
To finally come back to this one @riga:
- It's opt in so it shouldn't break any changes for people's analysis
- a bug is fixed in generated particle lookups (perhaps a confirmation of that @nsmith-?)

I think all that needs to be done is some warning in the documentation that this function may not work for all generators.
We can create an issue and maybe this gets fixed in time.

nsmith- (approved):
I'm happy with this if @riga is. I'm not sure what the WIP in the title refers to so I think he should decide. If it is incomplete functionality we can make an issue to remind us and merge this to clear the PR queue.

lgray:
The functionality desired was there since a long time, the reason this one got stuck is because the functionality only works for certain generators and there is no guarantee it won't be broken at any time.

However, it doesn't alter the current operation of gen particles and I added a warning to the function... which is good enough for me.

lgray:
@riga open another PR if you want to continue on this.

Issue #699, opened 2022-07-19
Reduce memory usage of function results


btovar:
**Is your feature request related to a problem? Please describe.**

When using compression, functions are wrapped with:

```
# compression
return lz4f.compress(
            pickle.dumps(item, protocol=_PICKLE_PROTOCOL), compression_level=compression
        )

# decompression
return pickle.loads(lz4f.decompress(item))
```

This works well for small results, but may cause excessive memory usage for results that are larger than 1GB.  The memory used consists of:
- the actual result item
- its pickled representation
- the pickled representation compressed

In one of our runs, the pickled representation compressed is around 3GB, which means than a lower bound of extra memory is 6GB (more because the pickle representation is larger than the compressed one).  The final accumulation result of this run needed a peak of 21GB to be expanded

**Describe the solution you'd like**

We would like the final memory usage to be as close as possible to the minimum required by the result. 

**Describe alternatives you've considered**

The first alternative that we tried was to decouple the calls to something like:

```
item_pickled = lz4f.decompress(item_compressed)
del item_compressed
item = pickle.loads(item_pickled)
del item_pickled  # with no effect really, as we are about to return anyway
return item
```

Such scheme turn our max memory consumption from 21GB to about 18GB.

Further, since we are working with the WorkQueue executor that transmits data from functions at remote workers via files, we were able to compress/decompress and read/write directly the pickled output as:

```
# compress
with lz4f.open(filename, 'wb', compression_level=compression) as rf:
   return pickle.dump(item, rf, protocol=_PICKLE_PROTOCOL)

# decompress
with lz4f.open(filename, 'rb') as rf:
   return pickle.load(rf)
```

With such scheme, the memory consumption went down to 12GB for the final accumulation task. With a specialization on the dict_accumulator, we tried to pickle each of the values in the dictionary one at a time, but this did not change the memory consumption by much (around 11.5GB), so probably `pickle.load` knows to interpret each value by itself and does not need to load the whole file into memory before unpickling.

Where do you think changes like this should be incorporated in coffea?  Currently the _compression_wrapper and related utility functions are part of the` processor.executor` module, but perhaps they can be methods an executor can specialize? In our current prototype we simply override the _compression_wrapper in the WorkQueueExecutor.

[...]

PR #700, opened 2022-07-21
simultaneously pickle and compress results of user functions

btovar:
See issue #699.  In the original implementation, coffea held in memory the item to be compressed, its pickled representation, and the compressed representation. Using a BytesIO object, we only need space (roughly) for the item and the compressed representation when decompressing, and the item and two times the compressed representation when compressing.

When compressing, the compressed representation has to be copied out of the BytesIO object, and that is why it has to be accounted twice. The memory savings come from the difference between the pickled and the compressed representation. This may open some opportunity for improvement if a scheme can be found where this copy is not necessary.

As an example of memory savings, one of topcoffea's workflows reduced its max memory consumption from 20.6GB to 13.7GB.

### DIFF
--- a/coffea/processor/executor.py
+++ b/coffea/processor/executor.py
@@ -17,6 +17,7 @@ import traceback
 import shutil
 from collections import defaultdict
 from cachetools import LRUCache
+from io import BytesIO
 import lz4.frame as lz4f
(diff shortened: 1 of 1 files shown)


btovar:
Thanks Lindsey!

PR #707, opened 2022-08-09
feat: support custom schema specification with ServiceX executor

alexander-held:
The schema used with the ServiceX executor was previously hardcoded to `auto_schema`. This adds a new `schema` keyword argument to `coffea.processors.servicex.LocalExecutor.execute` and `coffea.processors.servicex.DaskExecutor.execute` to set a schema. The default value of `None` is resolved in `run_coffea_processor` to again make use of the `auto_schema`.

The `LocalExecutor` path works, `DaskExecutor` is currently untested (I cannot get coffea+ServiceX+dask to work locally, and this is not as easy to test on coffea-casa as I think I would need to propagate the changes to workers as well).

### DIFF
--- a/coffea/processor/servicex/dask_executor.py
+++ b/coffea/processor/servicex/dask_executor.py
@@ -71,6 +71,7 @@ class DaskExecutor(Executor):
         data_type: str,
         meta_data: Dict[str, str],
         process_func: Callable,
+        schema,
     ):
         """Create a dask future for a dask task to run the analysis."""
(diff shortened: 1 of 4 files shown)


nsmith- (approved):
Looks ok to me, any reason not to merge now?

alexander-held:
I had not managed to properly test the version with `DaskExecutor` locally before, but it does look to be working fine. Thanks for merging!

PR #712, opened 2022-08-16
wq: option to activate ssl between manager and workers

btovar:
The WorkQueueExecutor accepts the parameter `ssl`. If `ssl=True`, then
temporary autosigned key and certs are generated and used. Otherwise,
ssl=(key,cert) should indicate the paths to a key and cert in pem
format.

When activating ssl in the wq manager, workers should be launched with
the --ssl command line argument.


btovar:
Included in #716

PR #719, opened 2022-08-31
Simplify chunk generation

btovar:
Cleans the chunk generation a little bit by removing the condition on dynamic_chunksize. Now, the generator correctly responds to updates to the chunksize regardless of whether the currently implemented mechanism for dynamic_chunksize is being used.

If no updates to chunksize are made, then the generator acts exactly as before. (Equally divide a file into chunks of about the same size based on the target chunksize.)

(Giving up for now in making maxchunks work as a generator too. iterators and `yield from` do not work well together.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,7 @@ jobs:
         conda create --name coffea-env python=3.7
         conda activate coffea-env
         conda install -c conda-forge ndcctools dill
-        python -m pip install .
+        python -m pip install --ignore-installed .
         cd tests
         python wq.py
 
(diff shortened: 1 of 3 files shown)


btovar:
Tests that fail pass locally. I wonder if it is a temporary error with those parquet tests, as the code modified here is under a `if self.format == "root":`. (Note it worked for windows py 3.6, but no py 3.10.)

nsmith- (approved):
Looks good. Is there a unit test for the logic already in place?

btovar:
I think that `test_local_executors.py` covers it, but I can write a more specific test if you need.

btovar:
Checking that conda error...

btovar:
It looks like a snag that appeared overnight related to both conda and pip trying to install setup.py. Usually pip does not reinstall a package if it sees it already installed by conda, but in this case setuptools is being reinstalled. (Same version but slightly different contents.)

PR #720, opened 2022-08-31
update Nano*Processor in test_items to hist

btovar:
Updates test_items/{NanoEventsProcessor.py,NanoTestProcessor.py} to use hist.

I had no complications following the guide in https://github.com/CoffeaTeam/coffea/discussions/705
(Spoke too soon, there seems to be a problem with the windows setup.)

### DIFF
--- a/coffea/processor/test_items/NanoEventsProcessor.py
+++ b/coffea/processor/test_items/NanoEventsProcessor.py
@@ -29,8 +29,11 @@ class NanoEventsProcessor(processor.ProcessorABC):
         pt_axis = hist.axis.Regular(30000, 0.24, 300, name="pt", label=r"$p_{T}$ [GeV]")
 
         accumulator = {
-            "mass": hist.Hist(dataset_axis, mass_axis, name="Counts"),
-            "pt": hist.Hist(dataset_axis, pt_axis, name="Counts"),
+            # replace when py3.6 is dropped
+            # "mass": hist.Hist(dataset_axis, mass_axis, name="Counts"),
+            # "pt": hist.Hist(dataset_axis, pt_axis, name="Counts"),
+            "mass": hist.Hist(dataset_axis, mass_axis),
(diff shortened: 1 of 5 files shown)


btovar:
Windows test is failing with:

```
2022-08-31T13:44:04.6316958Z         accumulator = {
2022-08-31T13:44:04.6317335Z >           "mass": hist.Hist(dataset_axis, mass_axis, name="Counts"),
2022-08-31T13:44:04.6317772Z             "pt": hist.Hist(dataset_axis, pt_axis, name="Counts"),
2022-08-31T13:44:04.6318156Z             "cutflow": defaultdict(int),
2022-08-31T13:44:04.6318713Z             "worker": set(),
2022-08-31T13:44:04.6318994Z         }
2022-08-31T13:44:04.6319677Z E       TypeError: __init__() got an unexpected keyword argument 'name'
```
It seems that the `name` keyboard is not valid there. It is valid on linux. Is there something known to be different with the hist module in windows?

nsmith-:
Looks like a more recent feature than the minimum version coffea requires:
https://github.com/scikit-hep/hist/pull/297
was added for hist 2.5.0, the CI was using 2.4.0

nsmith-:
I think since coffea is meant partially to be a framework it would be ok to bump the requirement in setup.py from `hist>=2` to 2.5

nsmith-:
aha and I guess they stopped supporting python 3.6 with 2.5

nsmith-:
Ok maybe for now we drop the name and put this as a TODO for python 3.6 deprecation (it probably should come soon)

btovar:
making a note that `.compatible` is also missing.

nsmith-:
Ah, I see. I will add that to the guide, I think you could replace with `newprocessor.accumulator["pt"].axes == aprocessor.accumulator["pt"].axes`. We'll also have to stop using `.identity()` (see e.g. the very last section of the migration guide)

PR #725, opened 2022-09-02
wq docs update

btovar:
- shows password option
- updates from to hist
- shows factory outside python
- adds comments for factory's scratch directory and conda env


nsmith- (approved):
Good to go?

btovar:
Yes, thanks!

PR #732, opened 2022-10-13
remove windows ci

lgray:


nsmith-:
What made us have to remove windows? Is there an issue to track possibly adding it back?

lgray:
It started failing and I had no machine to figure it out on. I was gonna put in the time to bring it back with 0.8 since with dask awkward we could even have analysis tests running on windows with dask_executor.

PR #733, opened 2022-10-13
preserve __array__ parameter = char for arrays of strings read by uproot

vindex10:
Hi! I had a problem reading a root file that as one of the branches contained a string. Here is an example of the `form`:

```
{
    "class": "ListOffsetArray64",
    "offsets": "i64",
    "content": {
        "class": "NumpyArray",
        "itemsize": 1,
        "format": "B",
        "primitive": "uint8",
        "parameters": {
            "__array__": "char"
        }
    },
    "parameters": {
        "__array__": "string"
    }
}
```

method [_lazify_form](https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/mapping/uproot.py#L28) removes the parameter "__array__" = "char" when `docstr` is provided.

This prevents awkward library to treat the ListOffsetArray as string, and it fails with the error:

```
Input In [109], in Process.process(self, batch)
      9 #display(batch.systName.layout)
---> 10 batch = batch[ak.argsort(batch["systName"])]
     11 names = np.unique(batch["systName"])

File ~/projects/venvs/sandbox/lib/python3.10/site-packages/awkward/operations/structure.py:2661, in argsort(array, axis, ascending, stable, highlevel, behavior)
   2658 layout = ak.operations.convert.to_layout(
   2659     array, allow_record=False, allow_other=False
   2660 )
-> 2661 out = layout.argsort(axis, ascending, stable)
   2662 return ak._util.maybe_wrap_like(out, array, behavior, highlevel)

ValueError: at  (ListOffsetArray64): __array__ = "string" must directly contain a node with __array__ = "char"
```

In this pull request I suggest to preserve the "__array__" parameter if it is present.

### DIFF
--- a/coffea/nanoevents/mapping/uproot.py
+++ b/coffea/nanoevents/mapping/uproot.py
@@ -28,7 +28,9 @@ class CannotBeNanoEvents(Exception):
 def _lazify_form(form, prefix, docstr=None):
     if not isinstance(form, dict) or "class" not in form:
         raise RuntimeError("form should have been normalized by now")
-    elif form["class"].startswith("ListOffset"):
+
+    parameters = _lazify_parameters(form.get("parameters", {}), docstr=docstr)
+    if form["class"].startswith("ListOffset"):
         # awkward will add !offsets
         form["form_key"] = quote(prefix)
         form["content"] = _lazify_form(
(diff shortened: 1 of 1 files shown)


vindex10:
That was quick! Thank you :)

PR #736, opened 2022-10-25
Awkward v2 transition

lgray:
- [ ] Get all tests working again regardless of laziness in nanoevents (I should break this out into sub-bullets)
- [x] base schema passes tests
- [x] nanoaod schema passes tests
- [x] protodune schema passes tests
- [x] physlite schema passes tests
- [x] delphes schema passes tests
- [x] treemaker schema passes tests

- [x] local executor tests pass
- [x] dask executor passes tests
- [x] parsl executor passes tests
- [ ] spark executor passes tests (requires spark upgrade and new data ingestion step!)

- [x] lookup tools passes tests
- [x] jetmet tools - individual correctors
- [x] jetmet tools - CorrectedJetsFactory (heavily lazy)

- [ ] the above, again, except with laziness implemented in dask_awkward


lgray:
@kratsg hey - what does a form key that looks like:
```
Gettting ( ad4cd5ec-123e-11ec-92f6-93e3aac0beef/%2FDelphes%3B1/0-25/data/GenJet%2FGenJet.SoftDroppedJet%2C%21load%2C%21content%2C%40instance_version%2C%21item ) : ad4cd5ec-123e-11ec-92f6-93e3aac0beef /Delphes;1 0 25 ['GenJet/GenJet.SoftDroppedJet', '!load', '!content', '@instance_version', '!item']
```
mean? In particular the `@instance_version` token? The internal structure of RecordArray forms changed quite a bit in awkwardv2 and nanoevents makes quite a few assumptions.

kratsg:
> In particular the `@instance_version` token? The internal structure of RecordArray forms changed quite a bit in awkwardv2 and nanoevents makes quite a few assumptions

Checking, but I don't recall an `instance_version` being added...

lgray:
Gotcha - things broke in interesting ways - there's only so much of this code nick/myself actually own and know how to maintain! Thanks for looking.

lgray:
@gordonwatts You should be aware that things have changed a lot with forms under the hood in awkward2. I have gotten your tests to pass again for the auto schema but there is not much confirmation that they work beyond that. @'ing you since it would be worth your time to work the rest of this out. 

I'm not sure if the servicex tests are awkward2 compatible yet either.

lgray:
@tejinc awkward2 brings quite some changes to the inner workings of coffea. I have gotten your nanoevents schema (pdune) to pass tests but you may want to check it to ensure proper functionality is kept.

lgray:
@kpedro88 awkward2 brings quite some changes to the inner workings of coffea. I have gotten your nanoevents schema (treemaker) to pass your tests but you may want to check it to ensure proper functionality is kept.

kpedro88:
> @kpedro88 awkward2 brings quite some changes to the inner workings of coffea. I have gotten your nanoevents schema (treemaker) to pass your tests but you may want to check it to ensure proper functionality is kept.

@yimuchen @cmadrid1 @Keane-Tan FYI

coffea/nanoevents/methods/candidate.py
```diff
@@ -12,14 +12,14 @@
 behavior = dict(vector.behavior)
 
 
-@awkward.mixin_class(behavior)
+@awkward.behaviors.mixins.mixin_class(behavior)
```
agoose77: We expose this under `ak.mixin_class` in the latest v2 pre-releases. I know at one point we did miss this, but it's not necessary to use the long-form any more.

lgray:
@nikoladze awkward2 brings quite some changes to the inner workings of coffea. I have gotten your nanoevents schema (physlite) to pass your tests but you may want to check it to ensure proper functionality is kept.

lgray:
@kratsg awkward2 brings quite some changes to the inner workings of coffea. I have gotten your nanoevents schema (delphes) to pass your tests but you may want to check it to ensure proper functionality is kept.

kratsg:
> @kratsg awkward2 brings quite some changes to the inner workings of coffea. I have gotten your nanoevents schema (delphes) to pass your tests but you may want to check it to ensure proper functionality is kept.

Checked it and it seemed ok out of the box? It at least works for me for the quick tests I ran so I think it's ok.

EDIT: works on some other older Delphes files I had too that crashed with coffea before... so I think the changes are going to give us better support(?!). Thanks for all the work!

yimuchen:
`TreeMakerSchema` looks good. But found another bug when loading in a large number events: 

```python
events = NanoEventsFactory.from_root("file:mytestfile.root", entry_start=0, entry_stop=100, schemaclass=TreeMakerSchema, treepath='TreeMaker2/PreSelection').events() # This works fine
events = NanoEventsFactory.from_root("file:mytestfile.root", entry_start=0, entry_stop=1000, schemaclass=TreeMakerSchema, treepath='TreeMaker2/PreSelection').events() # This errors
```

Traceback dump:
```
  File "coffeaenv/lib/python3.8/site-packages/coffea/nanoevents/factory.py", line 399, in events
    events = awkward.from_buffers(
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 76, in from_buffers
    return _impl(
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 121, in _impl
    out = reconstitute(form, length, container, getkey, backend, simplify)
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 296, in reconstitute
    contents = [
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 297, in <listcomp>
    reconstitute(content, length, container, getkey, backend, simplify)
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 274, in reconstitute
    content = reconstitute(
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 296, in reconstitute
    contents = [
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 297, in <listcomp>
    reconstitute(content, length, container, getkey, backend, simplify)
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 274, in reconstitute
    content = reconstitute(
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 296, in reconstitute
    contents = [
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 297, in <listcomp>
    reconstitute(content, length, container, getkey, backend, simplify)
  File "coffeaenv/lib/python3.8/site-packages/awkward/operations/ak_from_buffers.py", line 148, in reconstitute
    data = backend.nplike.frombuffer(raw_array, dtype=dtype, count=real_length)
  File "coffeaenv/lib/python3.8/site-packages/awkward/_nplikes.py", line 115, in frombuffer
    return self._module.frombuffer(*args, **kwargs)
ValueError: buffer is smaller than requested size
```

lgray:
@yimuchen thanks! what's the size of the file you're testing against?

yimuchen:
> @yimuchen thanks! what's the size of the file you're testing against?

This is a 2GB file containing ~60K events. So about 35KB (?) per event.

coffea/lookup_tools/lookup_base.py
```diff
@@ -18,33 +18,28 @@ def __call__(self, *args, **kwargs):
...
+                isinstance(
+                    x, (awkward.contents.NumpyArray, awkward.contents.EmptyArray)
                 )
-                for x in inputs
+                or not isinstance(x, (awkward.contents.Content))
+                for x in args
             ):
-                nplike = awkward.nplike.of(*inputs)
-                if not isinstance(nplike, awkward.nplike.Numpy):
+                nplike = awkward._nplikes.nplike_of(*args)
```
agoose77: @lgray we spoke about this on Slack, but in the interest of my not forgetting — on the Awkward side, we'd prefer Coffea not to rely on any private APIs such as `nplikes`. The solution here is to check `ak.backend(*args)`, and exit if it is not equal to the string `"cpu"`.

coffea/lookup_tools/lookup_base.py
```diff
@@ -18,33 +18,28 @@ def __call__(self, *args, **kwargs):
...
-        args = [
-            awkward.operations.convert.to_layout(
-                arg, allow_record=False, allow_other=True
-            )
-            for arg in args
-        ]
-        out = awkward._util.broadcast_and_apply(args, getfunction, behavior)
-        assert isinstance(out, tuple) and len(out) == 1
-        return awkward._util.wrap(out[0], behavior=behavior)
+        behavior = awkward._util.behavior_of(*args)
```
agoose77: `ak.transform` takes care of preserving the behaviour here :)
```suggestion
```

coffea/lookup_tools/lookup_base.py
lgray: yes - have been dealing with other things - was on my list to fix!

lgray:
@yimuchen hmm, and how many events?

lgray:
@yimuchen can you try seeing if it's a particular event that causes problems, or if it's just really the size of the event range you want to pick up?

lgray:
Note to self: python 3.11 is a limitation of at least numba

yimuchen:
> @yimuchen can you try seeing if it's a particular event that causes problems, or if it's just really the size of the event range you want to pick up?

Tried a few points, it looks like if the event size fails for some slice size (seems fail at about ~400 events in this case), it would consistently fail. So it isn't a case of a single very large event.

lgray:
@yimuchen so it's really just the absolute size of the slice? or where you are in the file? or both? It may be useful to scan through the file 100 events at a time to make sure there's not some pattern that's bugging out.

coffea/nanoevents/mapping/uproot.py
```diff
@@ -47,14 +49,21 @@ def _lazify_form(form, prefix, docstr=None):
...
             if "," in field or "!" in field:
+                # Could also skip here
                 raise CannotBeNanoEvents(
                     f"A subform contains a field with invalid characters: {field}"
                 )
-            form["contents"][field] = _lazify_form(
-                form["contents"][field], prefix + f",{field},!item"
-            )
+            elif field.startswith("@"):
+                # workaround uproot5 bug
```
nsmith-, line 60: Bug is fixed in https://github.com/scikit-hep/uproot5/pull/808

lgray:
Tests will be broken for a bit as I get things daskified.

lgray:
Current corrected jet/met factory tests will fail until:
https://github.com/dask-contrib/dask-awkward/pull/137

is merged and a new dask awkward is released.

Other things to figure out: CorrectedXFactory creates task graphs that are 100's - 1000's of node long. 
Presently this is extremely slow and needs to be significantly improved before we make a release.

lgray:
@matthewfeickert rejoice, coffea passes tests on arm!

matthewfeickert:
> @matthewfeickert rejoice, coffea passes tests on arm!

MontyPython_There-Was-Much-Rejoicing

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -52,11 +61,22 @@ def getfunction(layout, depth, **kwargs):
...
-    )
+    out = None
+    if backend == "cpu":
+        out = awkward.transform(
+            getfunction,
+            item,
+            behavior=item.behavior,
+        )
+    elif backend == "typetracer":
+        zlitem = item.layout.form.length_zero_array()
```
agoose77, line 72: We're trying to avoid the need to use `length_zero_array` in typetracer code unless it's 100% needed. Would you be able to keep tabs on places where you're using `length_zero_array` for workarounds so that we can remove them in future?

lgray: Yes - some of this is going through growing pains as I start using `dask_awkward` as the data source instead of `dak.from_awkward`, the former has rather more intense usage of typetracers.

Once we're on the other side I'll make sure it's minimal and let you know the rest.

lgray: The one you highlight here will probably stay since I need to feed a non-zero-length array to as a PRNG seed.

yimuchen:
> @yimuchen so it's really just the absolute size of the slice? or where you are in the file? or both? It may be useful to scan through the file 100 events at a time to make sure there's not some pattern that's bugging out.

Sorry it took me some time to get round to retesting this. 

It looks like there are certain patterns for TreeMaker n-tuples that is causing the NanoEventsFactory to fail, a simple test script for stepping over the TreeMaker n-tuples in increaments of 20 events, sometime it fails, other times it does not.

```python
for f in [
        f'root://cmseos.fnal.gov//store/group/lpcemj/ExoEMJAnalysis2020/Run2ProductionV20/Summer20UL18/QCD_HT1500to2000_TuneCP5_PSWeights_13TeV-madgraph-pythia8/{i}_RA2AnalysisTree.root'
        for i in range(5)
        ]:
  step_size = 20
  for i in range(0,1000,step_size):
   print('Running events', i, '-', i+step_size, ' in file', f)
   try :
                  f,
                  treepath="TreeMaker2/PreSelection", 
                  metadata={"dataset": "test"}, 
                  schemaclass=TreeMakerSchema,    
                  permit_dask=True,
                  entry_start=i, 
                  entry_stop=i+step_size,).events() 
    # This is a rather odd question, Using TreeMakerSchema, we will need the `.events()` but for BaseSchema this isn't needed?
    # DaskAwkwardTest().process(events)
   except Exception as err:
    print(err)
    raise err
```

lgray:
Weird - thank you for checking more deeply. Let me update TreeMakeSchema to properly use dask, then let's try again. It may be some issue of eager mode uproot as well (which will require fixing but changes priority).

lgray:
@yimuchen So I suspect there's something deeper going on here given the rather different interpretation everything is being cast to. In particular if I try to `uproot.dask` a the treemaker tree the base form I get out is not very useful.

This may require adding more flexibility to uproot.dask. Could you try taking a look as well?

lgray:
Ah so with permit_dask the schema class also needs to be `__dask_capable__` and while working on things the way `from_root` works depends on that. I'll clean it up. 

In any case if the schema is `__dask_capable__` and you set `permit_dask=True` then `from_root` will return the events object directly, for now.

lgray:
What's really interesting is that all the forms look right but it's mapped super weirdly into fields.

lgray:
Figured out the issue with TreeMaker - it looks like the schema processing is significantly simpler than before.

lgray:
@nikoladze I could use your help in daskifying PHYSLITESchema. Something is going wrong in zipping the forms but everything looks more or less OK.

I think if that can be fixed up it'll more or less work. You'll need to use uproot main to get it to run in the first place.

lgray:
@kratsg I think I did this correctly. Can you please check the usage of NanoEvents Delphes with dask_awkward? I have no idea what correct use really looks like.

nikoladze:
> @nikoladze I could use your help in daskifying PHYSLITESchema. Something is going wrong in zipping the forms but everything looks more or less OK.
> 
> I think if that can be fixed up it'll more or less work. You'll need to use uproot main to get it to run in the first place.

Thanks a lot for putting all this effort into this! See #776 for a fix for the form zipping.

lgray:
@nikoladze there's still some problems in the rendered arrays with the trackParticleLinks - look at the test failures. :-/

Ah - I see it's passing around some dask_awkward stuff within all your offsets functions. I'll get to it in time but it's a fair amount of rather specific code, it would be best for you to figure out where to put in the map_partitions calls and such.

Basically the big change with dask_awkward is that we can only use the factory-reference-through-behavior trick on the client side, so your arrays have to render to something that doesn't require reindexing later.

lgray:
These latest commits require to be-released features in dask-awkward.

lgray:
OK so what we're gonna do is merge this puppy and start release candidates. There is a very short list of things that need doing to get a full release on the coffea side.

@nikoladze @kratsg I will need your help to fix your schemas. 

Parquet nanoevents will have to wait for a better implementation of form-remapping in dask-awkward (which should also tidy up some code in uproot). @martindurant @douglasdavis

I will reformulate the task list from this PR into the remaining tasks that need doing, and release an RC for people to start exploring what's new.

lgray:
@btovar @dthain we should talk about workqueue + dask = taskvine? as well, times are a-changin'.

PR #740, opened 2022-11-01
Patching the implementation of files with missing trees as bad files

MoAly98:


MoAly98:
@lgray I am actually wondering -- when I run black locally it changes a non-negligible parts of the exisiting code base, that's why often my PR often fail black before I fix -- I avoid running black locally not to mess with existing code. How do you deal with that?

lgray:
You may be running an older version of black? Coffea is always linted with latest black in latest stable python (it will switch to 3.11 soon).

MoAly98:
> You may be running an older version of black? Coffea is always linted with latest black in latest stable python (it will switch to 3.11 soon).

If I do
```
 import black
black.__version__ 
 ```
I get `22.10.0` which is the newest vesion. But on command line when I do 
```
black --version
```
I get `19.10b0`. The environment I am in was created using `Coffea`'s setup so I don't have any other packages in this environment other than those needed by `coffea`.

lgray:
Try `python -m black --version`?
If that gives you the latest version you can then do `python -m black tests coffea setup.py`.

MoAly98:
> Try `python -m black --version`? If that gives you the latest version you can then do `python -m black tests coffea setup.py`.

yep `python -m black` does it. Thank you!

PR #743, opened 2022-11-02
Unrestrict numpy versions for more recent python

### LINKED ISSUE #741: Upperbound pin on `numpy` unncessary?
It appears `numba` 0.56 and 0.55.2 now support `numpy` 1.22. Can this upper bound pin be removed?

https://github.com/CoffeaTeam/coffea/blob/52e102fce21a3e19f8c079adc649dfdd27c92075/setup.py#L68

See: https://github.com/numba/numba/issues/8025

lgray:
See if we can specify this in setup.py

Fixes #741


lgray:
Nice, that seems to solve it.

PR #745, opened 2022-11-05
Deprecate nanoevents.methods.vector

lgray:
will be replaced with vector package in 0.8.0


nsmith-:
I'm pretty sure we may want to deprecate this slower. My thought was to add a `.p4` method that uses `vector` behavior to all coffea vectors, and then warn user if any of the coffea vector mixin attributes are accessed to instead switch to `.p4.X`

PR #749, opened 2022-11-14
Added support for multiple systematic variations of a single weight

valsdav:
This covers the need for multiple systematic variation of a single weight added to the event weight. 
E.g. the multiple btag SF variations in CMS.

A list of modifierNames is added as argument to the function `add_multivariation`. The final modifier to the event name is formatted as `name_modifierUp/Down`.

I added a separate function instead of modifying `add()` since the type of the input parameters is different. Please let me know if you prefer to do in another way. 

The corresponding pytest has been included.

### DIFF
--- a/coffea/analysis_tools.py
+++ b/coffea/analysis_tools.py
@@ -118,6 +118,82 @@ class Weights:
             weight.size,
         )
 
+    def add_multivariation(
+        self, name, weight, modifierNames, weightsUp, weightsDown, shift=False
+    ):
+        """Add a new weight with multiple variations
+
+        Each variation of a single weight is given a different modifier name.
+        This is particularly useful e.g. for btag SF variations.
+
+        Parameters
+        ----------
+            name : str
+                name of correction
+            weight : numpy.ndarray
+                the nominal event weight associated with the correction
+            modifierNames: list of str
+                list of modifiers for each set of weights variation
+            weightsUp : list of numpy.ndarray
+                weight with correction uncertainty shifted up (if available)
+            weightsDown : list of numpy.ndarray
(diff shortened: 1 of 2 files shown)


coffea/analysis_tools.py
```diff
@@ -118,6 +118,82 @@ def add(self, name, weight, weightUp=None, weightDown=None, shift=False):
...
+                weightUp[weight != 0.0] /= weight[weight != 0.0]
+                self._modifiers[name + "_" + modifier + "Up"] = weightUp
+            if weightDown is not None:
+                weightDown = coffea.util._ensure_flat(weightDown, allow_missing=True)
+                if isinstance(weightDown, numpy.ma.MaskedArray):
+                    weightDown = weightDown.filled(1.0)
+                if shift:
+                    weightDown = weight - weightDown
+                weightDown[weight != 0.0] /= weight[weight != 0.0]
+                self._modifiers[name + "_" + modifier + "Down"] = weightDown
```
nsmith-: It would be nice to break this almost-duplication of L97-112 into a helper function.

nsmith- (approved):
I think this looks all good. I guess there is a new convention implied by the `{name}_{modifier}Up` but it makes sense I think.
Certainly an improvement from what I would do before, which is just to send a set of `np.ones` as the weight.

coffea/analysis_tools.py
nsmith-: I'll say its optional, just let me know if you prefer to do so or not

valsdav: I was also thinking of reducing duplication there, I will try to do it today

lgray:
I think this looks done, @nsmith- @valsdav ?

PR #753, opened 2022-12-02
Don't copy histograms in accumulate loop

lgray:
see if we can get away with not copying histograms in the accumulate loop


lgray:
not really useful.

PR #765, opened 2023-02-27
Reading L5flavour Jet Energy Correction files

AndrissP:
1) I added a feature to `extractor` to read text files with L5Flavour jet energy corrections like in [tests/samples/Winter14_V8_MC_L5Flavor_AK5Calo.txt](https://github.com/CoffeaTeam/coffea/compare/master...AndrissP:coffea:L5Flavour_extractor?expand=1#diff-4ce76762f906b327c34ffc4a5e7c12b3ec91a8473ab4b61e2f1f96b2e35e65ad)

The solution is very similar to `coffea.lookup_tools.txt_converters.convert_junc_txt_file`, the two options could be united at some later point.
The extractor reads all the corrections from the text file and stores them as <file_name><component_name>.

The test in tests/test_lookup_tools.py is updated with a necessary line to test this.

The usefulness of such an update is approved by the CMS Jet Energy Resolution and Corrections (JERC) group (@garvitaa).

2) Updated [coffea.lookup_tools.jme_standard_function.wrap_formula](https://github.com/CoffeaTeam/coffea/compare/master...AndrissP:coffea:L5Flavour_extractor?expand=1#diff-fa2085edbba2667d10eb0e5e4172edb533378865821682d4e0707fa91e556fee) to parse `^` as `**` and not as binary or.


lgray:
Ah - you have touched one of the subjects I am a bit picky on (don't worry).

I think my first comment is to @garvitaa, which is how soon will these be available in correctionlib? (no problem if you can answer this as well)

garvitaa:
Hi Lindsey,
Once these corrections are finalized, they'll be available in JSON format as well.
May I ask why you why is this relevant for the changes to coffea jme tools?

lgray:
It's relevant since we're currently going through a major upgrade over in another branch where we are switching to correctionlib for far more things anyway (since it is already supported in coffea). I'd prefer to maintain less code whose functionality is already covered in another package, especially if that code has an extremely limited period of use.

If it's coming out in correction lib relatively soon, it may be better to work from that unless you need it as part of a workflow for generating that validation?

After talking to Garvita, though, it seems the time until the correctionlib version is like two months or so.

---

Should this correction also be something that the CorrectedJetFactory knows about?

lgray:
On the `**` vs `^` issue - I would advise against doing this. It is better to make the correction file itself use the standard operators for their purposes. If someone actually wants to use bitwise-or at some point in the future for some reason, it will cause problems.

I would recreate these files with standard operator notation, rather than making a hack in code that interprets the text file.
Is that possible?

lgray:
@AndrissP ping, when you have time could you please answer these questions?

AndrissP:
Hi @lgray,
Yes, I agree that the `**` to `^` replacement was a too ad hoc hack. I will undo those changes and in case there is no other solution, I can do find-replace in my correction files. I'm not sure if those files would be centrally reproduced.

I also agree that we should implement something into `CorrectedJetFactory` I just haven't had time to look into what to do there exactly. I plan to have time for that on Friday.

lgray:
@AndrissP a better solution to the `**` vs. `^` issue occurs to me. 
Make it so you dynamically rewrite the file when you're reading it, since it is file-format specific there's no chance we screw up other corrections and it solves the downstream parsing issue.

anmalara:
If I may comment here, I'd use the pow function directly. this way it's consistent with the other files.

lgray:
@AndrissP I can take care of the CorrectedJetFactory part, probably easier for me. Ping me when you've settled the correction itself and I'll add edits to this PR!

anmalara:
> If I may comment here, I'd use the pow function directly. this way it's consistent with the other files.

HI @lgray, I answered too fast, since I didn't notice that the txt file was an old one (Run1).

Let me clarify the situation. Newer versions of the same type of file will contain the pow option.
I'd advise to change this old file, as it would be inconsisted with GTs and other repositories. What do you think? Do you still want to keep this file as example or how do you prefer to move forward?

lgray:
Please use the latest files in that case. It might be useful to add a warning if a `^` character is detected in the file to tell the user to use the latest file versions.

My later solution (to replace `^` with `**` or `pow` dynamically on reading the file) would also be OK imho since no alteration to the file format would be necessary in that case.

AndrissP:
Hi, @lgray, I am not sure what you mean by replacing dynamically on reading the file and how it is different from replacing when parsing the function that I was doing before.
Anyway, I reverted the replacement I was doing before.
I also added a version of a file with the correct syntax.
A warning if `^` is detected could be useful but @anmalara can comment on this.

lgray:
Ah - sorry I somehow interpretted the replacement as something global and not only in the context of the L5 corrections. My misunderstanding. In any case - this is better especially if the new files are using `pow`.

lgray:
@AndrissP That looks like it - is there anything else you want to do for this PR? I need to work out an issue with uproot and I'll make a release at the end of the week.

AndrissP:
@lgray 
Yes, this is it from my side! Thank you!

PR #768, opened 2023-03-08
fix unknown chunks in Weights and PackedSelection

ikrommyd:
Fix "Uknown Chunks" error because of operations that have not been computed yet on dask arrays when calling `len(dask.array.Array)` in Weights and PackedSelection.

### DIFF
--- a/coffea/util.py
+++ b/coffea/util.py
@@ -81,6 +81,7 @@ def _ensure_flat(array, allow_missing=False):
             if isinstance(array, dask.array.Array)
             else dask_awkward.to_dask_array(array)
         )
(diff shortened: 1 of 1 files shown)


lgray:
I'll need to check this - there's a problem here because if this takes significant time to perform on a 2000 file dataset then we should probably think of another way.

ikrommyd:
Actually nevermind, this is a mistake. I thought `compute_chunk_sizes()` wouldn't do anything if you actually knew the chunk sizes. But I checked and  it does rerun every single time and it can take time. I'm closing this PR.

PR #771, opened 2023-03-08
fix: use uproot.open's dict-interface to avoid colon-parsing

jpivarski:
The interpretation of the first argument to `uproot.open` is complicated by the fact that colons are interpreted as separators between the file path or URL and an object path for objects within the file. It has caused numerous issues.

  * https://github.com/scikit-hep/uproot5/issues/47
  * https://github.com/scikit-hep/uproot5/issues/79
  * https://github.com/scikit-hep/uproot5/pull/80
  * https://github.com/scikit-hep/uproot5/pull/81
  * https://github.com/scikit-hep/uproot5/issues/129
  * https://github.com/scikit-hep/uproot5/discussions/365
  * https://github.com/scikit-hep/uproot5/discussions/541
  * https://github.com/scikit-hep/uproot5/discussions/543
  * https://github.com/scikit-hep/uproot5/issues/669
  * https://github.com/scikit-hep/uproot5/pull/670

For this reason, there's another way to access a file and object:

```
"file_path:object_path" →  →  → {"file_path": "object_path"}
```

and the object path may be `None`. In this form, there's considerably less ambiguity about how to parse the file path.

It looks (to me) like _all_ of Coffea's uses of a file path in `uproot.open` would not have an object path in the same string. If this is true, then it's always safer/more robust to use the dict form, rather than the string form. This PR applies that uniformly.

I also searched for instances of `uproot.iterate` and `uproot.concatenate`, which interpret the first argument in the same way. I didn't find any. It's also possible that the `open` function was imported from `uproot` and used in an unqualified way, but I doubt it, since that would conflict with the builtin Python `open`. (I didn't see any strange `import` shenanigans, either. I'm 99.7300203936740% sure I got all the cases.)

### DIFF
--- a/coffea/lookup_tools/root_converters.py
+++ b/coffea/lookup_tools/root_converters.py
@@ -15,7 +15,7 @@ graphTypes = ["TGraphAsymmErrors", "TGraph2D"]
 
 def convert_histo_root_file(file):
     converted_file = {}
-    fin = uproot.open(file.strip())
+    fin = uproot.open({file.strip(): None})
     for path, item in fin.iteritems(recursive=True):
         if isinstance(item, uproot.ReadOnlyDirectory):
             continue
--- a/coffea/nanoevents/factory.py
+++ b/coffea/nanoevents/factory.py
@@ -92,7 +92,7 @@ class NanoEventsFactory:
                 Pass a list instance to record which branches were lazily accessed by this instance
         """
         if isinstance(file, str):
-            tree = uproot.open(file, **uproot_options)[treepath]
+            tree = uproot.open({file: None}, **uproot_options)[treepath]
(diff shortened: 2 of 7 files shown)


jpivarski:
By the way, will there be another 0.7.x release before 0.8? The [last release](https://github.com/CoffeaTeam/coffea/releases/tag/v0.7.20) was in November, and before that, there was a (roughly) monthly cadence. The 0.8 stuff is still exclusively in the [awkward2_dev](https://github.com/CoffeaTeam/coffea/tree/awkward2_dev) branch, right?

lgray:
Yeah there's some odds and ends going in that I'll release in the next two weeks or so before 0.8 (probably now 2023.3.0).
I'll very likely setup future releases to go on a cron job.

PR #783, opened 2023-04-06
Attempt at a calver monthly tag reset

lgray:
Would appreciate commentary here on if this is a good idea. :-)

### DIFF
--- a/.github/workflows/month_tag.yml
+++ b/.github/workflows/month_tag.yml
@@ -18,7 +18,7 @@ jobs:
       - name: extract tag message
         id: tagmessage
         run: |
-          echo "::set-output name=tagmessage::$(git log --pretty='format: - %s' --first-parent $(git describe --tags --abbrev=0)..HEAD)"
+          echo "tagmessage=$(git log --pretty='format: - %s' --first-parent $(git describe --tags --abbrev=0)..HEAD)" >> $GITHUB_OUTPUT
       - uses: actions-ecosystem/action-push-tag@v1
         with:
           tag: ${{ steps.prefix.outputs.date }}.0.rc0
--- a/.github/workflows/month_tag.yml
+++ b/.github/workflows/month_tag.yml
@@ -1,0 +1,25 @@
+on:
+  schedule:
+    - cron:  '0 0 1 * *'
+
+name: monthly-release
+jobs:
+  monthly-release:
+    name: monthly release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: format calver prefix
+        id: prefix
+        run: |
+          echo "::set-output name=date::$(date +'%Y.%-m')"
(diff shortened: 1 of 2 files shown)


.github/workflows/month_tag.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          fetch-depth: 0
+      - name: format calver prefix
+        id: prefix
+        run: |
+          echo "::set-output name=date::$(date +'%Y.%-m')"
+      - name: extract tag message
+        id: tagmessage
+        run: |
+          echo "::set-output name=tagmessage::$(git log --pretty='format: - %s' --first-parent $(git describe --tags --abbrev=0)..HEAD)"
```
matthewfeickert:
```suggestion
          echo "tagmessage=$(git log --pretty='format: - %s' --first-parent $(git describe --tags --abbrev=0)..HEAD)" >> $GITHUB_OUTPUT
```

`set-output` commands via stdout is deprecated and will start failing soon (https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) so use the new approach of appending to the [`$GITHUB_OUTPUT` environment file](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files).

c.f. https://github.com/scikit-hep/pyhf/pull/2155 for further examples.

matthewfeickert (changes requested):

.github/workflows/month_tag.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: monthly release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: format calver prefix
+        id: prefix
+        run: |
+          echo "::set-output name=date::$(date +'%Y.%-m')"
```
matthewfeickert:
```suggestion
          echo "date=$(date +'%Y.%-m')" >> $GITHUB_OUTPUT
```

Missed this one.

matthewfeickert (changes requested):

jpivarski:
Now would be a very good time to switch from semantic versions to calendar versions, if you prefer calendar versions.

But... _automatic_ releases? Something that happens once a month is not a lot of manual burden, and while a human in the loop doesn't guarantee that it's bug-free, it's a safety measure that prevents obviously bad things from happening. What about automated reminders and a one-button procedure instead?

Then again, maybe more users of Coffea pin versions, since they're at the end of a dependency chain (applications, not libraries).

Maybe my question is, what projects actually do this?

lgray:
@jpivarski You make some good points - an automated reminder may be a better idea. Though I find even with those I'll end up not doing something.

I'd kind of prefer an automatic monthly release, just to deal with that mode of failure. 

Even better would be something that makes a release if there are commits since the last tag.  This would deal better with the random failures I am trying to avoid with this.

lgray:
obviated by #813

PR #786, opened 2023-04-10
feat: systematics handling for dask mode

lgray:
Getting the systematics book-keeping back into working order.

@alexander-held 

The awkward / dask-awkward / dask-histogram pins + this will make the next rc.

### DIFF
--- a/binder/systematics.ipynb
+++ b/binder/systematics.ipynb
@@ -10,9 +10,15 @@
     "import awkward\n",
     "import numpy as np\n",
     "from coffea import nanoevents\n",
+    "\n",
+    "def get_array(array):\n",
+    "    return array.compute() if nanoevents_mode == \"dask\" else array\n",
+    "\n",
+    "nanoevents_mode = \"dask\"\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


lgray:
ok, was able to save that recent patch.

lgray:
@ikrommyd aside from a little bit of comment/dead code clean up this is ready for review.

ikrommyd (changes requested):
Hi @lgray, I tested the `add_systematic` locally and it seems to work and gives me the same result of dask and virtual modes. Could we have a unit test please (`test_nanoevents_systematics.py` or something)? You could just steal the `muon_pt_scale` example from the notebook and just assert that virtual and dask results are equal to the eager one :)

lgray:
Thanks, will attend to this momentarily!

lgray:
@ikrommyd Should be what you were looking for now, please have a look.

ikrommyd (approved):

ikrommyd:
Perfect! Thanks :)

lgray:
oops - there's a small confusing typo... I'll submit a quick PR

PR #788, opened 2023-04-11
fix(wq): count chunks metric fix

btovar:
A couple of small wq fixes:

- Use the count of items processed to report number of chunks in "metrics".  Previous code tried to get a len of a generator, which is not defined.
- Also, ensure that that the parent of the staging directory exists.

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1936,7 +1936,11 @@ class Runner:
             processor_instance.postprocess(wrapped_out["out"])
 
         if "metrics" in wrapped_out.keys():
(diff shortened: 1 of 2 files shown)


lgray:
Ah - this PR goes into the discussion re: TaskVine. Since the user interface we want to present to people from coffea 2023 onwards is dask-based. 

The semantics of file access, metrics, etc. are very different.

lgray:
Changes will be coming to the executors soon.

lgray:
~@btovar could you rebase this to latest master?~ I have done this for you!

btovar:
Not used anymore.

PR #789, opened 2023-04-11
Get tests in passing state, mark heisenbugs

lgray:
- revert physlite back to eager awkward for now


lgray:
@agoose77 I'll get things passing again here and merge.

PR #797, opened 2023-04-20
feat: New features to PackedSelection

ikrommyd:
Starting a pull request to merge the new features of `PackedSelection` intto the master branch.
This isn't 100% complete yet as slightly more testing and actual writing of some more unittests is required


lgray:
Small comment - for better granularity can you make the tests a bit shorter and more focused? Right now one test will fail and we won't know what's wrong without digging in!

ikrommyd:
> Small comment - for better granularity can you make the tests a bit shorter and more focused? Right now one test will fail and we won't know what's wrong without digging in!

In bd37f9f, I split the tests into "basic", "nminusone" and "cutflow" tests.

lgray:
@iasonkrom just so the tests for your stuff are really exercising a realistic workflow (i.e. reading from a file), as we have found that to be a sticking point. Could you please add some tests that do so, so we aren't blindsided in the future? It's OK if the tests fail for, you can mark them with a pytest.xfail or test for the exception/failure to happen.

ikrommyd:
@lgray the "nminusone" and "cutflow" tests do use the nano_dy samples which I read once at the top of the test file to avoid reading in every test function. However for the "dak" version of the tests I currently use `from_awkward` because of the issues in https://github.com/dask-contrib/dask-awkward/issues/256. I could convert those to xfail tests if you want.

lgray:
Yes, please do - the `from_awkward` to get your data is quite unrealistic as we discussed.

lgray:
@iasonkrom can you rebase this to master?

lgray:
You are having a bad problem and will not go to space today.

ikrommyd:
@lgray Yeah I accidentally pressed "discard commits" on github and this closed the PR automatically because this synced my branch with master. Is it possible to reopen it?

lgray:
You can just force push back to the same branch.

ikrommyd:
@lgray Ah I thought someone with your permissions on the repo could only reopen the PR. It's now open and rebased per your request. Sorry for the small hiccup due to my inexperience.

lgray:
No worries, I'm mostly glad you didn't lose the commits. :-)

ikrommyd:
@lgray Should I change the tests in this PR to not contain `NanoEventsFactory` and only read files via `uproot.dask()`.
On one hand this is a PR in `coffea`, on the other `PackedSelection` is only about boolean masks and has nothing to do with `NanoEventsFactory`.
I was just thinking that because of https://github.com/dask-contrib/dask-awkward/issues/256, maybe simplifying the tests is better.

lgray:
We should make sure we test things in the manner we intend to use them.
If that fails because of incomplete/work-in-progress features in dependencies, we must wait.

OTOH having varying levels of test complexity can help you zero in on problems.
If you feel like writing more tests, keeping the old ones, please feel free to. :-)

ikrommyd:
I think I'm gonna repeat the cutflow tests avoiding `NanoEventsFactory` and using only `uproot.dask` and also parametrize them with `awkward.optimization.enabled` set to `True` and `False` and xfail the `True` case for now. I think it's good to keep track of all the issues.

ikrommyd:
tests work as expected with `uproot` only and with awkward column optimization set to `True` (xfail) or `False` (pass)

ikrommyd:
@lgray This is almost ready to review. I merged the master branch, removed the xfails and adapted to the new `from_uproot` of `NanoEventsFactory`. The only things that fail are the cutflow tests with `NanoEventsFactory` with the known `no field 'Electron' in record with 23 fields` error only when column optimization is disabled. When it's enabled they pass fine.

lgray:
Great! The fact it fails with optimization off is intensely weird. We should get the dask-awkward guys to look at that.

ikrommyd:
Should I also parametrize the daskified `Weights` tests in analysis tools to run with and without column optimization for completeness since that's what we do for `PackedSelection`?

lgray:
@iasonkrom yes it's a good idea - we should make sure all variants function

ikrommyd:
Weights work fine in both cases. We can also do this in jetmet tools and lookup tools where there are some "_dak" tests

lgray:
Sure - go to town! :-)

ikrommyd:
@lgray From slack discussion. `PackedSelection` doesn't work with `uint64`  only in delayed mode . Works fine in eager mode. This fails:
```python
import numpy as np
import awkward as ak
import dask_awkward as dak
from coffea.analysis_tools import PackedSelection

fizz = dak.from_awkward(ak.Array(np.arange(10) % 3 == 0), 1)
buzz = dak.from_awkward(ak.Array(np.arange(10) % 5 == 0), 1)

sel = PackedSelection(dtype="uint64")
sel.add("fizz", fizz)
sel.add("buzz", buzz)
sel.all("fizz", "buzz")
```
The error is
```
TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

    numpy.bitwise_and.__call__(
        <Array-typetracer [...] type='## * uint64'>
        3
    )
```
and is caused in the return statement of this chunk of the `require` method.
```python
consider = 0
require = 0
for name, val in names.items():
    val = bool(val)
    idx = self._names.index(name)
    consider |= 1 << idx
    require |= int(val) << idx
return (self._data & consider) == require
```
This has nothing to do with this PR, as it fails even in the master branch and this PR never touched this, but perhaps it's something we can fix while merging this.

ikrommyd:
Probably fixed and tested in my last commit

src/coffea/analysis_tools.py
```diff
@@ -409,6 +414,507 @@ def variations(self):
...
+                    arr = awkward.flatten(var[mask])
+                    honecut.fill(arr, awkward.full_like(arr, i, dtype=int))
+                histsonecut.append(honecut)
+
+                for i, mask in enumerate(self.result().maskscutflow, 1):
+                    arr = awkward.flatten(var[mask])
+                    hcutflow.fill(arr, awkward.full_like(arr, i, dtype=int))
+                histscutflow.append(hcutflow)
+
+        elif self._delayed_mode:
```
lgray: just use else?

src/coffea/analysis_tools.py
```diff
@@ -409,6 +414,507 @@ def variations(self):
...
+        labels = ["initial"] + list(self._names)
+
+        if not self._delayed_mode:
+            honecut = hist.Hist(hist.axis.Integer(0, len(labels), name="onecut"))
+            hcutflow = honecut.copy()
+            hcutflow.axes.name = ("cutflow",)
+            honecut.fill(numpy.arange(len(labels)), weight=self._nevonecut)
+            hcutflow.fill(numpy.arange(len(labels)), weight=self._nevcutflow)
+
+        elif self._delayed_mode:
```
lgray: just use else?

src/coffea/analysis_tools.py
```diff
@@ -409,6 +414,507 @@ def variations(self):
...
+                    hist.axis.Integer(0, len(labels), name="N-1"),
+                )
+                arr = awkward.flatten(var)
+                h.fill(arr, awkward.zeros_like(arr))
+                for i, mask in enumerate(self.result().masks, 1):
+                    arr = awkward.flatten(var[mask])
+                    h.fill(arr, awkward.full_like(arr, i, dtype=int))
+                hists.append(h)
+
+        elif self._delayed_mode:
```
lgray: just use else?

src/coffea/analysis_tools.py
```diff
@@ -409,6 +414,507 @@ def variations(self):
...
+                Either the filename (string) or an open file (file-like object)
+                where the data will be saved. If file is a string or a Path, the
+                ``.npz`` extension will be appended to the filename if it is not
+                already there.
+            compressed : bool, optional
+                If True, the data will be compressed in the ``.npz`` file.
+                Default is False.
+        """
+        labels, nevonecut, nevcutflow, masksonecut, maskscutflow = self.result()
+        if self._delayed_mode:
```
lgray: do we want this to always, explicitly start a computation?

For example `dask_awkward.to_parquet` provides the possibility of doing its computation later, depending on the opinion of the user.

src/coffea/analysis_tools.py
```diff
@@ -409,6 +414,507 @@ def variations(self):
...
+
+        bins = [None] * len(vars) if bins is None else bins
+        start = [None] * len(vars) if start is None else start
+        stop = [None] * len(vars) if stop is None else stop
+        edges = [None] * len(vars) if edges is None else edges
+        transform = [None] * len(vars) if transform is None else transform
+
+        if axes is not None:
+            axes = axes
+        elif axes is None:
```
lgray: just use else?

src/coffea/analysis_tools.py
```diff
@@ -593,8 +1143,120 @@ def any(self, *names):
...
+            mask = self.all(*(names[:i] + names[i + 1 :]))
+            masks.append(mask)
+        mask = self.all(*names)
+        masks.append(mask)
+
+        if not self.delayed_mode:
+            nev = [len(self._data)]
+            nev.extend(numpy.sum(masks, axis=1))
+
+        elif self.delayed_mode:
```
lgray: just use else?

src/coffea/analysis_tools.py
```diff
@@ -593,8 +1143,120 @@ def any(self, *names):
...
+            masksonecut.append(mask1)
+            maskscutflow.append(mask2)
+
+        if not self.delayed_mode:
+            nevonecut = [len(self._data)]
+            nevcutflow = [len(self._data)]
+            nevonecut.extend(numpy.sum(masksonecut, axis=1))
+            nevcutflow.extend(numpy.sum(maskscutflow, axis=1))
+
+        elif self.delayed_mode:
```
lgray: just use else?

src/coffea/analysis_tools.py
```diff
@@ -557,18 +1089,36 @@ def require(self, **names):
         returns a boolean array where an entry is True if the corresponding entries
         ``cut1 == True``, ``cut2 == False``, and ``cut3`` arbitrary.
         """
+        for cut, v in names.items():
+            if not isinstance(cut, str) or cut not in self._names:
+                raise ValueError(
+                    "All arguments must be strings that refer to the names of existing selections"
+                )
+
         consider = 0
```
lgray, line 1200: you can define `consider` as the appropriate dtype up here, as well as require.

src/coffea/analysis_tools.py
```diff
@@ -488,15 +1005,13 @@ def __add_eager(self, name, selection, fill_value):
...
-        ):
+        if isinstance(selection, numpy.ndarray) and self.delayed_mode:
             raise ValueError(
-                f"New selection '{name}' is not eager while PackedSelection is!"
+                f"New selection '{name}' is not delayed while PackedSelection is!"
             )
         elif len(self._names) == self.maxitems:
             raise RuntimeError(
-                f"Exhausted all slots in {self}, consider a larger dtype or fewer selections"
+                "Exhausted all slots in this PackedSelection, consider a larger dtype or fewer selections"
```
lgray: `f"Exhausted all slots in PackedSelection: {self}, consider a larger dtype or fewer selections"`

This may be more clear in the case you have multiple packed selections floating around.

src/coffea/analysis_tools.py
```diff
@@ -439,11 +945,24 @@ def __init__(self, dtype="uint32"):
...
         return self._names
 
+    @property
+    def delayed_mode(self):
+        if isinstance(self._data, dask_awkward.Array):
+            return True
+        elif isinstance(self._data, numpy.ndarray):
+            return False
+        else:
+            return "PackedSelection hasn't been initialized with a boolean array yet!"
```
lgray: This function has a problem due to the multiple return value types. 
```python3
if "hello":
    print("world")
```
will output "world", which could lead to unexpected behavior for someone running eager code.

This should likely raise an exception or print a warning and return `False` since uninitialized is technically also not delayed mode (but also not eager).

lgray:
@iasonkrom left a number of review comments, nothing structurally wrong - just a few stylistic and clarity issues.

Before we merge this you should add a notebook with a decent amount of realism as documentation to the `binder` directory.

src/coffea/util.py
```diff
@@ -98,6 +99,25 @@ def _ensure_flat(array, allow_missing=False):
     return array
 
 
+def _getaxis(name, var, bins, start, stop, edges, transform, delayed_mode):
```
lgray: more descriptive function name here please - there are at least three kinds of axes I can think of.

ikrommyd:
@lgray I'll adress all your comments and fix them. For the `.to_npz` method, the computation needs to be initialized and if I remember correctly numpy does it while saving even if you don't call it yourself. I will therefore add a `.to_parquet` as you said.
For the notebook, as I said in slack, I will modify an existing one that I have and add it to the binder directory.

Could you please also review the tests and make sure that they test the expected output correctly? The logic there is: `assert result_from_packedselection == result_if_doing_cutflow_manually`.

lgray:
@iasonkrom I was not meaning for you to add a to_parquet! More to mimic its behavior to either delay the computation or run it when called. That's all!

src/coffea/analysis_tools.py
ikrommyd: That didn't work when I was trying it out. I had to to at the return statement after performing operations with them. It would also fail in eager mode with int64 if I did that.

lgray:
The tests are fairly comprehensive, I don't think they're missing much if anything. 
They are being run with the windows and macos tests, so everything seems fine.

ikrommyd:
@lgray do not merge yet. I may have found a problem in dask-awkward while creating the tutorial

src/coffea/analysis_tools.py
```diff
@@ -409,6 +415,507 @@ def variations(self):
...
+        Returns
+        -------
+            hists : list of hist.Hist or hist.dask.Hist objects
+                A list of 2D histograms of the variables for each step of the N-1 selection.
+                The first axis is the variable, the second axis is the N-1 selection step.
+            labels : list of strings
+                The bin labels of y axis of the histogram.
+        """
+        for name, var in vars.items():
+            if len(var) != len(self._masks[0]):
```
lgray: This will not scale well since you are asking for the len of each *complete* array for every variable linearly. 
Will be extremely slow for large datasets.

Suggest switching to https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/core.py#L2079 which is a marginally weaker test but infinitely faster.

src/coffea/analysis_tools.py
```diff
@@ -409,6 +415,507 @@ def variations(self):
...
+                raise ValueError(
+                    f"The variable '{name}' has length '{len(var)}', but the masks have length '{len(self._masks[0])}'"
```
lgray: and likewise raise an incompatible partitions error here.

src/coffea/analysis_tools.py
```diff
@@ -409,6 +415,507 @@ def variations(self):
...
+                A list of 1D histograms of the variables of events surviving each cut alone.
+                The first axis is the variable, the second axis is the cuts.
+            histscutflow : list of hist.Hist or hist.dask.Hist objects
+                A list of 1D histograms of the variables of events surviving the cumulative cutflow.
+                The first axis is the variable, the second axis is the cuts.
+            labels : list of strings
+                The bin labels of the y axis of the histograms.
+        """
+        for name, var in vars.items():
+            if len(var) != len(self._masksonecut[0]):
```
lgray: ditto about partitions as opposed to checking length of arrays.

src/coffea/analysis_tools.py
```diff
@@ -593,8 +1147,120 @@ def any(self, *names):
...
+        masks.append(mask)
+
+        if not self.delayed_mode:
+            nev = [len(self._data)]
+            nev.extend(numpy.sum(masks, axis=1))
+
+        else:
+            nev = [
+                dask_awkward.sum(
+                    dask_awkward.from_awkward(awkward.Array([len(self._data)]), 1)
```
lgray: more getting rid of `len(dask_awkard_array)` since it is eager and slow. This will be very inefficient at scale (unless the partitions are precomputed), couldn't this be expressed in a completely delayed way?

lgray: `dak.count(self._data, axis=0)` seems like it should do, no?

lgray:
The tutorial notebook is comprehensive, nice! :-)

src/coffea/analysis_tools.py
ikrommyd: I'll try to fix this one. I'm now fixing the above comments and changing the `to_npz` to mimmick the `to_parquet` behavior

ikrommyd:
@lgray Any idea how to get raising incompatible partitions to work since `compatible_partitions` doesn't detect incompatiability if `var` is sliced as I told you in slack dms?
```python
        if self._delayed_mode:
            for name, var in vars.items():
                if not compatible_partitions(var, self._masks[0]):
                    raise IncompatiblePartitions("plot_vars", var, self._masks[0])
```

ikrommyd:
And I can't find I way to get rid of this last `len` call that I want to avoid.
```python
            h.fill(
                dask_awkward.from_awkward(awkward.Array([0]), 1),
                weight=dask_awkward.from_awkward(awkward.Array([len(weight)]), 1),
            )
```
I basically just want to fill a delayed histogram with zeros, `len(weight)` times but the layouts just don't work if I just do `dak.count`

lgray:
what's the error you get if you try to do `dak.count` ?

lgray:
Also why doesn't `h.fill(dak.zeros_like(weight), weight=1.0)` work?

lgray:
for incompatible partitions can you print out the dask arrays you're getting at that stage, since we know it fails later when trying to apply a mask for exactly the same reason.

ikrommyd:
> h.fill(dak.zeros_like(weight), weight=1.0)

yup this works. I'm just stupid and forgot about it.

ikrommyd:
> for incompatible partitions can you print out the dask arrays you're getting at that stage, since we know it fails later when trying to apply a mask for exactly the same reason.

not anymore, because this was just replaced by the `h.fill(dak.zeros_like(weight), weight=1.0)`. This was the line that was raising the error at `_getitem_outer_bool_or_int_lazy_array`

lgray:
oh, fun.

lgray:
that cannot possibly be that line does not fill anything with a mask?

ikrommyd:
wait maybe I'm just missing something. It's just that the only test failure was "did not raise incmpatible error" after I changed this line.

lgray:
Honestly for this I'm willing to save it for a follow-up PR to give some time to think about it correctly.

lgray:
Oh, and I mean for the case that should raise, when it is testing for partition compatibility can you just print out the dask_awkward.Array's and we take a look at them?

ikrommyd:
Oh you mean to find the bug? If yes then I'll create a reproducer.
What do I do with pytests though that now fail due to not raising incompatible partitions? Just skip?

lgray:
yeah just comment out that part of the test for now and let's take our time to do it the right way.

ikrommyd:
I think I finalized this for now. I believe you can just check today's changes and just merge.

PR #798, opened 2023-04-21
feat: ML inference tools

yimuchen:
Adding tools to allow inference tools to be compatible with dask_awkward computations.

List of target features:

- [X] Add proper import guards to so that builds will work again.
- [X] Add triton wrapper.
- [X] Add pytorch wrapper.
- [X] Add xgboost example.
- [X] Make the example run standalone in CI
  - [X] Add xgboost test 
  - [X] Add pytorch test 
  - [X] Add triton test  

Bonus features (may push back to next new MR)

- [X] Add documentation to demonstrate the how to use the example wrappers
- [X] Add type/format checking and type conversion to triton and xgboost wrappers (there isn't a simple way to extract type conversion for torch models).
- [X] Automate dask_column listing instead of requiring user input. (:warning:  Might need additional input from dask_awkward regarding this) https://github.com/CoffeaTeam/coffea/issues/833


coffea/ml_tools/triton_wrapper.py
```diff
@@ -0,0 +1,495 @@
...
+
+# For data manipulation
+import numpy
+
+# For triton specific handling
+try:
+    import tritonclient.grpc
+    import tritonclient.http
+    import tritonclient.utils
+except ImportError as err:
```
yimuchen: What would be the best practice here? 
Just a `pass` statement?

lgray: Add a message to the exception that says what needs to be done in order to use this module and reraise!

tests/test_ml_tools.py
```diff
@@ -0,0 +1,119 @@
...
+
+            return {
+                k: ak.concatenate(
+                    [x[:, np.newaxis, :] for x in fmap[k].values()], axis=1
+                ).to_numpy()
+                for k in fmap.keys()
+            }
+
+        def dask_touch(self, jets):
+            jets.eta.layout._touch_data(recursive=False)
```
yimuchen: Currently, no matter what I do in this dask_touch method, the `dask_awkward.necessary_columns` still shows `jets.pt` as a necessary column. Not sure if this is a dask_awkward issue or wrong handling in the `triton_wrapper.__call_dask__` method.

tests/test_ml_tools.py
yimuchen, line 1: For testing the ML inference tools for CI, would the best path be: 

- Rename this file to **not** start with `test_` (excluded from the main `pytest` routine). then add a separate CI routine in the `.github/workflow/ci.yaml`. This is more inline with what  is done for `wq.py`
- Keep filename as is, and modify the standard CI routine to also install the tools required for ML inference tests. (Maybe use pytest.skip flags so that we have a centralized testing script that everyone can use as a starting point, but not run directly in CI)

yimuchen: Leaning toward the first option right now, since testing the tritonclient interface will require spinning up a triton server in CI (which is something that users wouldn't typically want)

setup.py
```diff
@@ -108,6 +108,10 @@ def get_description():
     "pyinstrument",
     "ipython",
 ]
+EXTRAS_REQUIRE["mltools"] = [
```
lgray: I would suggest splitting this up by ML framework. If someone does `pip install 'coffea[mltools]'` and they install 3GB of packages in addition to the one they want I suspect there will be expressions of unhappiness.

tests/ml_tools.py
```diff
@@ -0,0 +1,117 @@
...
+    pfcands = pfcands[idx < jets.ncands]
+    jets["pfcands"] = pfcands[:]
+
+    ak_jets = jets[:]
+    ak.to_parquet(jets, "ml_tools.parquet")
+    dak_jets = dak.from_parquet("ml_tools.parquet")
+    return ak_jets, dak_jets
+
+
+def triton_testing():
```
lgray: unless `__name__ == "__main__"` is picked up by pytest you'll eventually need to turn this into something named `test_triton` or similar for the CI to pick it up. 

If I missed something about how to hook it in let me know!

lgray:
We should also add xgboost in the first iteration of this package, there's still an enormous amount of BDT usage in the wider community.

NJManganelli:
A good dummy model for torch would be Claire’s ParticleNet based model on the EAF, “pn_demo”, which has untrained weights and is included in https://github.com/cgsavard/triton_multiuser_benchmarks

since PN has become ubiquitous for jet tagging and other examples, and is only a couple years from the cutting edge, I think it would work nicely. Also a good demonstration of a model that you can’t practically run at scale cpu-only, due to the computation time.

(edit: a separate renamed copy of it)

lgray:
I mean - mostly we need something we can run in CI :-) (but a particle net example would be great for documentation!)

tests/test_ml_tools.py
lgray: Given that these tests should be pretty fast there's no reason to not include them with the main CI?

tests/test_ml_tools.py
lgray: I'll take a look at this in the coming week and see what's up.

lgray:
Once the models get really heavy there are some tricks we can do to make the serialization of ml model itself less of a burden (basically using the `.persist()` trick I use in the corrections). This is clearly not a problem for triton, but not everyone will have that available.

tests/test_ml_tools.py
yimuchen: So basically what we would want here is to have something like `pip install .e [mltool_torch,mltool_triton]` here [1]

And maybe we will have to have `pip install tritonserver` somewhere here [2]? (Since `tritonserver` shouldn't be included as part of `ml_tools` IMO)

[1] https://github.com/CoffeaTeam/coffea/blob/master/.github/workflows/ci.yml#L61
[2] https://github.com/CoffeaTeam/coffea/blob/master/.github/workflows/ci.yml#L80

tests/ml_tools.py
yimuchen: Yeah, CI testing is probably going to take a little while for me to finialize, as I will also need to check how to properly setup a local triton server easily in Github workflows.

yimuchen:
> I mean - mostly we need something we can run in CI :-) (but a particle net example would be great for documentation!)

I think it would be nice also to have these examples live somewhere where everyone can see. Where would be the best place to put this? And I guess it also depends on the level at which we would want to promote this.

- Would we want to promote this as a standard `torch.nn.Module` that is people can call like `coffea.ml_tools.ParticleNetBase`.
- Would we place this into say `docs/source` so, that there is a common place where people can copy/modify the code?

lgray:
We have an area for notebooks, I think it would be best to add them there!

tests/test_ml_tools.py
lgray: @yimuchen I would suggest naming the extra installs like:
```
pip install coffea[tritonclient]
pip install coffea[torch]
```
i.e. the extra should just be the name of the additional python package. That, or we go another way and just tell people in an ImportError message that they should `pip install <the appropriate package>` in order to be able to use a particular sub-package.

For installing triton server for tests, we should add the appropriate dependencies into the `test` extras.

lgray:
So for xgboost we can start with:
https://gitlab.cern.ch/HiggsDNA-project/HiggsDNA/-/blob/master/higgs_dna/tools/diphoton_mva.py#L84-95
(just fill it with random numbers for the example, and link to the full example in Higgs to gamma gamma).

You can get the weights here: https://gitlab.cern.ch/HiggsDNA-project/HiggsDNA/-/blob/master/higgs_dna/metaconditions/diphoton/re_training_2022_weights2017.xgb

tests/test_ml_tools.py
yimuchen: Ah, I left the `map_partitions(opt_touch_all=True)`.... Investigating solution

yimuchen: Yes, just confirmed that removing the `opt_touch_all=True` arguments fixes the necessary column issue. However, that means if I missed a field in the `dask_touch`, dask awkward seg faults. We probably want the failure mode to be more descriptive (probably something to be filed upstream)?

lgray: Yeah - this gets a bit touchy since it's probably not the best to expose users directly to the data-touching interface.

Maybe what can be done is something like:
- a function where the user prepares data for the ML model solely through awkward operations (no to-numpy)
  - therefore no need to worry about type tracers
- from that function the user passes the *separated* stack of arrays in some dict that we stack in automated way
  - {"column_group1": [a, b, c, ...], "columns_group2":, [x, y, z, ...], ...}

Then we can do the touching on those awkward arrays which are guaranteed to be necessary?

yimuchen: In preparation for this, I've promoted the `awkward_to_numpy` -> `numpy_evaluate` -> `numpy_to_awkward` calculation routine to a new base class in the `helper.py` file. 

So what we are saying is basically we split the `awkward_to_numpy` method into a `awkward_to_stack` + `stack_to_numpy` method, where the `stack_to_numpy` method is something automatic, and also tallies the arrays to be touched for dask operation? 

Not sure how general this needs to be.

lgray: My main goal here is to hide from users an interface that's not (yet) publicly exposed and may change often. That kind of thing should be the responsibility of library developers. I think it's something we can automate to the point of hiding it from users. 

Of course, if it gets to clunky trying to hide we, we'll have to expose it. We should, however, try out a few ways of hiding it.

yimuchen: With the existing interface, an "easy" middle ground solution would probably to have a function that returns all the required fields: 

```python
dask_touch(jets): 
   return [jets.field1, jets.field2, jets.nested.field1, jets.nested.field2]
```

Then have the function in the typetrace section iterate over the function return and touch these fields. It's one more function book-keeping that is required of the user (and with dask_awkward as it is right now this, will unfortunately hard segfault if the user misses anything). If I cannot come up with an easy to interface and general solution for stacking/ak-to-np, this might be the route we need to take…

yimuchen: Changing to the have users overload a `dask_columns` method and return columns of interest, instead of having the users explicitly call `_touch_data`.

This as chosen over the `stack_to_numpy` method, as I found the `awkward.concatenate` has different behavior for normal awkward arrays and awkward arrays used for typetracer.

lgray: If you haven't already you may want to check `ak.concatenate` again with the most recent release of awkward. It changed a bit.

yimuchen: This was tested with awkward==2.1.4 and dask_awkward==main (as of today), and I am still seeing this issue. Right now, having the user implement a `dask_columns` used (can always be udpated in future updates)

lgray: OK - I'll try my hand at it but it might be a little bit before I have time.

lgray: @yimuchen For now just get tests running and xgboost in. Then we can get the finishing touches.

docs/source/ParticleNet.py
```diff
@@ -0,0 +1,289 @@
...
+    self.conv = nn.Sequential(
+        nn.BatchNorm1d(in_chn),
+        nn.Conv1d(in_chn, out_chn, kernel_size=1, bias=False),
+        nn.BatchNorm1d(out_chn), nn.ReLU())
+
+  def forward(self, x):
+    return self.conv(x)
+
+
+class ParticleNetTagger(nn.Module):
```
lgray, line 265: We shouldn't actually need the ParticleNet code in coffea right? Just a jitted/traced serialized model?

yimuchen: This might be the case, I'm still going over the examples to see the minimum files required for local testing. I'm having Claire push an example now so that I have something to work with and have this contribution be logged under her name.

lgray:
@yimuchen some of the mlpackages you want to test against are not available on all OSes (particularly triton client). You'll need to add a pytest importer skip.

.github/workflows/ci.yml
```diff
@@ -58,23 +58,42 @@ jobs:
...
+        python -m pip install xgboost
+        # install checked out coffea
         python -m pip install -q -e '.[dev,dask]'
         python -m pip list
         java -version
+
+    - name: Start triton server with example model
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        docker run  --rm   -p 8001:8001 -p 8002:8002  -v$(realpath ./coffea/tests/samples/triton_models_test):/models nvcr.io/nvidia/tritonserver:23.04-py3 tritonserver --model-repository=/models
```
yimuchen: Final testing issue! For triton testing, we will need to spin up a parallel docker instance. Throwing the command I used for local testing on my laptop here while I look up the docs on how you do this properly in GitHub workflows.

lgray:
@yimuchen there's something funny with pytorch on arm - I'll fix it up over the weekend / during chep.

Otherwise looking really good :D

lgray:
This project in particular may benefit from https://github.com/scikit-hep/awkward/pull/2127

@agoose77 will `empty_if_typetracer` touch data correctly when you call it on a typetracer?

agoose77:
> This project in particular may benefit from [scikit-hep/awkward#2127](https://github.com/scikit-hep/awkward/pull/2127)
> 
> @agoose77 will `empty_if_typetracer` touch data correctly when you call it on a typetracer?

Yes, it touches recursively.

coffea/ml_tools/torch_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+    def _create_device(self):
+        """
+        Torch device run calculations on. This wrapper class will always attempt
+        to use GPU if possible. Setting this as a lazy object so that remote
+        worker can have a different configuration the interactive session.
+        """
+        # Setting up the default torch inference computation device
+        return torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+    def _create_model(self):
```
lgray: @yimuchen so there's an issue here. It's fixable but may be slightly painful.

In particular models can be really large in memory footprint, and when running in dask mode this can incur some extremely hefty serialization penalties. My trick for things like correctionlib or other hefty corrections was to store a weak reference to the object and/or a dask future from a `dask.persist` call so that the serialization was only ever done once per worker.

This is a bit of a hack and I think it could be gotten around by figuring out how to get dask_awkward to understand which function calls are pure so they don't need to be deserialized over and over again. @martindurant @douglasdavis: commentary appreciated!

martindurant: Dask assumes that function calls are pure, so long as they hash to the same dask key. It is possible to define how keys are made for this purpose. It is also possible to configure the serialization process on a class-based dispatch mechanism.

lgray: OK - I've run into some issues then on your last point. Some examples of how to control the deserialization of functions would be super helpful. 

Already ran into this problem with `dask_histogram`, and then also `coffea.lookup_tools.lookup_base` where for every partition the histogram or correction was getting deserialized per partition, causing incredible serialization time for their first implementations. This would be a problem as well, for sure, for ML models.

The solution in dask histogram was to make `_histref` the axes and metadata as opposed to the completely rendered histogram. The solution, so far, for `lookup_base` was to exploit `dask.persist` and weakrefs.

coffea/ml_tools/torch_wrapper.py
```diff
@@ -37,7 +37,11 @@ def __init__(self, torch_model: torch.nn.Module):
 
         # Reference to the original pytorch model, loading in the state, and
         # set to evaluation mode
-        self.orig_model = torch_model
+        self._dask_model = dask.delayed(
```
lgray: @yimuchen so you need to be a bit careful here! If you persist the model then when you ask for it later it has to get it from wherever it was put! This might be on distributed compute and getting it would be slow! So you need a local copy.

What I did in lookup_base was to keep both weakref (for local evaluation) and a dask handle, so it's efficient on both sides of the fence.

lgray: You can do silly things like:
```
self._model = weakref.ref(torch_model)
self._dask_model = dask.delayed(self.model(),...)
```

yimuchen: Right, I'm still in the processes of testing this new wrapping. Is there a nice way to test for potential issues when scaling up?

lgray: The only way I was able to work through it was running a realistic work case and sticking it in a profiler.

yimuchen: Got the first working version, now attempting to add additional test routines for the dask.persist fall back method.

yimuchen: Found issue when using cluster-like operations (explicitly spawning a local cluster using `client = distribtuted.Client()`), showing the PyTorch models are not guaranteed to be pickleable. Moving torch_wrapper to use file-based lazy loading (potentially slower, but guaranteed to work with cluster, and will loaded exactly once).

lgray:
@yimuchen had a deeper look through the PR and this is looking fairly solid.

I think the documentation is necessary to stick this in the codebase though. The tests are pretty opaque. I think a fully worked out examplefor ParticleNet using PFNano (or a closely related variant) in triton and torch would suffice to start since that covers the common use cases and is varied enough to let people guess well about xgboost.

lgray:
~@yimuchen can you rebase this to master?~ Was able to merge in the pyproject migration stuff for you!

yimuchen:
> ~@yimuchen can you rebase this to master?~ Was able to merge in the pyproject migration stuff for you!

Thanks! 

> @yimuchen had a deeper look through the PR and this is looking fairly solid.
> 
> I think the documentation is necessary to stick this in the codebase though. The tests are pretty opaque. I think a fully worked out examplefor ParticleNet using PFNano (or a closely related variant) in triton and torch would suffice to start since that covers the common use cases and is varied enough to let people guess well about xgboost.

A first iteration of the example code has been given in the latest commit. Written using a randomly generate events array. I would rewrite this to use a standard PFNano events file later today.

lgray:
> Automate dask_column listing instead of requiring user input.

@yimuchen this one can come later if there is not a clear solution now, depending on direction we should open a follow-up issue and get it done eventually.

lgray:
I think I've figured out a way to accomplish the automation of `dask_column` using `empty_if_typetracer`.

lgray:
So indeed - I've found a solution but it first needs https://github.com/scikit-hep/awkward/issues/2495 to be solved.
I'd prefer to get this merged sooner rather than later, so since a way forward is identified I think we can merge this as-is, whiling opening an issue and then update it later.

Essentially what we can do is as follows:
- instead of `awkward_to_numpy` as a user defined function instead it is `prepare_awkward_to_numpy` which outputs args/kwargs such that it's the awkward arrays that can be turned into numpy arrays. i.e. all that is missing is the `.to_numpy()` call.
- then `awkward_to_numpy` becomes automated and instead of just `.to_numpy()` we do `ak.typetracer.empty_if_typetracer(arg).to_numpy(allow_missing=False)` when converting the numpy-ready arrays.
- since `ak.typetracer.empty_if_typetracer()` recursively touches the data if it's a typetracer we then do not need to specify a list!

In this way we will trace back along the input arguments to determine the precise list of necessary data without the user having to supply any hinted list of data to touch.

It will also error in a reasonable way if the shape isn't something that can be converted into numpy, so that appears to be in decent shape already.

yimuchen:
Cool! I'll see if I can complete the MR tomorrow with a proper PFNano root file in the example (requires a bit of work getting a file). If we get this done before https://github.com/scikit-hep/awkward/issues/2495, I guess we can merge this first, and I'll follow up when the awkward update is complete.

lgray:
https://github.com/scikit-hep/awkward/issues/2496 would also significantly simplify the implementation, this improvement can be yet another issue to follow up on once available.

lgray:
@yimuchen here's an example of the automated type-tracing working: 
https://github.com/lgray/coffea/commit/2ca652b7f82993c9ac3e6173767c20caaaec5250

The `test_torch()` test passes if you use the awkward branch https://github.com/scikit-hep/awkward/pull/2497 locally!

There are a few things to clean up around it in your code, I'll leave that to you as well as the final implementation for when that PR in awkward is released.

Furthermore if https://github.com/scikit-hep/awkward/issues/2496 gets implemented quickly there's even **less** code to write and some of the functions to deal with creating various shapes go away. Pretty nice!

Update: https://github.com/scikit-hep/awkward/pull/2499 is already in the pipeline.

yimuchen:
Lastest documentation requires #832 to be merged to run.

src/coffea/ml_tools/torch_wrapper.py
```diff
@@ -0,0 +1,81 @@
...
+        Creating the model instance of the using either the weakref or the
+        dask.persist instance (prioritizes using weakref). Enforcing the eval
+        mode after initialization.
+        """
+        if torch.cuda.is_available():
+            model = torch.jit.load(self.torch_jit).cuda()
+        else:
+            model = torch.jit.load(self.torch_jit)
+        model.eval()
+        return model
```
lgray, line 63: The way this is implemented currently it will deserialize the entire model for each chunk (when N_chunks is large). Significant inefficiency. Would prefer it rolls out without scaling issues but we can also fix it in another PR.

Ditto for xgboost. Triton is fine for understood reasons. :-)

lgray: However, I think this one can also come later.

.cirrus.yml
```diff
@@ -19,6 +19,11 @@ task:
       install_script: |
         . ../venv/bin/activate
         python -m pip install --upgrade pip setuptools wheel
+        # mltool installs
+        # python -m pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
+        python -m pip install xgboost
+        # python -m pip install tritonclient[grpc,http]
+        # install checked out coffea
```
nsmith-, line 28: The commented out installs are for things we don't regularly test? Or what is the story

binder/mltools.ipynb
nsmith-, line 1: In your example you should be able to simply use `events.Jet.constituents` to get the candidates:
https://github.com/CoffeaTeam/coffea/blob/f8a4eb97137e84dd52474d26b8100174da196b57/src/coffea/nanoevents/methods/nanoaod.py#L416

nsmith-: Also `ak.num(jets)` instead of `ak.count(jets.pt)` for `njets`

src/coffea/ml_tools/helper.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def _call_numpy(self, *args, **kwargs):
+        """
+        Thin wrapper such that the validate_numpy_inputs is called.
+        """
+        self.validate_numpy_input(*args, **kwargs)
+        return self.numpy_call(*args, **kwargs)
+
+    @abc.abstractclassmethod
+    def numpy_call(self, *args, **kwargs):
```
nsmith-, line 173: Are these abstract bound methods or abstract class methods? I suspect the former since you are using `self` rather than `cls` for the first argument. In which case you want the `@abc.abstractmethod` decorator.

src/coffea/ml_tools/helper.py
```diff
@@ -0,0 +1,356 @@
...
+        Highest level abstraction to be directly called by the user. Checks
+        whether the inputs has any awkward arrays or dask_awkward arrays, and
+        call the corresponding function if they are found. If no dask awkward or
+        awkward arrays are found, calling the underlying _call_numpy method.
+        """
+        all_arrs = [*args, *kwargs.values()]
+        if any(isinstance(arr, awkward.Array) for arr in all_arrs):
+            return self._call_awkward(*args, **kwargs)
+        elif any(isinstance(arr, dask_awkward.Array) for arr in all_arrs):
+            return self._call_dask(*args, **kwargs)
```
nsmith-: Wouldn't we want to preferably dispatch to dask_awkward if any array is a dask_awkward array?

src/coffea/ml_tools/torch_wrapper.py
nsmith-: Can you run a quick test to confirm this?

src/coffea/ml_tools/triton_wrapper.py
```diff
@@ -0,0 +1,300 @@
...
+
+    @property
+    def client_args(self) -> Dict:
+        """
+        Function for adding default arguments to the client constructor kwargs.
+        """
+        if self.protocol == "grpc":
+            kwargs = dict(verbose=False, ssl=True)
+        elif self.protocol == "http":
+            kwargs = dict(verbose=False, concurrency=12)
```
nsmith-: Would be good to make 12 a class static member. Can you add some discussion on the value?

src/coffea/ml_tools/triton_wrapper.py
```diff
@@ -0,0 +1,300 @@
...
+                ][0]
+            elif "max_batch_size" in model_config:
+                self._batch_size = model_config["max_batch_size"]
+            else:
+                warnings.warn(
+                    "Batch size not set by model! Setting to default value 128. "
+                    "Contact model maintainer to check if this is expected",
+                    UserWarning,
+                )
+                self._batch_size = 10
```
nsmith-: Warning and value don't match. Can make this a class static member and use it in the string

src/coffea/ml_tools/xgboost_wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class xgboost_wrapper(numpy_call_wrapper, lazy_container):
+    """
+    Very simple wrapper for xgbooster inference. The xgboost.Booster object is
+    setup as a lazy object.
+    """
+
+    def __init__(self, fname):
+        lazy_container.__init__(self, ["xgbooster"])
```
nsmith-: After seeing a few uses of `lazy_container` I wonder if a better name would be `nonserializable_attributes` or something?

tests/test_ml_tools.py
```diff
@@ -0,0 +1,217 @@
...
+    # dask_awkward arrays
+    NFEAT = 100
+    jets = ak.zip(
+        {
+            "pt": ak.from_numpy(np.random.random(size=njets)),
+            "eta": ak.from_numpy(np.random.random(size=njets)),
+            "phi": ak.from_numpy(np.random.random(size=njets)),
+            "ncands": ak.from_numpy(np.random.randint(1, 50, size=njets)),
+        },
+        with_name="LorentzVector",
```
nsmith-: The `with_name` here doesn't do anything because you don't also set `behavior=coffea.nanoevents.methods.vector.behavior`. Which is good because `LorentzVector` mixin expects `x,y,z,t` attributes and this wouldn't work without using `PtEtaPhiMLorentzVector` as the mixin.

nsmith- (changes requested):
This will save a lot of ML boilerplate for coffea users and hopefully prevent a few gotchas in user code, so it's a good addition!

binder/mltools.ipynb
yimuchen: `events.Jet.constituent` only has the `pt` fields (I believe it maps to the `JetsPFCands_*` branches [1], and doesn't automatically carry over the `PFCands` information), we need a vector-like fields (`eta` and `phi`) to carry out the later computations.

[1] https://annika-stein.web.cern.ch/PFNano/AddDeepJetTagInfo_desc.html#JetPFCands

src/coffea/ml_tools/helper.py
```diff
@@ -0,0 +1,358 @@
...
+        return self.numpy_to_awkward(np_rets, *args, **kwargs)
+
+    def dask_columns(self, *args, **kwargs):
+        """
+        Given the same inputs as _call_dask, return a list of columns that is
+        required for the awkward_to_numpy conversion to complete. If you return
+        a None object, then the default method will kick in (touch everything
+        recursively) which has potential performance penalties
+        """
+        pass
```
nsmith-: Is the need for this due to some missing features in dask_awkward's `necessary_columns` utility? I would have expected that we could trace awkward_to_numpy properly, obviating this.

src/coffea/ml_tools/helper.py
yimuchen: Right now, awkward and dask_awkward cannot be mixed (`x = dask_awkward.Array([...]); y= awkward.Array([...]); x+y ` will fail). So this is implantation effectively is which ever it finds first (and fails as usual if mixed arrays are passed in)

src/coffea/ml_tools/helper.py
yimuchen: 3rd issue listed here: https://github.com/CoffeaTeam/coffea/issues/833

I think Lindsey is currently working this.

lgray: Yeah - forgot to remove this - with the latest awkward and the most recent commits in this PR this interface surface is no longer needed.

binder/mltools.ipynb
nsmith-: Ah sorry I forgot there's one extra layer of indirection: `events.Jet.constituents.pf`. That's because the constituents have some info that is only in relation to the jet (e.g. btag variables). So the first layer returns https://github.com/CoffeaTeam/coffea/blob/f8a4eb97137e84dd52474d26b8100174da196b57/src/coffea/nanoevents/methods/nanoaod.py#L545 which then has the `.pf` to get the actual PF candidate.

lgray:
I'll fix the triton test and remove `dask_columns` in a bit.

.cirrus.yml
lgray: oh tritonclient/torch/xgboost doesn't have wheels for some OSes, commented. I'll just remove them.

src/coffea/ml_tools/helper.py
lgray: Then we need to harden this a little bit, if you detect awkward.Array (or np.ndarray), then make sure it's *only* awkward arrays that are the other array types. If dask_awkward.Array, *only* dask_awkward.Array are the other array types.

.cirrus.yml
lgray: Actually I'm going to leave them here commented in case wheels become available. I'll add a comment to clarify.

src/coffea/ml_tools/triton_wrapper.py
yimuchen: This value of 12 is a "given" for the LPC triton server. I'll see if I can track down why it was set to this value.

lgray:
@yimuchen if you can update the notebook to remove mentions of `dask_columns` and associated error modes I would appreciated it. :-)

lgray:
ok - some further problems with the awkward/numpy boundary - will fix in a moment.

lgray:
@yimuchen with this I should be done messing around with the PR please go ahead and attend to the rest!

binder/mltools.ipynb
yimuchen: Ah, OK! Thanks! I was thinking whether I should make a PR to PFNanoAODSchema to get the PF candidates. Nice to know it is already inside.

lgray:
Drat - the latest awkward re-introduced a bug on column touching that seems to be causing quite some trouble in terms of getting CI done.

I'll deal with that, we can clean the PR up in the mean time - I don't consider it a blocker.

yimuchen:
Requires fix mentioned in this PR: https://github.com/CoffeaTeam/coffea/pull/835

lgray:
@yimuchen #835 is merged please re-enable the dask notebook stuff. :-) (be careful though some of the data touching might look funny with things in flux in dask-awkward/awkward)

tests/test_ml_tools.py
yimuchen: Removing, as Lorentz-Vector structure is not strictly needed for testing.

binder/mltools.ipynb
yimuchen: Using PFNano exposed a quirk with wrapping entire nanoevents in a map_parition method. Had to make slight changes to the interface. Now this is done!

lgray:
This is looking very good now that we figured out the proper way to factorize the interface.

@nsmith- if it looks good to you I'm happy with it!

nsmith- (approved):
🚀

PR #803, opened 2023-05-03
Check for `Weight.partial_weight()`

rkansal47:
Small edit adding a check that the `include` and `exclude` inputs are list. Adding this to avoid cases where if you pass a string as an input to include/exclude a single weight (as I was doing accidentally 😢) it treats the string as a set of characters.

### DIFF
--- a/coffea/processor/helpers.py
+++ b/coffea/processor/helpers.py
@@ -115,9 +115,9 @@ class Weights:
 
         Parameters
         ----------
-            include : list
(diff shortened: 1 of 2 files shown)


lgray:
numba 0.57 is broken on arm right now - otherwise this looks fine.

@rkansal47 do you need this backported into coffea 0.7?

rkansal47:
No, I personally don't need it backported.

PR #811, opened 2023-05-13
fix: Allow lookup_base args to be in any order

### LINKED ISSUE #810: Use inspect.bind in lookup_base
Follow implementation here: https://gist.github.com/agoose77/9fc1938bf15d1dedb0bd72a4cd501ed1

lgray:
also: use proper client context on workers when using latest dask

Fixes #810.

Turns out bind was just the idea of what I wanted to do rather than the implementation @agoose77.
Probably means I want to re-engineer lookup_base at some point to be less strange.

### DIFF
--- a/coffea/lookup_tools/lookup_base.py
+++ b/coffea/lookup_tools/lookup_base.py
@@ -8,10 +8,21 @@ import numpy
 
 
 def getfunction(
-    args, thelookup_dask=None, thelookup_wref=None, __pre_args__=tuple(), **kwargs
+    args,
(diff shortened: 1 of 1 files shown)


lgray:
@martindurant This is what I would like to replace, since it's a workaround for repeated deserialization of possibly very heavy correction functions. If I could instead encode the necessary metadata in a way that doesn't get deserialized every partition the correction/lookup data everything would be significantly cleaner!

Finding a more pleasant solution to this would also help out #798.

PR #819, opened 2023-05-16
build: move to pyproject style setup

### LINKED ISSUE #784: Migrate coffea 2023 to pyproject.toml
@matthewfeickert @kratsg 

coffea python modernization chores.

lgray:
Fixes #784 

Needs a few things smoothed over but it starts to work locally.


lgray:
Hrmm editable installs aren't working... @matthewfeickert @martindurant?

agoose77:
I think this is, strangely enough, a bug in the `hatchling` build backend: https://github.com/pypa/hatch/issues/860

ofek:
Screenshot 2023-05-19 112603

lgray:
@agoose77 I still don't really understand why, for instance, this works for dask-awkward but not for me? It's rather strange that the project setups are basically the same but it's just failing in my case. :shrug:

PR #821, opened 2023-05-19
chore: update to empty_if_typetracer

lgray:
awkward 2.2.1 updates


agoose77:
@lgray this should be mergeable now?

lgray:
@agoose77 yes - but I need to make sure I've gotten all the spots its needed. There's some funny business in jetmet tools I need to double check.

lgray:
There are a few other instances of `form.length_zero_array` but they don't make sense re: empty_if_typetracer, so no need to convert them.

PR #822, opened 2023-05-23
feat(wip): edm4hep development 

jbrewster7:
A schema for edm4hep files in nanoevents.

It's very similar to Delphes / Treemaker FWIW.

- [x] cover all four-vector types in edm4hep (done for PFOs, genparticles, what else?)
- [x] association maps between gen particles and reconstruction objects (clusters, PFOs)
- [ ] association maps between PFOs and clusters / tracks

@jbrewster7 we should keep a list here of the things we need to do unless we think we've found a good feature set!


lgray:
@jbrewster7 don't worry about the failures on arm, they're intermittent (probably depending on the host machine the CI job lands on)

lgray:
@jbrewster7 don't worry about the failures on arm, they're intermittent (probably depending on the host machine the CI job lands on)

coffea/nanoevents/schemas/edm4hep.py
```diff
@@ -0,0 +1,199 @@
...
+      - The "PrimaryVertices" branch in a TreeMaker Ntuple branch stores
+        'XYZPoint's corresponding to the coordinates of the primary vertices, The
+        resulting collection after this first step wold contain the coordinate
+        variables in the form of PrimaryVertices.x, PrimaryVertices.y,
+        PrimaryVertices.z.
+
+    - Extended quantities of physic objects are stored in the format
+      <Object>_<variable>, such as "Jets_jecFactor". Such variables will be
+      merged into the collection <Object>, so the branch "Jets_jetFactor" will be
+      access to in the array format as "Jets.jecFactor". An exception to the
```
tmadlener: In "podio speak" these are *`VectorMembers`* and they are used in several of the datatypes of [edm4hep](https://github.com/key4hep/EDM4hep/blob/1145e20b9ee238e1110f643232cabd061646582e/edm4hep.yaml).

Note that there are also *`OneToOneRelations`* and *`OneToManyRelations`* which are currently still stored using the convenction `<collection-name>#<index>`. These are the branches you would need to use to resolve e.g. mother/daughter relations for MCParticles. The `<index>` is currently somewhat "magic" but **this is going to change** with AIDASoft/podio#405, where the new schema of the branch names will be `_<collection-name>_<name>` for both `VectorMembers` and the relations. Let me know if you want to wait for that or whether you want some information about the magic.

I am not sure how general you want to make this in the end, but we could potentially help with some code generation on the podio side if you think that would be in any way useful.

lgray: @tmadlener Thanks for the info. Sorry, the documentation is for another dataformat type - I'll clean it up next week. In any case I can always find the constituents of a given physics object and wrap them together. It was pretty clear what to do in edm4hep!

So far as code generation - once we work it out here it could be nice to move it properly to `podio` (though that's not a python package but rather a generator for one)... Could envision some CI to put something on PyPI. Though that may not be well advised since what's being generated here is the schema as opposed to the implementation (which in our case is handled by more general functions + an embedded domain specific language that does know how to handle cross references if you give it the source/mapping/target). It may be best that it stays here, it only needs to be done once anyway since it gets distributed with coffea? We can work it out in time.

I think in the mean time if you could help us with the "magic" cross references so we know what we're supposed to be referencing that would be really useful! We can add in the better-named xref branches when they're there and files with it are available.

tmadlener: I agree on the code generation. Let's see how this works out and how frequent changes would need to be.

So the "magic" is effectively that the indices get generated sequentially in the order the relations appear in the yaml definition file, **starting with `OneToManyRelations` and then continuing with `OneToOneRelations`**. So as an example the [`edm4hep::ReconstructedParticle`](https://github.com/key4hep/EDM4hep/blob/1145e20b9ee238e1110f643232cabd061646582e/edm4hep.yaml#L409-L437) has
```yaml
    OneToOneRelations:
      - edm4hep::Vertex          startVertex    //start vertex associated to this particle
      - edm4hep::ParticleID      particleIDUsed //particle Id used for the kinematics of this particle
    OneToManyRelations:
      - edm4hep::Cluster               clusters     //clusters that have been used for this particle.
      - edm4hep::Track                 tracks       //tracks that have been used for this particle.
      - edm4hep::ReconstructedParticle particles    //reconstructed particles that have been combined to this particle.
      - edm4hep::ParticleID            particleIDs  //particle Ids (not sorted by their likelihood)
```

and on file the branches will be (assuming the collection name is `"reco"`)
- `reco#0` for the related Clusters
- `reco#1` for the related Tracks
- ...
- `reco#4` for the (single) startVertex
- `reco#5` for the (single) particleIDUsed

lgray: Indeed that's a little bit hacky to keep working since there's no clear way to keep track of the schema over time.

The next version sounds much more tractable in the long run!

lgray:
@jbrewster7 how are things going w.r.t. cross references?

jbrewster7:
@lgray Hi! The cross references seem to all be working well and I'm trying to set them up in a way that makes them easy enough to change when they go from magic numbers to names

tmadlener:
Hi @jbrewster7, if you want to make this transparent, it should be enough to use some information that is available from the `podio_metadata` Tree in the `XXX___idTable` branch(es). This contains a [`podio::CollectionIDTable`](https://github.com/AIDASoft/podio/blob/master/include/podio/CollectionIDTable.h) for which we generate a dictionary. Internally it is effectively two vectors, one containing the (numerical) IDs and the other the (string) names. I think it should be possible to get to the `m_collectionIDs` and `m_names` leaves in the branches directly. With those
```python
name_to_id = {n: i for (n,i) in zip(names, collectionIDs)}
```
should effectively give you a map of names to IDs. This will also hold after we have switched to something more robust than we currently have.

The `XXX` above is most likely always `events` for coffea, but podio can in principle also write other categories. I am not entirely sure if it makes sense at the moment to make this more general than it needs to be.

lgray:
@tmadlener I'll poke around with that table - I remember uproot complaining about it. If it won't read I'll make a PR to uproot to get it happening.

lgray:
@jbrewster7 Hooray! That's awesome!

lgray:
@tmadlener ah, that was something else, the idTable is easily accessible:

```python3
>>> x["podio_metadata"].keys()
['events___idTable', 'events___idTable/m_collectionIDs', 'events___idTable/m_names', 'events___CollectionTypeInfo', 'events___CollectionTypeInfo/events___CollectionTypeInfo._3', 'events___CollectionTypeInfo/events___CollectionTypeInfo._2', 'events___CollectionTypeInfo/events___CollectionTypeInfo._1', 'events___CollectionTypeInfo/events___CollectionTypeInfo._0', 'PodioBuildVersion', 'PodioBuildVersion/major', 'PodioBuildVersion/minor', 'PodioBuildVersion/patch', 'EDMDefinitions', 'EDMDefinitions/EDMDefinitions._1', 'EDMDefinitions/EDMDefinitions._0']
>>> x["podio_metadata"]["events___idTable"].array()[0].show()
{m_collectionIDs: [1, 2, 3, 4, 5, 6, 7, 8, ..., 40, 41, 42, 43, 44, 45, 46, 47],
 m_names: ['AllCaloHitContributionsCombined', ..., 'RecoMCTruthLink']}
 ```
 
 A bit annoying we can't just do this with the awkward form alone though. I'll need to add in a hook to nanoevents to let the schema get organizational data that's available in the file!
 
 @nsmith- thoughts?

**Comment by lgray** [comment]
@jeyserma FYI

File: coffea/nanoevents/methods/vector.py [hunk 2]
```diff
@@ -746,6 +748,251 @@ def nearest(
         return out
 
 
+@awkward.mixin_class(behavior)
+class LorentzVectorM(ThreeVector):
```
nsmith-: You should double-check, but I think all that's needed is to subclass LorentzVector and implement two overrides:
```
@property
def t(self):
    return numpy.sqrt(self["mass"]*self["mass"] + self.rho2)

@property
def mass(self):
    return self["mass"]
```
basically the magic is that `getitem` always gets what's actually in the awkward array and all `getattr` is for derived or almost-derived. That way all the other methods should pick the right override

coffea/nanoevents/factory.py
```diff
@@ -291,6 +297,12 @@ def from_root(
             elif schemaclass is TreeMakerSchema:
                 from coffea.nanoevents.methods import base, vector
 
+                behavior = {}
+                behavior.update(base.behavior)
+                behavior.update(vector.behavior)
+            elif schemaclass is EDM4HEPSchema:
```
nsmith-: This is following the existing pattern, so it's a question for @lgray: why not use the class property `EDM4HEPSchema.behavior` here?

coffea/nanoevents/methods/vector.py
lgray: This needs cleanup in a separate PR yo harmonize the interface with the eager one. You're right but marking it for later.

coffea/nanoevents/factory.py
lgray: This needs cleanup in a separate PR to harmonize the interface with the eager one. You're right but marking it for later.

nsmith-:
On the issue of needing additional metadata to properly reconstruct the NanoEvents schema (i.e. awkward form) from the list of branches, the only issue I see is that, in the case of ROOT files, there is no standard way of storing it. So each file type will need its own. Perhaps what can be done is to add a new hook in the schema class interface: a classmethod that takes in the TFile and does what it needs to find the rest of the metadata. But this of course is only for ROOT, since parquet can mostly capture awkward data types and there is a standard spot (iirc) to put the mixin behavior names. So `NanoEventsFactory.from_root` would call this method. It breaks a bit the separation between the source mapping and the events object, but I think we don't have as strong a reason to keep that around long-term.

lgray:
@jpivarski given @nsmith-'s comment (which is where I had more or less landed as well) - is there a reasonable way to get additional serialized metadata that's within the (first) root file (even if it is globbed) as a tree or class, when calling `uproot.dask`?

I suppose I have the file string, but it might be globbed, which I can't use in `uproot.open` reliably. I guess I could take the same string, `uproot.dask` it with the known location of the metadata and take the first partition only for a compute? 

That should be sufficiently quick.

lgray:
Indeed it's pretty easy:

```python3
import uproot

x = uproot.dask({"~/Downloads/rv02-02.sv02-02.mILD_l5_o1_v02.E250-SetA.I402004.Pe2e2h.eR.pL.n000.d_dstm_15090_*.slcio.edm4hep.root": "podio_metadata/events___idTable"})
```

```
(coffea-dev) lgray@dhcp-131-225-173-191 coffea % python -i edm4hep_idTable.py
>>> x.partitions[0].compute()
<Array [{m_collectionIDs: [...], ...}] type='1 * {m_collectionIDs: var * in...'>
```

jpivarski:
`uproot.dask` does open the first file in order to know the names and types of TBranches, so the metadata from the first file should be available. It might even be public (no underscore) because Uproot exposes a lot of its inner workings.

lgray:
Hey, @jbrewster7, just to check: how are things coming along with the cross references. I think whatever you have now is fine and then I'd be happy to take that and mutate it to use this stuff I was talking about with @nsmith- and @jpivarski.

However, I need to see what you've got first to act on that! There's no need to hone it to perfection, just good enough, going further is really more a collective process.

lgray:
So yeah @nsmith- it looks like I can instead just take the `filename` and `objectpath` that's passed into `.from_root` and just ask for the appropriate ID table using uproot.dask on the first partition.

Should be enough, and there's only one additional file open on the user side for a small amount of data to get the necessary bits of metadata.

lgray:
@jbrewster7 so I've been able to get cross references working as they do in CMS NanoAOD nanoevents.
I had to implement a new function `_apply_global_mapping` (that is a little delicate right now) which does the following things (and draws heavily from your first attempt, that was very far along towards an answer!):
- creates the sparse mapping between the "source" (mapping from) and the "target" (mapping on to), where your code started from!
- creates a new dense index that defaults all entries not in the sparse index to `-1` so they show up as `None`s in arrays using an IndexedOptionArray.
- repacks the sparse mapping into the dense mapping
- creates the new cross-reference to the target 

I have implemented this for PandoraPFOs and MCParticlesSkimmed, and it should extend to other collections reasonably cleanly. 

Some examples of usage are:
```python3
>>> events.PandoraPFOs[abs(events.PandoraPFOs.pdgId) == 13].matched_gen.pt.compute()
<Array [[[38.4], [56.3]], ..., [[...], ...]] type='9400 * var * option[var ...'>
```
and similarly:
```python3
>>> events.MCParticlesSkimmed[abs(events.MCParticlesSkimmed.pdgId) == 13].matched_pfos.pt.compute()
<Array [[None, None, ..., None, None], ...] type='9400 * var * option[var *...'>
```

The next step is to extend to the other collections. It would be interesting to see if we could get this to track -> hit -> sim matching or similar! Perhaps in time.

@tmadlener You may want to try this out, it is pretty neat. Note it only works on old-style edm4hep, not the new patches with fewer magic numbers. We'll finish the magic-numbers version and then work on that!
@nsmith- The `_apply_global_mapping` code is painfully fragile, I would appreciate some thoughts on how to clean it up.

Of course this can all wait until the 20th to follow up. I mostly just wanted to get this out of my head.

jbrewster7:
Hi @tmadlener, we successfully have some one-to-one relations working well with Coffea and we are now trying to interpret some of the one-to-many relations (specifically the parents and daughters of MCParticles). We've tried to interpret them similarly to the one-to-one relations, but the physical processes in the parents and daughters don't make sense. Could you point us to some documentation for how the one-to-many indices are actually mapping? Or possibly give a brief explanation here?

lgray:
@jbrewster7 could you give a screenshot of something not making sense?

jbrewster7:
@lgray Here is a picture of the beginning of the first event in the file written out. 

IMG_0422
IMG_0423


The first array is the pdgIDs of the first 20 particles in the event, and the second and third arrays are the MCParticles one-to-many relations `parents` and `daughters` respectively. We've tried interpreting this in a couple of ways:

1. The particle at the value in the `parents` array is the parent of the particle at the value in `daughters` array for the same index in the `parents` and `daughters` arrays. In the example, this would mean the electron at index 0 in the particle array goes to the electron at index 2 which goes to the electron at index 4 which goes to a higgs and a muon pair. This seems to be pretty wrong right away and there is also the problem that other one-to-many relations (such as the ones for RecoParticles) don't have two arrays of the same length.
2. The value of an item in the `parents` array is the index of the parent of the particle situated at the index of that value in the `parents` array. This is a bit confusing to write out but an example would be: looking at item number 8 of `parents` array would tell us that particle number 4, the electron, is the parent of particle number 8, the muon. This has also led to problems, mainly not being able to contain all of the information if there are multiple parents, and getting weird circular things when looking at the `daughters` array with the `parents` array.
3. The `parents` and `daughters` arrays are flattened arrays that would be vectorized in c++. This is what we believe is the most likely case as by looking at the edm4hep source code we could tell that in the c++ class these relations are essentially vectors of pointers. However, we are not sure how these are getting flattened in the root files and read in by python or how to read them in without using the c++ edm4hep class structure.

tmadlener:
Hi @jbrewster7, I am afraid the only "documentation" we have for things concerning how we store things files at the moment is the templates from which we generate the code which is probably pretty useless for anyone who is not already quite familiar with it. Let me try to explain what we do here in a mix of c++ and python as well as some text.

Similar to the on-to-one relations we only store a flat `vector<podio::ObjectID>` per collection. Resolving these for the individual objects works the same for one-to-one and one-to-many relations, i.e. you figure out the collection via the `collectionID` and then use the `index` into that collection to get the related object. The main difference between the two is that for one-to-one relations the `vector<podio::ObjectID>`  has the same length as the original array, while for one-to-many relations it effectively has an arbitrary length. (From what I gather from your questions and comments, you already have this figured out, I just repeat it here for completeness).

We store the information of which related object belongs to which object in `<relation_name>_begin` and `<relation_name>_end` members in the [`Data`](https://github.com/AIDASoft/podio/blob/c4e11bc7c662cc13694c006a0e60425823a37c47/python/templates/Data.h.jinja2#L21-L24) objects that we store, e.g. for the `MCParticle`, what is written to the root file is a `vector<MCParticleData>` with
```cpp
struct MCParticleData {
  // all the other members
  
  unsigned int daughters_begin;
  unsigned int daughters_end;
  unsigned int parents_begin;
  unsigned int parents_end;
};
```
these `begin` and `end` indices are the usual semi-open range of `[begin, end)` and denote all the related objects that belong to a given object (from the flattened vector with the `ObjectID`s in it). In more concrete terms using the daughters of MCParticles (using the new file format in this case, but that effectively just changes a few branch names in this case):
```python
import uproot as uproot

branches = [s.format("MCParticlesSkimmed") for s in ("{}/{}.daughters_begin", "{}/{}.daughters_end", "_{}_daughters.index")]
events = uproot.open("rv02-02.sv02-02.mILD_l5_o1_v02.E250-SetA.I402004.Pe2e2h.eR.pL.n000.d_dstm_15090_0.edm4hep.root:events").arrays(branches)

dau_idx = events["_MCParticlesSkimmed_daughters.index"][0]
# [2, 3, 2, 3, 4, 6, 5, 7, 8, 9, ..., 187, 188, 189, 190, 191, 188, 189, 190, 191]

dau_begin = events["MCParticlesSkimmed/MCParticlesSkimmed.daughters_begin"][0]
# [0, 2, 4, 6, 8, 11, 14, 14, 14, ..., 215, 215, 215, 215, 215, 215, 215, 215]

dau_end = events["MCParticlesSkimmed/MCParticlesSkimmed.daughters_end"][0]
# [2, 4, 6, 8, 11, 14, 14, 14, 15, ..., 215, 215, 215, 215, 215, 215, 215, 215]
```

In this case the first MCParticle has two daughters (`end - begin == 2`) and looking into `_MCParticlesSkimmed_daughters.index` array we can see that these are at index `2` and `3` of the `MCParticlesSkimmed` array. By the same logic we find that the second mc particle has the same daughters. Two more slightly interesting examples that can be seen from above are the 5th particle, for which:
```python
dau_begin[4], dau_end[4]
# (8, 11)
dau_idx[8:11]
# [8, 9, 10]
```
i.e. it has 3 daughters, and for the 8th particle, for which:
```python
dau_begin[7], dau_end[7]
# (14, 14)
```
so it has no daughters at all.

With these indices you should then be able to do the same resolving as with the one-to-one relations. I hope this hasn't been to confusing. Let me know if you have further questions or need to clarify some points.

jbrewster7:
Hi @tmadlener, this is fantastic thank you! We were looking into the begin/end objects but couldn't quite figure out how to relate them. Thank you so much for the help!

lgray:
@kjvbrt

lgray:
@jbrewster7 I started working on the notebook you sent me with various problems. All look rather surmountable - I'll push to this branch with necessary additions! I'll send you back the notebook when it's as wrapped up as it can be.

lgray:
@tmadlener could you produce some new ZH files in the new EDM4HEP format (with the more human-readable index branch names)? In addition, could you make one that's ~40 events so that we can have a test file in our repo? Thanks!

lgray:
@jbrewster7 could you fix up the flake8 errors? (click red x next to `pre-commit.ci - pr`). Looks like some improperly formatted strings.

jbrewster7:
@lgray Yes, committing the changes now!

tmadlener:
@lgray apologies for the slight delay. New files (and one with 40 events) are on cernbox: https://cernbox.cern.ch/s/eYmuXTRimfgmdZg

The 40 event one is produced from one using
```bash
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh
```

and the following lines of python
```python
from podio.root_io import Reader, Writer

reader = Reader("rv02-02.sv02-02.mILD_l5_o1_v02.E250-SetA.I402004.Pe2e2h.eR.pL.n000.d_dstm_15090_0.edm4hep.root")
events = reader.get("events")

writer = Writer("output.edm4hep.root")
for i in range(40):
    writer.write_frame(events[i], "events")
```

(There will be a few warnings about some missing schema evolution which can be safely ignored)

lgray:
Awesome, thanks!

lgray:
@tmadlener FYI https://pypi.org/project/podio/ is not taken. It would be nice if it was pip installable.

lgray:
TODO: update to interface changes in dask-awkward, build tests.

davidlange6:
hi @lgray, @jbrewster7  - @prayagyadav was going to investigate a edm4hep coffea schema and stumbled on this PR. It looks like an open item for someone to pick up at this point, correct?

lgray:
@davidlange6 this PR is still usable but I know edm4hep as developed since we last touched it. I was working on finishing this up and then they changed everything. It remains a good starting point!

PR #825, opened 2023-06-01
feat: added rapidity getter in coffea/nanoevents/methods/vector.py

raeubaen:
Added a method to calculate rapidity for massive particles, for CMS higgs to gamma gamma columnar analysis (HiggsDNA) @emanueledimarco

### DIFF
--- a/coffea/nanoevents/methods/vector.py
+++ b/coffea/nanoevents/methods/vector.py
@@ -628,7 +628,7 @@ class LorentzVector(ThreeVector):
     def rapidity(self):
         pz = self.z
         e = self.energy
-        return 0.5*numpy.log((e+pz)/(e-pz))
+        return 0.5 * numpy.log((e + pz) / (e - pz))
 
     @property
     def boostvec(self):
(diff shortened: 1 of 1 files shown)


coffea/nanoevents/methods/vector.py
```diff
@@ -624,6 +624,12 @@ def pvec(self):
             behavior=self.behavior,
         )
 
+    @property
+    def rapidity(self):
+        pz = self.z
+        e = self.energy
+        return 0.5 * numpy.log((e + pz) / (e - pz))
```
lgray: `.5 * (numpy.log(e + pz)  - numpy.log(e - pz))` is more numerically stable

lgray:
@raeubaen this will go into coffea 2023 - if you need it in coffea 0.7 for now please implement your own rapidity calculation!

nsmith-:
Does this need to be backported to coffea 0.7?

lgray:
Don't think so it's simple enough to implement as a free function or monkey patch.

PR #835, opened 2023-06-10
fix: PFnano format for dask

yimuchen:
- [X] Fixing missing parenthesis for nested jet object access
- [X] Adding test file for automatic format tests
- [x] Fix issue where opening the same file multiple files can cause the access to nested structures to fail.


yimuchen:
This fixes the issue with first time opening, but has a strange side effect if you first open the file with dask and without dask: 

```python
from coffea.nanoevents.schemas import PFNanoAODSchema


def open_events(permit_dask=False):
        "file:./pfnano.root",
        schemaclass=PFNanoAODSchema,
        permit_dask=permit_dask,
    )
    return factory.events()
```

The following works fine
```python
events = open_events(permit_dask=False)
print(events.Jet.constituents.pf.pt)
events = open_events(permit_dask=False)
print(events.Jet.constituents.pf.pt)
```

The following is also fine: 
```python
events = open_events(permit_dask=False)
print(events.Jet.constituents.pf.pt)
events = open_events(permit_dask=False)
print(events.Jet.constituents.pf.pt)
```

But the following fails: 

```python
events = open_events(permit_dask=True)
print(events.Jet.constituents.pf.pt)
events = open_events(permit_dask=False)
print(events.Jet.constituents.pf.pt)
```

lgray:
figured this one out - thanks for isolating this problem, it was really starting to bug me!

PR #875, opened 2023-08-03
fix: Make rochester_lookup compatible with dask_awkward

### LINKED ISSUE #871: Rochester corrections not compatible with dask_awkward
**Describe the bug**
The Rochester corrections of given by `coffea.lookup_tools.rochester_lookup.rochester_lookup` are not compatible with dask_awkward. In particular, trying to execute `kSmearMC` on dask_awkward arrays will give a NotImplementedError. The reason seems to be that it uses `scipy.stats` functionality (through `doublecrystalball`).

**To Reproduce**
```python
import awkward as ak
import dask_awkward as dak
import coffea.lookup_tools


rochester_data = coffea.lookup_tools.txt_converters.convert_rochester_file(
    "tests/samples/RoccoR2018.txt.gz", loaduncs=True
)
rochester = coffea.lookup_tools.rochester_lookup.rochester_lookup(rochester_data)

x = ak.Array([[1.0]])
x = dak.from_awkward(x, 1)
mc_ksmear = rochester.kSmearMC(x, x, x, x, x, x)
```
Where `tests/samples/RoccoR2018.txt.gz` can be found in the coffea repository.

 - Coffea Version 2023.7.0rc0

jrueb:
Before, `rochester_lookup.kSmearMC` and `rochester_lookup.kSmearMCerror` would not work with dask-awkward arrays. This was because of a call to `doublecrystalball.ppf` (scipy) and `numpy.zeros_like`.
Fixes #871

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
   "Topic :: Utilities",
 ]
 dependencies = [
-  "awkward>=2.4.1",
+  "awkward>=2.4.2",
   "uproot>=5.0.10",
   "dask[array]>=2023.4.0",
   "dask-awkward>=2023.7.1,!=2023.8.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
   "Topic :: Utilities",
 ]
 dependencies = [
-  "awkward>=2.3.3",
+  "awkward>=2.4.1",
   "uproot>=5.0.10",
   "dask[array]>=2023.4.0",
   "dask-awkward>=2023.7.1,!=2023.8.0",
--- a/src/coffea/lookup_tools/rochester_lookup.py
+++ b/src/coffea/lookup_tools/rochester_lookup.py
@@ -230,11 +230,17 @@ class rochester_lookup:
         args = (u_flat, cbA_flat, cbA_flat, cbN_flat, cbN_flat, loc, cbS_flat)
 
         if any(isinstance(arg, dak.Array) for arg in args):
+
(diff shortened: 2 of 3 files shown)


lgray:
@ssrothman FYI we got the old version working - we should settle on a plan to correctionlib-ify rochester corrections.

The tool in coffea to extract the current rochester corrections into something pythonic should be a good starting point for onboarding since you got a good bit of the rest of it done iiuc.

@nsmith- @andrzejnovak @NJManganelli - adding you here since we should keep in mind that we'd like to go this direction (correctionlib rochester)

lgray:
Let me put the tests in dask and we'll be good to go I think.

jrueb:
I fixed the test a bit. The issue was that the first argument to `ak.unflatten` was a numpy array, while the second one was a dask-awkward array. This is not possible with awkward yet.
However, the test now seems to fail because of https://github.com/dask-contrib/dask-awkward/issues/313, which is still left unsolved.

lgray:
Thanks I'll take a look. I had to pause on this to take care of other issues.

lgray:
@jrueb just checking in here - the underlying issues in awkward / dask-awkward that are impeding progress here are being worked on my Angus and Jim. But it seems we wait on those for now.

lgray:
So with some fixes from awkward and dask-awkward this fails in a *different* way. I'll keep using this to exercise the changes coming down the pipeline that aim to fix the present failure.

lgray:
@jrueb FYI https://github.com/scikit-hep/awkward/pull/2652 now fixes the failures in this PR.

PR #879, opened 2023-08-17
fix: Make LumiMask compatible with dask-awkward

jrueb:
This PR makes it possible to provide dask-awkward arrays to lumi_tools.LumiMask.


lgray:
@jrueb Can you change the test so that it's using a dask-distributed Client instead of just local threaded execution? This will ensure it works in condor, etc.

PR #882, opened 2023-08-22
feat!: switch over to dask-based processing idioms, improve dataset handling

lgray:
This will:

- [x] add a tool for dask-based pre-processing of "filesets" (which are the uproot file-dictionary spec + metadata support)
   - Note that this produces json-serializable output so we can now easily store the results and do preprocessing a minimal number of times instead of every run.
- [x] add support for dask-driven execution of processors
- [x] provide tools for users that no longer wish to use processors which make their lives easier


lgray:
@valsdav Here's the PR with the fileset pre-processor. The query tool you're working on with rucio / servicex backend should produce json that's compatible with this spec (and we can go so far as to define a schema if you like).

Essentially - nests the spec mentioned [here](https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_dask.py#L123-L129) in uproot. So that you've got an object that looks like 
```
dataset = {
    "dataset_name": {"files": <the-uproot-spec>, "metadata": {...}, ...},
    "dataset2_name": ...,
    ...
}
```

The preprocessing step, as it is, strips out `files` from the file set, and runs a preprocessing step on each file to yield the desired chunking. It then returns the original input fileset with the files it was able to parse and determine the chunking of. It then returns the "available" dataset that it was able to parse and the "complete" dataset is all files which were input and only those reachable are updated, leaving unreachable files as given.

I'm going to add some modifiers on top of this to recover capabilities like `max_chunks`, etc.

valsdav:
Hi @lgray thanks! Starting to work on the query side :)

ikrommyd:
Pre-processing fails on a distributed Client with
```
TypeError: ('Could not serialize object of type HighLevelGraph', '<ToPickle: HighLevelGraph with 1 layers.\n<dask.highlevelgraph.HighLevelGraph object at 0x7f525471cfd0>\n 0. -get-steps-6f666e09968af4341f760475fe63382d\n>')
```
which fails during `all_processed_files = dask.compute(files_to_preprocess)[0]` with
```
AssertionError: bug in Awkward Array: attempt to get length of a TypeTracerArray

See if this has been reported at https://github.com/scikit-hep/awkward/issues

The above exception was the direct cause of the following exception:
```
throughout the error message. Works fine with local threads or processes

lgray:
@iasonkrom this is a limitation in the latest awkward array, I think. @agoose77 would know better.

Though I'm a bit confused why it's trying to serialize a typetracer in the first place, that shouldn't be happening.

ikrommyd:
It's not a big issue yet. You can just fire up the Client after you're done with preprocessing and are running your actual analysis code. It's just people tend to fire up the Client at the beggining of their script and that raises this error.

lgray:
The point is more that I'm not quite sure why this error is happening, I'm not creating a raw typetracer anywhere.

lgray:
@agoose77 here's a repro - you'll need to install coffea from this branch.
```python3
from coffea.dataset_tools import preprocess

if __name__ == "__main__":
    dataset = {
        "ZJets": {
            "files": {
                "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root": "Events",
            }
        }
    }

    with Client():
        dataset_runnable, dataset = preprocess(
            dataset, maybe_step_size=10, align_clusters=False, files_per_batch=10
        )

    with gzip.open("normalized_data.json.gz", "wb") as f:
        f.write(bytes(json.dumps(dataset), encoding="ascii"))

```

ikrommyd:
@lgray I was just running on some MC samples with preprocessing and there were just some sites that were not willing to respond and the preprocessor was running into the xrootd operation error. I also wasn't able to `xrdcp` such files. I can see you handle `FileNotFoundError` with a try-except. Perhaps a good idea would be a looper that loops over the sites rucio finds for each file with a try-except for xrootd `OSError` until it finds a site it can actually open the file from. I had to manually find such a site and it was the University of Florida site eventually. After that, the preprocessing and my code ran perfectly.

lgray:
Note to self for coming additions: we may also save the union over all the forms available in preprocessed files to further reduce the need for opening remote finals locally on subsequent runs.

lgray:
@valsdav is there any way we can test the rucio stuff in CI?

valsdav:
Ehy @lgray we would need a certificate for a service account, but that needs to be renovated automatically. Clemens is working on that for the GitLab runners. I will try to understand what's possible..

lgray:
@iasonkrom what you're mentioning is a feature that should go into fsspec-xrootd (this is basically the workaround that's in CMSSW) @nsmith- @lobis

.github/workflows/ci.yml
```diff
@@ -126,33 +126,33 @@ jobs:
...
-    - name: Test work_queue
-      shell: bash -l {0}
-      run: |
-        conda create --yes --name coffea-env -c conda-forge python=${{ matrix.python-version }} ndcctools dill conda-pack conda
-        conda activate coffea-env
-        python -m pip install --ignore-installed .
-        cd tests
-        conda-pack --output coffea-env.tar.gz
-        python wq.py coffea-env.tar.gz
+#  testwq:
```
nsmith-: Not much reason to test workqueue if taskvine tests that it can eat dask graphs.
Perhapse we might want to keep some sort of smoke test for various dask backends just to ensure they play well?

pyproject.toml
```diff
@@ -85,6 +85,11 @@ servicex = [
   "servicex>=2.5.3",
   "func-adl_servicex",
 ]
+rucio = [
+  "rucio-clients>=32;python_version>'3.8'",
+  "rucio-clients<32;python_version<'3.9'",
+  "cmd2",
```
nsmith-: Just to check, is some of the interactive console stuff available via `rich` (already a dependency) or do we need `cmd2`?

src/coffea/dataset_tools/__init__.py
```diff
@@ -0,0 +1,11 @@
+from coffea.dataset_tools.apply_processor import apply_to_fileset, apply_to_one_dataset
+from coffea.dataset_tools.manipulations import max_chunks, slice_chunks
+from coffea.dataset_tools.preprocess import preprocess
+
+__all__ = [
+    "preprocess",
+    "apply_to_one_dataset",
```
nsmith-: `apply_to_dataset`? One seems implied

src/coffea/dataset_tools/apply_processor.py
```diff
@@ -0,0 +1,38 @@
+import copy
+from typing import Callable, Union
+
+from coffea.nanoevents import NanoAODSchema, NanoEventsFactory
+from coffea.processor import ProcessorABC
+
+
+def apply_to_one_dataset(
+    data_manipulation: Union[ProcessorABC, Callable],
```
nsmith-: `function`, maybe in docs it can discuss the function needs to be traceable via dask_awkward

src/coffea/dataset_tools/apply_processor.py
```diff
@@ -0,0 +1,38 @@
...
+        raise ValueError("data_manipulation must either be a ProcessorABC or Callable")
+
+
+def apply_to_fileset(
+    data_manipulation: Union[ProcessorABC, Callable], fileset, schemaclass=NanoAODSchema
+):
+    out = {}
+    for name, dataset in fileset.items():
+        metadata = copy.deepcopy(dataset.get("metadata", {}))
+        metadata["dataset"] = name
```
nsmith-: Maybe better to `metadata.setdefault` in case users have preferred names? Or alternatively, warn/error if that key is already supplied and we are going to clobber it

src/coffea/dataset_tools/dataset_query.py
```diff
@@ -0,0 +1,366 @@
...
+
+        # Now let's print the results
+        tree = Tree(label=f"[bold orange]Replicas for [green]{dataset}")
+        for site, files in files_by_site.items():
+            T = tree.add(f"[green]{site}")
+            for f in files:
+                T.add(f"[cyan]{f}")
+        self.console.print(tree)
+
+    def do_whitelist_sites(self, args):
```
nsmith-: Blocklist/allowlist, deny/allow, exempt/allowlist or block/permit all are good synonyms
(https://www.ietf.org/archive/id/draft-knodel-terminology-09.html)

nsmith-: Also, docs should be clear about the precedence block-then-allow or allow-then-block

src/coffea/dataset_tools/preprocess.py
```diff
@@ -0,0 +1,169 @@
...
+
+def preprocess(
+    fileset,
+    maybe_step_size=None,
+    align_clusters=False,
+    recalculate_seen_steps=False,
+    files_per_batch=1,
+    skip_bad_files=False,
+    file_exceptions=(FileNotFoundError, OSError),
+):
```
nsmith-: Could the return type be described by a dataclass or protocol? (do we want to return a proper class here?)
We could normalize all instances to a proper dataclass on input to the various functions.

src/coffea/dataset_tools/rucio_utils.py
```diff
@@ -0,0 +1,294 @@
...
+import json
+import os
+import re
+import subprocess
+from collections import defaultdict
+
+from rucio.client import Client
+
+# Rucio needs the default configuration --> taken from CMS cvmfs defaults
+os.environ["RUCIO_HOME"] = "/cvmfs/cms.cern.ch/rucio/current"
```
nsmith-: Check if it is already set, only override if not. (for our non-CMS users!)

src/coffea/dataset_tools/rucio_utils.py
```diff
@@ -0,0 +1,294 @@
...
+    -------
+        nativeClient: rucio.Client
+            Rucio client
+    """
+    try:
+        if not proxy:
+            proxy = get_proxy_path()
+
+        nativeClient = Client(
+            rucio_host="https://cms-rucio.cern.ch",
```
nsmith-: This should automatically be taken from `rucio.cfg` in `$RUCIO_HOME`

src/coffea/processor/__init__.py
```diff
@@ -2,13 +2,7 @@
 
 
 """
-# deprecated run_uproot_job & executor usage:
-from functools import partial
-
-from coffea.nanoevents.schemas import NanoAODSchema, TreeMakerSchema
-
 from .accumulator import (
-    Accumulatable,
```
nsmith-, line 11: Why remove the protocol? I'd prefer that going forward over the ABC

src/coffea/dataset_tools/rucio_utils.py
lgray: @valsdav

src/coffea/dataset_tools/rucio_utils.py
lgray: @valsdav

src/coffea/dataset_tools/dataset_query.py
lgray: @valsdav

src/coffea/dataset_tools/apply_processor.py
lgray: yeah I'll pull the older list of protected keywords in the metadata.

src/coffea/dataset_tools/apply_processor.py
lgray:
```python3
def f():
    pass
print(isinstance(f, Callable))
```
prints `True` and other folks may want to implement functors with state and such, so Callable seemed most appropriate.

We could even restrict to 
```python3
Callable[
    [dask_awkward.Array], 
    Tuple[Union[
        dask.base.DaskMethodsMixin,
        Dict[str, dask.base.DaskMethodsMixin],
        Set[dask.base.DaskMethodsMixin],
        List[dask.base.DaskMethodsMixin],
        Tuple[dask.base.DaskMethodsMixin],
    ], ...]
]
```

src/coffea/dataset_tools/preprocess.py
lgray: The return type here is actually the same as the input type but in the case of the "runnable" output all Optional types are removed.

.github/workflows/ci.yml
lgray: It's been clear in the past that it's nice to have coffea do its own full-stack tests for things like this.

I don't think we've achieved the level of parity where tests on dask distributed completely confirm they will work on taskvine or the other way around. 

@btovar do you think that's achievable?

src/coffea/dataset_tools/apply_processor.py
lgray: From discussion in person I'll leave a comment in the code concerning the nested versions of the return arguments, it should be sufficient.

pyproject.toml
lgray: @valsdav

.github/workflows/ci.yml
btovar: Yes, definitely. I can make a pr. Further, since tasvine is orthogonal to coffea+dask, should this more in something like an examples directory?

src/coffea/dataset_tools/rucio_utils.py
valsdav: I'm trying to read it from `/cvmfs/cms.cern.ch/rucio/current/etc/rucio.cfg` but  that's not  a valid toml (missing " around strings).  Should I hack around it or can we ask rucio admins to fix the original file maybe?

```
# Copyright European Organization for Nuclear Research (CERN)
#
# Licensed under the Apache License, Version 2.0 (the "License");
# You may not use this file except in compliance with the License.
# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
#
# Authors:
# - Vincent Garonne, <vgaronne@gmail.com>, 2017
# - Eric Vaandering, <ewv@fnal.gov>, 2018

[common]
[client]
rucio_host = http://cms-rucio.cern.ch
auth_host = https://cms-rucio-auth.cern.ch
auth_type = x509_proxy
ca_cert = /cvmfs/cms.cern.ch/grid/etc/grid-security/certificates/ 
request_retries = 3
```

.github/workflows/ci.yml
lgray: @btovar a test that also serves as a worked-out example would be great! I don't think it's *so* orthogonal since, for instance, dask and distributed have deeper ties than dask and taskvine (so far as I know right now), but I still test that distributed works in coffea tests for the essential ingredients.

src/coffea/dataset_tools/apply_processor.py
lgray: actually no the setdefault makes more sense. though will that work as a parameter in awkward?

lgray: ah - yep

src/coffea/dataset_tools/rucio_utils.py
nsmith-: Sorry I wasn't clear here, the point is that the rucio config should provide all necessary arguments and it is sufficient to call `nativeClient = Client()`. e.g.
```
[ncsmith@lxplus909 ~]$ RUCIO_HOME=/cvmfs/cms.cern.ch/rucio/current python3
Python 3.9.18 (main, Sep  7 2023, 00:00:00)
[GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from rucio.client import Client
>>> client = Client()
>>> client.host
'http://cms-rucio.cern.ch'
>>> client.whoami()
{'email': 'nick.smith@cern.ch', 'deleted_at': None, 'updated_at': '2020-04-28T23:49:13', 'suspended_at': None, 'account_type': 'USER', 'account': 'ncsmith', 'status': 'ACTIVE', 'created_at': '2020-04-28T23:49:13'}
>>>
```

src/coffea/dataset_tools/preprocess.py
lgray: @nsmith- I've added typing here now to define the interface more concretely - please have a look.

lgray:
@valsdav now that I have the interface pinned down do you think you can make it so the rucio utils produce filesets that can be eaten by preprocess? Once we have that we're in a pretty good spot! :D

ikrommyd:
> @valsdav now that I have the interface pinned down do you think you can make it so the rucio utils produce filesets that can be eaten by preprocess? Once we have that we're in a pretty good spot! :D

@lgray may I suggest that this is optional (or provide an option to disable it and keep it as it is now) however as users may want to use their own key names for the fileset dictionary?

Also the ability to create one big fileset from a set of querries would be nice

lgray:
Sure, I was thinking a separate function in any case, then it's all nice and composable if you don't want to customize too much.

lgray:
I guess another way to put it is it would be really pleasant to start from a data card defined something like:
```python3
my_input_datasets = {
    "ZJets": "/DYJetsToLL*generatorsettings*/*SomeSearchCriteria*/NANOAODSIM",
    "DiMuon": "/DoubleMuon/*OtherSearchSettings*/NANOAOD",
    ...,
}
```
or
```python3
my_input_datasets = {
    "/DYJetsToLL*generatorsettings*/*SomeSearchCriteria*/NANOAODSIM": {"short_name": "ZJets", "metadata": {"xsec": xyz.w, ...}},
    "/DoubleMuon/*OtherSearchSettings*/NANOAOD": {"short_name": "DiMuon", "metadata": {...}},
    ...,
}
```
and then be able to work your way to concrete dataset list and sites, and then finally render that into lists of files that we can preprocess.

A progression like this would make switching Runs or otherwise evolving an analysis to more datasets while also keeping it pretty concise and explainable.

src/coffea/dataset_tools/rucio_utils.py
valsdav: I see :) thanks!

valsdav:
> @valsdav now that I have the interface pinned down do you think you can make it so the rucio utils produce filesets that can be eaten by preprocess? Once we have that we're in a pretty good spot! :D

Sure,  I can also add a command in the CLI to call direcly the preprocessing.  And another one to start from a data card like one of your examples so that they appears in the list of datasets to process for replicas. 

I will work on it in the next days!

lgray:
@valsdav cool - I think with that and addressing Nick's comments this will be good to go.

lgray:
@valsdav I think I prefer the second example of the high-level datacard, but if other folks have ideas I'm rather open (though handling either variant shouldn't be difficult FWIW).

pyproject.toml
valsdav: switched to full rich cli, removed cmd2 dependency

lgray:
@nsmith- can you give this another run through? Intend to merge later today unless there is something big.

PR #887, opened 2023-08-31
feat: also pass `uproot_options` to `uproot.dask` in `from_root()`

ikrommyd:
I also had to comply to the other changes in nanoevents in `local_executors_to_dask` which were changing the `chunks_per_file` to `None` and also adding this `elif chunks_per_file is None`.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -232,7 +232,7 @@ class NanoEventsFactory:
         treepath="/Events",
         entry_start=None,
         entry_stop=None,
-        chunks_per_file=None,
+        chunks_per_file=1,
         runtime_cache=None,
         persistent_cache=None,
         schemaclass=NanoAODSchema,
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -232,7 +232,7 @@ class NanoEventsFactory:
         treepath="/Events",
         entry_start=None,
         entry_stop=None,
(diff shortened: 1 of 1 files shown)


ikrommyd:
@lgray apparently changing the default `chunks_per_file` to `None` causes the local_executor tests to fail. I checked locally

lgray:
Don't worry about those they're going away. Would remove that fix from this patch. Chunks_per_file=None should be in the PR with preprocess.

ikrommyd:
alright, but I'm keeping the `elif chunks_per_file is None` since there is `uproot_options` passing there

nikoladze:
i also wanted to suggest passing `uproot_options` to `uproot.dask` in dask mode - i'm currently experimenting with custom uproot sources that i could then pass as `file_handler` or `http_handler` arguments here

lgray:
yeah! this is definitely going in - especially if we want fspec to work with root files correctly, many more options may need to be passed.

lgray:
@iasonkrom please make sure these changes work well for you. Now I'm using uproot's mechanism for setting the default of the file chunking. Should work reasonably in all cases now.

ikrommyd:
It should be fine like that. It's just that now `chunks_per_file=None` is no longer an option and it will run into `unsupported operand type(s) for *: 'int' and 'NoneType'` from uproot. So we are just complying to uproot's interface of requiring only `int` `steps_per_file` while `None` corresponds to just "leave the default option"

PR #888, opened 2023-08-31
fix: allow for collections that contain non-jagged arrays in PHYSLITE schema

nikoladze:
So far the PHYSLITE schema didn't handle collections that contain both scalars and lists (like `EventInfo` and `xTrigDecision`). This should be now fixed. In addition i also added treatment for branches that are not grouped with anything (newer PHYSLITE files have a branch named `index_ref` that falls in this category).
This PR should be possible to merge independently of #872 (i tested it also on top of these changes).


lgray:
merging this since it does not introduce new failures (waiting on awkward fix)

PR #891, opened 2023-09-07
build: pin dask-awkward >= 2023.9.3, awkward >= 2.4.3

lgray:
this turns on form rehydration and should unstick problems with sum/ones_like/etc.


lgray:
unexpected problems with awkward. will be a few days since @agoose77 is in transit.

lgray:
superseded by #900

PR #895, opened 2023-09-15
fix: comply with `boost-histogram` 1.4.0

ikrommyd:
`boost-histogram` 1.4.0 now states that "filling an integer-based axis requires an integer array - a floating point array is disallowed (a single float was always disallowed), due to issues with rounding around 0". So, I add `dtype=int` to all `ak.zeros_like` and `np.arange` in analysis_tools.

### DIFF
--- a/src/coffea/analysis_tools.py
+++ b/src/coffea/analysis_tools.py
@@ -582,16 +582,10 @@ class NminusOne:
         else:
             return out
 
-    def print(self, compute=False):
+    def print(self):
         """Prints the statistics of the N-1 selection"""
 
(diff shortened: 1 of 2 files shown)


lgray:
There's a broken test from awkward right now. I'll check that none of the associated tests for analysis tools are broken by this.

ikrommyd:
I've only touched `analysis_tools.py`. I control-F'ed in vscode but couldn't find another such instance in the source code. Tests pass locally so unless we encounter another apple silicon vs x86 case it should be good.

ikrommyd:
Touched a few more minor stuff since we're here.

PR #900, opened 2023-10-03
feat: support `shape_touched` from Dask

agoose77:
This PR *at present* represents a hack of the existing uproot remapping loader to support dask-awkward's new buffer projection interface.

@lgray the Coffea `form_mapping` object that is passed to `uproot.dask` needs to satisfy the `ImplementsFormMapping` interface:
```python
class ImplementsFormMapping(Protocol):
    def __call__(self, form: Form) -> tuple[Form, ImplementsFormMappingInfo]:
        ...
```

The idea is that the form mapping generates _both_ the remapped form _and_ any auxiliary state required for projection. The return tuple's second item, `ImplementsFormMappingInfo` has the following protocol:
```python
class ImplementsFormMappingInfo(Protocol):
    @property
    def behavior(self) -> dict | None:
        ...

    buffer_key: str | Callable

    def parse_buffer_key(self, buffer_key: str) -> tuple[str, str]:
        ...

    def keys_for_buffer_keys(self, buffer_keys: frozenset[str]) -> frozenset[str]:
        ...

    def load_buffers(
        self,
        tree: HasBranches,
        keys: frozenset[str],
        start: int,
        stop: int,
        options: Any,
    ) -> Mapping[str, AwkArray]:
        ...
```

This PR hacks the `_map_schema_uproot` object to provide this interface, but it could be cleaner; one could remove any unused code here.


src/coffea/processor/executor.py
```diff
@@ -1718,7 +1718,7 @@ def _work_function(
                     import dask_awkward
 
                     to_compute = processor_instance.process(events)
-                    materialized = dask_awkward.necessary_columns(to_compute)
+                    # materialized = dask_awkward.report_necessary_buffers(to_compute)
```
agoose77, line 1721: @lgray I wasn't clear on the scope of `materialized`, as it's populated in two different places. Could you figure out what you'd like to do here? `dak.report_necessary_buffers` produces a mapping from layer name to a `NecessaryBuffers` or `None`. The `NecessaryBuffers` namedtuple contains two fields, one for data+shape touched buffers, the other for shape-only.

agoose77:
@lgray this PR implements the protocols that uproot expects for form remapping in `factory.py` for the Uproot source. I'm not yet au-fait with how the parquet equivalent is handled here, but you'd probably need to replicate the logic in `uproot._dask.UprootReadMixin` for the parquet IO function.

Can I hand this over to you?

lgray:
Yes I can handle the parquet reader stuff - no big deal there.

I'll give this a try today.

lgray:
@agoose77 we're just missing a release of dask-histogram and then I can mess with pins here. I'll let you handle the pre-release vs. full release state of things.

Once we get everything to a full release we can merge this one.

src/coffea/processor/executor.py
lgray: Oh this code is about to be removed in any case that was kept around to keep tests passing and exercise processing workflows.

It's doing the job of some old code in awkward1, the pattern for which is now obsolete.

pyproject.toml
```diff
@@ -37,11 +37,11 @@ classifiers = [
...
 ]
 dependencies = [
-  "awkward>=2.4.2",
-  "uproot>=5.0.10",
+  "awkward>=2.4.5",
+  "uproot>=5.1.0rc1",
   "dask[array]>=2023.4.0",
-  "dask-awkward>=2023.7.1,!=2023.8.0",
-  "dask-histogram>=2023.6.0",
+  "dask-awkward>=2023.10a1,!=2023.8.0",
```
agoose77:
```suggestion
  "dask-awkward>=2023.10a1",
```

lgray:
For the two new protocols defined we should update dask-awkward since that's where the present protocol `ImplementsFormTransformation` is kept and I have no problem with changing that.

agoose77:
> For the two new protocols defined we should update dask-awkward since that's where the present protocol `ImplementsFormTransformation` is kept and I have no problem with changing that.

Could you clarify what you mean here @lgray ? :)

lgray:
@agoose77 

This:
https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/io/io.py#L45-L61

agoose77:
I've made a PR to remove it, as it's a better fit for uproot! :)

lgray:
@agoose77 but doesn't that interface apply equally to parquet / json / etc.? The operation of form remapping seems pretty generalizable, no?

agoose77:
We don't use the interface in dask-awkward; only coffea implements remapping, and only uproot uses the interface. So, I feel that it's a non-dask-awkward feature; that uproot or coffea should define these interfaces. Realistically, uproot?

lgray:
This certainly isn't blocking and I agree coffea is the only thing that does remapping but it will be doing remapping on any possible data source, and remapping is not in general specific to coffea. This is why I think it makes more sense in dask-awkward or eventually another package since the concept is rather general (even if not everyone needs to use it).

I'm happy to keep the definitions of those protocols here for now but I don't think it is their long term home.

src/coffea/nanoevents/factory.py
```diff
@@ -125,7 +141,41 @@ def __call__(self, form):
...
+        class TranslateBufferKeys:
+            def __getitem__(this, key):
+                form_key, attribute = self.parse_buffer_key(key)
+                return mapping[
+                    buffer_key(form_key=form_key, attribute=attribute, form=None)
+                ]
+
+        return TranslateBufferKeys()
 
     def create_column_mapping_and_key(self, tree, start, stop, interp_options):
```
agoose77: @lgray can we remove this from the base definition, or does someone out there likely use it?

tests/test_jetmet_tools.py
```diff
@@ -837,9 +837,9 @@ def test_corrected_jets_factory(optimization_enabled):
             **{name: evaluator[name] for name in jec_stack_names[5:6]}
         )
 
-        print(dak.necessary_columns(jets.eta))
+        print(dak.report_necessary_buffers(jets.eta))
```
agoose77:
```suggestion
        print(dak.report_necessary_columns(jets.eta))
```

tests/test_jetmet_tools.py
```diff
@@ -837,9 +837,9 @@ def test_corrected_jets_factory(optimization_enabled):
...
         print(
-            dak.necessary_columns(
+            dak.report_necessary_buffers(
```
agoose77:
```suggestion
            dak.report_necessary_columns(
```

src/coffea/nanoevents/factory.py
lgray: interface is completely private and rather obscure to boot, so I am fine to remove it.

tests/test_jetmet_tools.py
lgray: is `dak.necessary_columns` deprecated? if so on what time scale?

agoose77: `necessary_columns` → `report_necessary_columns`.

We can make it as long as necessary. At one point, I was going to remove it in favour of `report_necessary_buffers`, but the new `report_necessary_columns` is designed to allow IO layers to report columns iff. they support them.

tests/test_jetmet_tools.py
```diff
@@ -837,9 +837,9 @@ def test_corrected_jets_factory(optimization_enabled):
             **{name: evaluator[name] for name in jec_stack_names[5:6]}
         )
 
-        print(dak.necessary_columns(jets.eta))
+        print(dak.report_necessary_columns(jets.eta))
```
agoose77, line 840: @lgray note that `report_necessary_columns` reports whatever the IO layer thinks a "column" is. It doesn't necessarily mean "dotted field" right now, rather I envisage it talking about the concept of a "key" or "column" that can be read from the IO source. 

@douglasdavis this is something I didn't fully think about when we made `necessary_columns` shim `report_necessary_columns`.

The difference will be that for remapped forms, the `report_necessary_columns` (and by proxy, `necessary_columns`) will report the underlying ROOT file keys, not the remapped keys.

agoose77: It therefore might mean that we should remove that alias for `necessary_columns`.

lgray: I actually prefer the new behavior since users will probably expect to get a list of columns in the file they're reading! But seeing the remapped keys is nice for debugging.

lgray: Can we make it so that we get a version or two of dask-awkward before `dak.necessary_columns` is gone and give migration instructions? I know a few end-users that use that function.

lgray:
@nsmith- I think this one is ready for review now. If you could give it a bit of time before the end of the week it would be appreciated.

src/coffea/nanoevents/factory.py
```diff
@@ -77,24 +76,50 @@ def __init__(
...
-        for form_key in form_keys:
-            base_columns.extend(
+    def keys_for_buffer_keys(self, buffer_keys):
+        base_columns = set()
+        for buffer_key in buffer_keys:
+            form_key, attribute = self.parse_buffer_key(buffer_key)
+            operands = urllib.parse.unquote(form_key).split(",")
+
+            it_operands = iter(operands)
+            next(it_operands)
```
agoose77: This function behaves like `extract_form_key_base_columns`, but supports operands e.g. attribute lookup.

src/coffea/nanoevents/factory.py
```diff
@@ -147,8 +175,17 @@ def create_column_mapping_and_key(self, tree, start, stop, interp_options):
             use_ak_forth=True,
         )
         mapping.preload_column_source(partition_key[0], partition_key[1], tree)
```
agoose77, line 178: I've deliberately tried not to touch this logic at all. Hopefully, by presenting the correct interface to uproot, internal refactoring can happen here in future without any side effects.

tests/test_jetmet_tools.py
agoose77: @lgray I'm not 100% clear if we're on the same page. 

`report_necessary_columns` is a new function, so I'm comfortable with its semantics deviating from `ak.necessary_columns`. 

It would be possible to restore the old behavior, but it will require a reasonable amount of work on the uproot side (I am guessing).

`ak.necessary_columns` is currently an alias for this new function, but for uproot this wouldn't be a non-breaking change; if previously it would report `ROOT_KEY.record.foo.bar`, now it would report `ROOT_KEY`. Is this OK? I don't see it being that useful if it breaks existing users.

lgray:
suspicious new failure...

tests/test_jetmet_tools.py
lgray: OK I suppose for my most typical use case I never see keys with `ROOT_KEY.record.foo.bar` formatting, it's really only ever `ROOT_KEY` and any structure is up in the form being remapped to. So they appear the same to me.

ikrommyd:
I'd like to report that I've been using this all day on LPC running stuff with the newest awkward, uproot, dask-awkward and dask-histogram and have not ran into any errors.

agoose77:
Todo: bump the uproot and dask-awkward versions

lgray:
@nsmith- a gentle reminder for a review, please!

src/coffea/processor/executor.py
```diff
@@ -1734,11 +1734,11 @@ def _work_function(
                     metrics = {}
                     if isinstance(file, uproot.ReadOnlyDirectory):
                         metrics["bytesread"] = file.file.source.num_requested_bytes
+                    # metrics["data_and_shape_buffers"] = set(materialized)
+                    # metrics["shape_only_buffers"] = set(materialized)
```
nsmith-, line 1738: Are these placeholders for something in the future?

src/coffea/nanoevents/factory.py
```diff
@@ -77,24 +76,50 @@ def __init__(
...
+                    if maybe_transform == "!load"
                 ]
             )
-        return list(set(base_columns))
+        return base_columns
+
+    def parse_buffer_key(self, buffer_key):
+        prefix, attribute, form_key = buffer_key.rsplit("/", maxsplit=2)
+        if attribute == "offsets":
+            return (form_key[: -len("%2C%21offsets")], attribute)
```
nsmith-: This seems to also be in the previous code as well, but some of these string splitting and urllib name mangling things were put into https://github.com/CoffeaTeam/coffea/blob/master/src/coffea/nanoevents/util.py and could be used here

src/coffea/nanoevents/factory.py
```diff
@@ -68,7 +67,7 @@ def _key_formatter(prefix, form_key, form, attribute):
     return prefix + f"/{attribute}/{form_key}"
 
 
-class _map_schema_base(ImplementsFormTransformation):
+class _map_schema_base:  # ImplementsFormMapping, ImplementsFormMappingInfo
```
nsmith-, line 71: Any reason not to inherit the protocol(s)?

nsmith- (approved):
We need to do some cleanup, but it can happen later I think.
@agoose77 does the protocol you defined in the initial comment have some explanatory docstrings available somewhere?

src/coffea/processor/executor.py
lgray: This is just to get tests to run - executor is going to be significantly changed in a subsequent PR.

src/coffea/nanoevents/factory.py
agoose77: They weren't exported yet! I think a new release includes them. Let me check.

PR #916, opened 2023-10-27
build: try ak 2.5.0rc0

lgray:
@agoose77

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
   "Topic :: Utilities",
 ]
 dependencies = [
-  "awkward>=2.4.6",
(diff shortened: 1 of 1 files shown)


agoose77:
And here I was thinking that my afternooon was going to be simple ...

I know what the cause is. It's really strange - NumPy didn't report this in the changelogs, so it looks like an old feature. In fact, I think it's very very new. We'll need to add a backwards compatible shim until then.

lgray:
also - thank you for making this an rc :-)

agoose77:
I think the breakage might also be in 2.4.7 — the RC was intended to catch a typetracer change. This wasn't caught in our CI — I'll need to check our lower bounds on NumPy.

PR #921, opened 2023-11-05
build: pin up to awkward 2.4.8, dask-awkward 2023.11.0

lgray:


lgray:
@agoose77 new error popped up in workqueue tests ```ValueError: defined_divisions only works when divisions are known.```

agoose77:
@douglasdavis would you be able to take a peek at this?

lgray:
same error in the normal CI! so should be easy to pin down.

lgray:
sigh guess I can't just do that. annoying.

agoose77:
@lgray you can permit direct references with 
```yaml
[tool.hatch.metadata]
allow-direct-references = true
```

lgray:
cool so we just need a new dask-awkward release and a repin @agoose77

douglasdavis:
> cool so we just need a new dask-awkward release and a repin @agoose77

`2023.11.0` is on PyPI!

PR #922, opened 2023-11-07
feat: switch permit_dask to True by default

lgray:
This PR makes dask-based nanoevents the default.

Currently there are some unrelated blockers for from (dask-)awkward related to numpy reducers.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -245,7 +245,7 @@ class NanoEventsFactory:
         access_log=None,
         iteritems_options={},
         use_ak_forth=True,
-        permit_dask=True,
+        delayed=True,
     ):
         """Quickly build NanoEvents from a root file
 
@@ -277,8 +277,8 @@ class NanoEventsFactory:
                 Pass a list instance to record which branches were lazily accessed by this instance
             use_ak_forth:
                 Toggle using awkward_forth to interpret branches in root file.
-            permit_dask:
-                Allow nanoevents to use dask as a backend.
+            delayed:
+                Nanoevents will use dask as a backend to construct a delayed task graph representing your analysis.
         """
 
         if treepath is not uproot._util.unset and not isinstance(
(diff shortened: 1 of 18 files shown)


lgray:
@nsmith- I've had to mangle some parts of the nanoevents schema system to get this all to fit right. I think it's worth a look from you to determine a way to make the non-user facing parts of the interface decent again.

lgray:
@nsmith- can you give this a look and lodge your complaints? May not get fixed in this PR but we can plan it for the future.

src/coffea/nanoevents/factory.py
```diff
@@ -252,7 +245,7 @@ def from_root(
         access_log=None,
         iteritems_options={},
         use_ak_forth=True,
-        permit_dask=False,
+        permit_dask=True,
```
nsmith-: This is a big API change, but it is coming in a yet-unreleased major coffea version change, so it is ok. It also helps users to be exposed to dask. My only complaint is that the argument name is still a bit strange to me, rather I would think `delayed=True` or `eager_mode=False` would be better?

tests/test_analysis_tools.py
```diff
@@ -10,14 +10,14 @@
 fname = "tests/samples/nano_dy.root"
 eagerevents = NanoEventsFactory.from_root(
     {os.path.abspath(fname): "Events"},
-    schemaclass=NanoAODSchema.v6,
+    schemaclass=NanoAODSchema,
```
nsmith-, line 13: Does the `.v6` specialization still work but is just not needed?

lgray: Right now the `.vX` specializations are a little broken in their original implementation. The functional sub-classing thing you were doing doesn't quite fit into the daskiness.

src/coffea/nanoevents/factory.py
lgray: I think `delayed=True` makes more sense. I'll change this.

PR #923, opened 2023-11-08
build: try dask-awkward 2023.11.1

agoose77:


lgray:
@agoose77 So the issue was indeed back only in dask-awkward. I bamboozled myself with a descriptive variable that changed to being dask instead of eager, so the issue lies entirely in dask-awkward.

douglasdavis:
I've tested coffea's `tests/test_analysis_tools.py` with dask-awkward `2023.11.2` and things pass (and awkward `2.4.10`)

lgray:
Thanks! Running the CI here.

PR #926, opened 2023-11-13
feat: use new dask methods

agoose77:
Leverages dask-contrib/dask-awkward#409


lgray:
@agoose77 this is in the right direction but there's one big problem in that now iterated `__getattr__` seems to not be supported:

repro:
```python3
    "tests/samples/nano_dy.root:Events",
).events()

events.Jet.matched_muons.matched_jet
```

now fails with:
```
AttributeError: 'function' object has no attribute 'matched_jet'
```

lgray:
@agoose77 

Still running into some regressions:
```
pytest tests/test_nanoevents_pfnano.py tests/test_nanoevents_physlite.py
```

results in a couple failed tests with:
```
TypeError: meta must be an instance of an Awkward Array, not <class 'dask_awkward.lib.core.Array'>.
```
and
```
NotImplementedError: 
```

Which were not encountered before!

lgray:
@agoose77 ok this last edit gets rid of all the regressions on my side.

agoose77:
OK, let's look to merging this in dask-awkward, then :)

lgray:
@agoose77 is there any way we can make it so I don't have to repeat functions like metadata that access awkward-array information but don't need dask? That seems rather pesky.

agoose77:
@lgray sure! Let me think about it.

agoose77:
@lgray try this (with new dask-awkward in my branch)

lgray:
all green in local tests.

lgray:
Once all that gets merged I'll finally switch nanoevents to use transient attrs and can get rid of that nasty copyreg I use right now. Woo.

PR #929, opened 2023-11-16
feat!: drop use of `opt_touch_all=False`

### LINKED ISSUE #915: `opt_touch_all` argument of `dak.map_partitions` should no longer be necessary
The `opt_touch_all` argument to `map_partitions` was added to make layers in dask-awkward graphs manually touchable (that is, tell the optimization code to not rely on the typetracer graph for touching in that layer, just touch all awkward data at that point). This allowed us to workaround operations where automatic typetracer touching wasn't working. I think we're in a place now where we don't need it. I see it's used in coffea here:

https://github.com/CoffeaTeam/coffea/blob/c541a8ae095aa8db71f734bfbe085d4dba6dadf1/src/coffea/ml_tools/helper.py#L351

---

More info just for completeness:

The only place where we needed this was in `dask_awkward.zip`, where if we did something like:

```python
upstream = get_data()
new_data = dak.zip({"foo": upstream.foo, "bar": upstream.bar})
new_data.foo.compute()
```
That compute would fail if we didn't also touch `upstream`'s `bar` data, but `bar` wouldn't get touched by the compute on only `foo`, hence the need to manually touch the whole zipped up array.

I've made draft PR https://github.com/dask-contrib/dask-awkward/pull/395 for testing

agoose77:
In line with https://github.com/dask-contrib/dask-awkward/pull/395

Fixes #915


lgray:
Both this and https://github.com/dask-contrib/dask-awkward/pull/395 look good to go!

PR #933, opened 2023-11-25
test: add skip for missing dependency, fixture for local tests directory

lobis:
While working on https://github.com/CoffeaTeam/coffea/pull/930 I had to run the tests locally and I ran into some problems that other new users may face when running tests:

- I only installed the `dev` dependencies (did not install `dask` dependencies) at first. Tests were failing due to missing dependencies. I added a pytest import skip if `dask.distributed` is missing.
- I use the PyCharm IDE and when running a given test function from the IDE it fails as PyCharm tries to run the test function from the location of the code file, instead of running it from the root directory. I added a pytest fixture so that test sample file paths are independant of the directory where the test is run. I didn't cover all instances where a local samples path appears but covered many instances.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,8 @@
+import os
+
+import pytest
+
+
+@pytest.fixture(scope="module")
+def tests_directory() -> str:
+    return os.path.dirname(os.path.realpath(__file__))
--- a/tests/test_lookup_tools.py
+++ b/tests/test_lookup_tools.py
@@ -371,24 +371,26 @@ def test_jec_txt_effareas():
     print(evaluator["photon_id_EA_Pho"])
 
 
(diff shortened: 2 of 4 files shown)


lgray:
@lobis It may make more sense to include distributed now as part of `[dev]` or `[test]` given how much stuff uses it now. Also, a lot of the files you changes are getting deleted in #882. I don't think I'll merge this PR as is.

lgray:
Once #882 is merged this PR will need to be redone - but the changes are a good idea.

lgray:
@lobis dask distributed is now included in `[dev]`, the addition for pycharm seems nice! Please update this PR at your leisure.

lobis:
@lgray I guess this is ready to merge once the tests pass!

PR #935, opened 2023-11-25
feat: TreeMakerSchema Skimming (r2023)

yimuchen:
- Providing a standardized method for TreeMakerSchema events to be written by uproot
- Allow TreeMakerSchema events to read back uproot-saved events


lgray:
This will need a follow-up PR later to properly integrate it with dask-awkward.

yimuchen:
> This will need a follow-up PR later to properly integrate it with dask-awkward.

This is on my radar, I'll try and follow up this weekend to test running this with dask-awkward (Only made sure the eager version runs without issues for now)

PR #940, opened 2023-11-28
fix: improvements to dataset_query tools

valsdav:
- [x] Fixed allowlist/blocklist naming
- [x] More general scope (not only cms hardcoded paths)
- [x] Read rucio config from file
- [x] select and query replicates with an "all" option in the cli 
- [x] Read  datasets definition from file
- [x] Call preprocessing macro from the cli
- [x] Move away from cmd2 and just use rich.Prompt
- [ ] **docs**: 
   - [x] notebook


lgray:
@valsdav ping me when you're done and I'll merge to the main PR!

valsdav:
Moving now the implementation of the cli to pure `rich` to remove `cmd2` dependency and reuse the functions for "dataset config" loading.

PR #949, opened 2023-12-01
fix: remove copyreg workaround, use attrs to store reference to original array

lgray:
As in title, this fixes a long-standing nasty override over dask-awkward.Array copying behavior.

There are a few failed tests that I'm tracking down, otherwise it looks like it was straightforward.


lgray:
This PR is waiting on https://github.com/dask-contrib/dask-awkward/pull/429 in a release and a new release of awkward array.

@agoose77 @douglasdavis

PR #952, opened 2023-12-04
test: taskvine+coffea+dask with pytest

btovar:
Incudes two tests, one using the local conda environment, and another that sends the environment with the dask jobs.


lgray:
@btovar lemme know when you're done and I'll merge!

btovar:
I think we can merge it now.  I'll work in adapting the second test in another pr. Thanks!

PR #962, opened 2023-12-12
build(deps): pin up to awkward 2.5.1, fsspec-xrootd 0.2.3

lgray:


matthewfeickert:
Thank you for this, as I just hit an error with `fsspec-xrootd` `v0.2.2`. :) All good now with the bumped version! :+1:

PR #964, opened 2023-12-13
test: taskvine with remote env

btovar:
Enables and fixes the taskvine test that needs an environment tarball to run at a compute node.


btovar:
Thanks @lgray!

PR #965, opened 2023-12-13
fix: improved DataDiscoveryCLI interface

valsdav:
Improving the DataDiscoveryCLI

- `do_replicas` and `load_dataset_definitions` now returns uproot compatible metadata
- Added "first" option for rucio replicas
- Returning both overall and updated datasets from preprocess
- Fixes and docs

@iasonkrom


src/coffea/dataset_tools/dataset_query.py
```diff
@@ -553,7 +571,11 @@ def load_dataset_definition(
         and selected results and replicas following the options.
 
         - query_results_strategy:  "all" or "manual" to be prompt for selection
-        - replicas_strategy: "round-robin", "choose" (to manually choose the sites), "manual": to be prompt for manual decision case by case
+        - replicas_strategy:
+            - "round-robin": select randomly from the available sites for each file
+            - "choose": filter the sites with a list of indices for all the files
+            - "first": take the first risult returned by rucio
```
ikrommyd:
```suggestion
            - "first": take the first result returned by rucio
```

ikrommyd: Just saw this typo while looking at the changes, will give it a spin later. Thanks @valsdav!

src/coffea/dataset_tools/dataset_query.py
```diff
@@ -578,16 +600,17 @@ def load_dataset_definition(
...
         else:
-            if replicas_strategy not in ["round-robin", "choose"]:
+            if replicas_strategy not in ["round-robin", "choose", "first"]:
                 print(
-                    "Invalid replicas-strategy: please choose between manual|round-robin|choose"
+                    "Invalid replicas-strategy: please choose between manual|round-robin|choose|first"
                 )
                 exit(1)
-            self.do_replicas(mode=replicas_strategy, selection="all")
+            out_replicas = self.do_replicas(mode=replicas_strategy, selection="all")
```
ikrommyd, line 608: If I'm seeing correctly, `do_replicas()` doesn't return anything so `out_replicas` will be `None`

ikrommyd:
Should we have "first" or "round-robin" be the default though? Opinions?

valsdav:
`round-robin` by default distributes the load more between sites.

ikrommyd:
Does it "reduce" the quality of the sites though? Are "bigger" sites generally better?

lgray:
Round-robin is fine and probably scales better. If it turns out that it causes problems we can change it later, but I think that gets determined by experimentation rather than how we think it is supposed to work.

ikrommyd:
One final thing to ask is a method that returns the dictionary that `do_save` would write to a json file. Maybe `.as_dict` like @lgray mentioned in Slack. After that, it looks good to me!

lgray:
Either `as_dict` or `to_dict` would be fine, but yes, that's needed @valsdav .

ikrommyd:
Oh I got one more thought. What if we also keep `out_available` and `out_updated` dictionaries as class attributes if preprocess was run? Just to have direct access to them from the class and not only from the json files.

src/coffea/dataset_tools/dataset_query.py
```diff
@@ -542,19 +544,19 @@ def do_preprocess(
             "[red] Preprocessing files to extract available chunks with dask[/]"
         ):
             with Client(dask_cluster) as _:
-                out_available, out_updated = preprocess(
+                self.preprocess_available, self.preprocessed_total = preprocess(
```
ikrommyd: `self.preprocessed_available` not `preprocess_available`

src/coffea/dataset_tools/dataset_query.py
```diff
@@ -542,19 +544,19 @@ def do_preprocess(
...
                 )
         with gzip.open(f"{output_file}_available.json.gz", "wt") as file:
             print(f"Saved available fileset chunks to {output_file}_available.json.gz")
-            json.dump(out_available, file, indent=2)
+            json.dump(self.preprocessed_total, file, indent=2)
         with gzip.open(f"{output_file}_all.json.gz", "wt") as file:
             print(f"Saved all fileset chunks to {output_file}_all.json.gz")
-            json.dump(out_updated, file, indent=2)
-        return out_available, out_updated
+            json.dump(self.preprocess_available, file, indent=2)
```
ikrommyd: same here

src/coffea/dataset_tools/dataset_query.py
```diff
@@ -542,19 +544,19 @@ def do_preprocess(
...
+        return self.preprocessed_total, self.preprocess_available
```
ikrommyd: same here

lgray:
@valsdav @iasonkrom ping me when you've converged, and I'll merge it.

ikrommyd:
@lgray `do_preprocess()` takes in `dask_cluster=None` and spawns a `with Client(dask_cluster) as _` to do the preprocessing. But what if the user already has a `Client` running? No reason to spawn another one right? I'm good with all the rest

ikrommyd:
I don't know if passing in `dask_client` and not `dask_cluster` is the correct approach here

valsdav:
if you provide a dask scheduler url the Client is going to connect to that. if None it will spawn a local multiprocessing cluster, isn't it?

ikrommyd:
I'm not sure about that. Lindsey will have to confirm that it's fine. I'm good with all the rest. @lgray you're free to review

lgray:
That's all the functionality we should expect in a CLI system, see https://distributed.dask.org/en/stable/client.html.

So I think we can provide the scheduler endpoint as a string and call it a day here.

If you need to do more complicated preprocessing, you can get the raw list and do what you want, or call the pre-processing command in a script in the appropriate context manager with your dask cluster setup.

What's there is fit to task.

lgray:
@valsdav change `dask_cluster` to `scheduler_url` and I think the usage becomes more clear.

ikrommyd:
so if I've spawned an lpc condor client with:
```python
from lpcjobqueue import LPCCondorCluster


cluster = LPCCondorCluster()
cluster.adapt(minimum=0, maximum=100)
client = Client(cluster)
```
what should I do to make `do_preprocess` use that `Client`?

lgray:
`as_dict` -> call `coffea.dataset_tools.preprocess`

ikrommyd:
> `as_dict` -> call `coffea.dataset_tools.preprocess`

ah okay, so we just don't use the CLI class's method in that case. Thanks

I'm good with merging as soon as you review! Thanks a lot @valsdav!

PR #969, opened 2023-12-14
build(deps): pin to uproot 5.2.0

lgray:


lgray:
cirrus py38 failure is PyPI just being grumpy. Ignoring.

PR #978, opened 2024-01-04
feat: preprocess a dataset's base form and hint it to uproot

lgray:
This results in significant speedups when processing large numbers of datasets on files hosted in xrootd.

### DIFF
--- a/src/coffea/dataset_tools/preprocess.py
+++ b/src/coffea/dataset_tools/preprocess.py
@@ -265,7 +265,7 @@ def preprocess(
             calculate_form=calculate_form,
         )
 
-    all_processed_files, = dask.compute(files_to_preprocess, scheduler=scheduler)
+    (all_processed_files,) = dask.compute(files_to_preprocess, scheduler=scheduler)
 
     for name, processed_files in all_processed_files.items():
         not_form = processed_files[["file", "object_path", "steps", "uuid"]]
--- a/src/coffea/dataset_tools/preprocess.py
+++ b/src/coffea/dataset_tools/preprocess.py
@@ -16,14 +16,14 @@ import uproot
 from coffea.util import compress_form
 
 
-def _get_steps(
+def get_steps(
     normed_files: awkward.Array | dask_awkward.Array,
     maybe_step_size: int | None = None,
     align_clusters: bool = False,
     recalculate_seen_steps: bool = False,
     skip_bad_files: bool = False,
     file_exceptions: Exception | Warning | tuple[Exception | Warning] = (OSError,),
-    calculate_form: bool = False,
+    save_form: bool = False,
 ) -> awkward.Array | dask_awkward.Array:
     """
     Given a list of normalized file and object paths (defined in uproot), determine the steps for each file according to the supplied processing options.
@@ -43,7 +43,7 @@ def _get_steps(
             Instead of failing, catch exceptions specified by file_exceptions and return null data.
         file_exceptions: Exception | Warning | tuple[Exception | Warning], default (OSError,)
             What exceptions to catch when skipping bad files.
-        calculate_form: bool, default False
+        save_form: bool, default False
             Extract the form of the TTree from the file so we can skip opening files later.
 
     Returns
@@ -70,12 +70,12 @@ def _get_steps(
 
         form_json = None
         form_hash = None
-        if calculate_form:
+        if save_form:
             form_bytes = (
                 uproot.dask(tree, ak_add_doc=True).layout.form.to_json().encode("utf-8")
             )
 
-            form_hash = hashlib.blake2b(form_bytes).hexdigest()[:16]
+            form_hash = hashlib.md5(form_bytes).hexdigest()
             form_json = gzip.compress(form_bytes)
 
         target_step_size = num_entries if maybe_step_size is None else maybe_step_size
@@ -119,7 +119,7 @@ def _get_steps(
                 "steps": out_steps,
                 "uuid": out_uuid,
                 "form": form_json,
-                "form_hash": form_hash,
+                "form_hash_md5": form_hash,
             }
         )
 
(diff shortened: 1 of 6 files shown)


lgray:
@nsmith- would appreciate your thoughts here!

src/coffea/dataset_tools/preprocess.py
```diff
@@ -19,6 +23,7 @@ def _get_steps(
     recalculate_seen_steps: bool = False,
     skip_bad_files: bool = False,
     file_exceptions: Exception | Warning | tuple[Exception | Warning] = (OSError,),
+    calculate_form: bool = False,
```
nsmith-: This is more of a comment for the earlier PR but given the thorough docstring here, do we want to promote `get_steps` to be public?

nsmith-: On this PR though, I feel like `cache_form` might be more clear?

src/coffea/dataset_tools/preprocess.py
```diff
@@ -61,6 +68,16 @@ def _get_steps(
...
         num_entries = tree.num_entries
 
+        form_json = None
+        form_hash = None
+        if calculate_form:
+            form_bytes = (
+                uproot.dask(tree, ak_add_doc=True).layout.form.to_json().encode("utf-8")
+            )
+
+            form_hash = hashlib.blake2b(form_bytes).hexdigest()[:16]
```
nsmith-: Why blake2b and only the first 8 bytes?

src/coffea/dataset_tools/preprocess.py
```diff
@@ -101,13 +118,22 @@ def _get_steps(
                 "object_path": arg.object_path,
                 "steps": out_steps,
                 "uuid": out_uuid,
+                "form": form_json,
+                "form_hash": form_hash,
```
nsmith-: Having the hash algo in the name would be useful (e.g. `form_hash_md5` or `form_hash_sha512` etc)

src/coffea/dataset_tools/preprocess.py
```diff
@@ -147,12 +173,14 @@ class DatasetSpecOptional:
...
     )
     metadata: dict[Hashable, Any] | None
+    form: str | None
 
 
 @dataclass
 class DatasetSpec:
     files: dict[str, CoffeaFileSpec]
     metadata: dict[Hashable, Any] | None
+    form: str | None
```
nsmith-: Where is `form_hash`?

src/coffea/dataset_tools/preprocess.py
```diff
@@ -147,12 +173,14 @@ class DatasetSpecOptional:
         dict[str, str] | list[str] | dict[str, UprootFileSpec | CoffeaFileSpecOptional]
     )
     metadata: dict[Hashable, Any] | None
+    form: str | None
```
nsmith-, line 174: Can `DatasetSpecOptional` inherit from `DatasetSpec`?

src/coffea/dataset_tools/preprocess.py
```diff
@@ -229,18 +262,48 @@ def preprocess(
...
             skip_bad_files=skip_bad_files,
             file_exceptions=file_exceptions,
+            calculate_form=calculate_form,
         )
 
-    all_processed_files = dask.compute(files_to_preprocess)[0]
+    (all_processed_files,) = dask.compute(files_to_preprocess, scheduler=scheduler)
 
     for name, processed_files in all_processed_files.items():
+        not_form = processed_files[["file", "object_path", "steps", "uuid"]]
```
nsmith-: `processed_files_without_forms`?

src/coffea/util.py
```diff
@@ -195,3 +197,15 @@ def rewrap_recordarray(layout, depth, data):
     if isinstance(layout, awkward.layout.RecordArray):
         return lambda: data
     return None
+
+
+# shorthand for compressing forms
+def compress_form(formjson):
+    return binascii.b2a_base64(
```
nsmith-: I have a personal preference for `base64.base64encode` but I can't see any real difference between them https://stackoverflow.com/questions/20927325/why-is-there-a-difference-between-binascii-b2a-base64-and-base64-b64encode

nsmith-:
I feel like the DatasetSpec could it self have methods like `.normalize()` or something, so you could construct it from some variety of ways and then choose what to cache.
Then maybe it is more straightforward to invalidate parts by simply setting them `dataset.cached_form = None` or `del dataset.steps` etc.

src/coffea/dataset_tools/preprocess.py
lgray: It's not really a cache until it hits the Fileset spec, so I think `save_form` is more accurate?

src/coffea/dataset_tools/preprocess.py
lgray: because blake2b is fast for large input data and the first 8 bytes are random enough to ensure uniqueness (typically), just trying to save some network bandwidth and keep things reasonably appetizing to human eyes.

src/coffea/dataset_tools/preprocess.py
lgray: will do!

src/coffea/dataset_tools/preprocess.py
lgray: need to add!

src/coffea/dataset_tools/preprocess.py
lgray: Yeah I can reorganize this, I forgot that python was all virtual inheritance.

src/coffea/dataset_tools/preprocess.py
lgray: sure

lgray:
For your higher level comment I'll address it in another PR. I think it's a good idea but requires changes outside the functionality being introduced here.

PR #991, opened 2024-01-14
refactor!: back nanoevents.methods.vector with scikit-hep vector

### LINKED ISSUE #992: Use scikit-hep vector to as the backend for vector math in nanoevents
We had discussed this migration quite some time ago, and we think it's the right way to go since nanoevents vector definitely influenced scikit-hep's vector.

We also need to understand how to properly deal with `nearest` and `metric_table`. It might just be better to promote those to free functions and then we can just have the higher level physics concept classes inherit directly from scikit-hep vector.

In addition to the above which would be a rather breaking change, we'll have to be careful to plan out a few deprecation cycles so that users are caught unaware by small differences like `delta_phi` vs. `deltaphi` and similar.

lgray:
Fixes #992

Right now this is breaking in terms of some low-level definitions of physics variables, in particular:
- `rho` is now *always* transverse, as opposed to it being previously `r`, this can certainly screw up math if you're not expecting it

Breaking interface differences (as you can see in tests):
- we're now beholden to vector's much more strict typing of vectors you have to use `to_VectorND()` to correctly compare vectors.

But that doesn't seem to have too many side effects?

However, all the tests pass and the current implementation is backwards compatible in terms of `delta_phi` vs `deltaphi` and `delta_r` vs. `deltaR` so we can actually deprecate that in a reasonable way.

Another thing we need to address is where the functions `nearest` and `metric_table` should live, which are fairly often used but could be advertised better.

@nsmith- This will require a thorough review as well as contribution from you, and we should discuss how we want to roll this out.


lgray:
Fixes #992

src/coffea/nanoevents/methods/vector.py
```diff
@@ -669,53 +641,14 @@ def pvec(self):
...
-    def rapidity(self):
-        pz = self.z
-        e = self.energy
-        return 0.5 * (numpy.log(e + pz) - numpy.log(e - pz))
-
     @property
     def boostvec(self):
         """The `x`, `y` and `z` components divided by `t` as a `ThreeVector`
 
         This can be used for boosting. For cases where `|t| <= rho`, this
```
nsmith-: This is now `r` in the docstring, right?

tests/test_nanoevents_vector.py
```diff
@@ -444,12 +444,18 @@ def test_lorentz_vector_numba(a_dtype, b_dtype):
...
         110.68423555321688,
     ]
-    assert pytest.approx(a.delta_phi(b), abs=1e-7) == [
-        0.03369510734601633,
-        -0.1798534997924781,
-        0.33292327383538156,
-        0.6078019961139605,
-    ]
+
+    computed_dphi = a.delta_phi(b).to_numpy()
```
nsmith-, line 457: Is the explicit cast required now?

tests/test_nanoevents_vector.py
```diff
@@ -444,12 +444,18 @@ def test_lorentz_vector_numba(a_dtype, b_dtype):
...
+
+    assert pytest.approx(computed_dphi, abs=1e-6) == np.array(
```
nsmith-, line 459: Any idea why the tolerance had to be increased?

tests/test_nanoevents_vector.py
```diff
@@ -532,9 +538,9 @@ def test_inherited_method_transpose(lcoord, threecoord, twocoord):
     assert record_arrays_equal(a.delta_phi(c), c.delta_phi(a))
     assert record_arrays_equal(b.delta_phi(c), c.delta_phi(b))
 
-    assert record_arrays_equal(a - b, -(b - a))
-    assert record_arrays_equal(a - c, -(c - a))
-    assert record_arrays_equal(b - c, -(c - b))
+    assert record_arrays_equal((a - b).to_Vector3D(), -(b - a))
+    assert record_arrays_equal((a - c).to_Vector2D(), -(c - a))
+    assert record_arrays_equal((b - c).to_Vector2D(), -(c - b))
```
nsmith-: `a` is a 4-vector, `b` a 3-vector, and `c` a 2-vector. So I guess the type demotion behavior depends on only the first argument of the infix operator here? Is this expected?

src/coffea/nanoevents/methods/vector.py
```diff
@@ -91,10 +97,39 @@ def delta_r(eta1, phi1, eta2, phi2):
 
 
 behavior = {}
+behavior.update(vector.backends.awkward.behavior)
+
+
+class TwoVectorArray:
```
nsmith-: What do these empty classes do?

nsmith- (changes requested):
I guess the only painful part is `r` vs. `rho`. Otherwise looks like it's drag and drop. @jrueb might want to review this as well.

lgray:
unfortunately @jrueb has left academia for industry, unless he wants to review out of the kindness of his heart.

src/coffea/nanoevents/methods/vector.py
lgray: These satisfy the linter and are overwritten by `@awkward.mixin_class(behavior)`, otherwise flake complains the variable doesn't exist since it's patched in and I have to assign the projection classes.

tests/test_nanoevents_vector.py
lgray: @saransh-cpp do we expect this (3D to 2D vector demotion / promotion)?

Saransh-cpp: Yes, the behavior is uniform across all backends.

tests/test_nanoevents_vector.py
lgray: It needs further investigation but yeah it looks like there's some conversion going on or something that we should investigate. 

This is also related to the above, I was just making sure there wasn't weird roundoff error happening or anything.

src/coffea/nanoevents/methods/vector.py
lgray: ah yeah.

tests/test_nanoevents_vector.py
lgray: In that you expect when you do `==` between two vectors you have to cast them to the same dimensionality first?


@nsmith- what was the expected behavior in nanoevents.methods.vector?

tests/test_nanoevents_vector.py
lgray: see the comment on approx.

src/coffea/nanoevents/methods/vector.py
lgray: Actually, also need to make sure that the sign in the metric is the same for coffea vs. scikit-hep vector.

tests/test_nanoevents_vector.py
Saransh-cpp: Yes, for `==` the vectors should have same dimensions. `to_Vector*D` or `to_{coordinate_names}` methods can be used to change the dimensions. `to_Vector*D` additionally allows users to update the newly created coordinate (when the vector's dimension is promoted).

lgray:
I'm putting in a deprecation for ~6 months from now over in #997.

tests/test_nanoevents_vector.py
nsmith-: So `-(c-a)` being automatically demoted to a 2-vector makes sense, but I'm puzzled why `(a-c)` is _not_ and requires an explicit cast.

lgray: My opinion is you should promote to the highest vector in the operation and then cast down in dimensionality as you need. That has the most reasonable behavior from my perspective.

However right now we do neither!

ikrommyd:
@lgray Is the plan to keep the public `delta_phi` and `delta_r` functions as is? I suggest you do because there are some cases where a user defined "eta" instead of the `.eta` attribute is needed

lgray:
In the first phase of this, yes we'll keep those functions for backwards compatibility. 

In the long run, when we remove `nanoevents.methods.vector`, vector supplies `deltaR` and `deltaphi` and folks should migrate to use what is already supplied. You can see we now just use those functions within their underscore equivalents.

src/coffea/nanoevents/methods/vector.py
lgray: changed this to use noqa instead, it's more readable.

tests/test_nanoevents_vector.py
nsmith-: I disagree. By promoting the 2-vector to a 3- or 4-vector, you're assuming a value of 0 for the missing dimension(s). If someone wants to add MET to a 4-vector they should have to explicitly upcast it from a 2-vector and define explicitly the pseudorapidity and mass.
I also don't think that `a - c` and `c - a` should result in different types.

lgray: `a - c` vs. `c - a` implies some sort of promotion or demotion has to happen. So you are asking that we always demote?

Then common patterns like `some_4_vector + 2_vector_spatial_rotation` no longer result in a 4-vector, which would be very confusing.

lgray: Oh, I misread. You're asking that in order to add a rotation someone does `4_vector + 2_vector.to_Vector4D()` always and specifically, or something even different from that? This is a breaking change we should make sure people know it is coming for the shim-in.

However, if we don't require that explicit dimensionality cast, that means we must choose which direction we go, since `a-c` vs. `c-a` have different starting types.
Do we always choose 0s for the lower dim vector?

nsmith-: Yes I think we should always demote. I don't understand what you mean by adding a rotation. A rotation depends on the current vector and some (euler) angles. Anyway the current coffea vectors demote always:
```
>>> a-c
<TwoVectorArray [{x: 20, y: -22}, ..., {x: 15, ...}] type='3 * TwoVector[x:...'>
>>> c-a
<TwoVectorArray [{x: -20, y: 22}, ..., {x: -15, ...}] type='3 * TwoVector[x...'>
>>> a+c
<TwoVectorArray [{x: 0, y: 2}, {...}, {x: 45, ...}] type='3 * TwoVector[x: ...'>
>>> c+a
<TwoVectorArray [{x: 0, y: 2}, {...}, {x: 45, ...}] type='3 * TwoVector[x: ...'>
>>> b-c
<TwoVectorArray [{x: 0, y: -22}, ..., {x: -45, ...}] type='3 * TwoVector[x:...'>
>>> b+c
<TwoVectorArray [{x: -20, y: 2}, ..., {x: -15, ...}] type='3 * TwoVector[x:...'>
>>> c-b
<TwoVectorArray [{x: 0, y: 22}, ..., {x: 45, ...}] type='3 * TwoVector[x: f...'>
>>> c+b
<TwoVectorArray [{x: -20, y: 2}, ..., {x: -15, ...}] type='3 * TwoVector[x:...'>
>>> a+b
<ThreeVectorArray [{x: 0, y: -20, z: 10}, ..., {...}] type='3 * ThreeVector...'>
```

lgray: Arg, morning brain glitch(re: rotations).

nsmith-: @Saransh-cpp I think this is just a bug in vector: https://github.com/scikit-hep/vector/issues/412

lgray: There's also an override in the nanoevents classes, that right now has weird behavior, we can just enforce it correctly there. That's what I was planning to do.

lgray:
@nsmith- tests fail after appeasing linter :-(

nsmith-:
Yes, I added tests and am now working to improve them

lgray:
Ah, gotcha - yeah putting this together with https://github.com/scikit-hep/vector/pull/413 I figured you may be messing around.

lgray:
I think one place where we could benefit a lot from a thorough look-over by @Saransh-cpp is to make sure we're not unintentionally clobbering any of the vector interface with parts of nanoevents.vector, and likewise @nsmith- if there's anything we can do to make the delta between skh-vector-backed nanoevents.vector closer to the original implementation we should try to smooth all that out.

lgray:
@Saransh-cpp @nsmith- what else needs to happen on this PR for validation and to prepare folks for switching? Does a release target for March still make sense?

nsmith-:
Where I left off is that we are now properly casting, but there is possibility to remove more of our methods: for example all `__add__` `__subtract__` `__multiply__` mixin methods could point to the vector implementations, but we would have to manually register each.

Saransh-cpp:
Do you mean changing all the operations in mixin classes from this -
```py
def add(self, other):
    """Add two vectors together elementwise using `x`, `y`, and `z` components"""
    return awkward.zip(
        {
            "x": self.x + other.x,
            "y": self.y + other.y,
            "z": self.z + other.z,
        },
        with_name="ThreeVector",
        behavior=self.behavior,
    )
```
to this? -
```py
def __add__(self, other):
    """Add two vectors together elementwise using `x`, `y`, and `z` components"""
    return super().add(other)
```
Or, could we eliminate these methods and fall back to the inherited methods? That should ideally work too.

This will be  much cleaner IMO and will also move "skh-vector-backed nanoevents.vector closer to the original implementation" 

I can pick this up if required!

nsmith-:
I mean to eliminate the methods and fall back to the scikit-hep vector methods. If you have time to take it, please go ahead!

src/coffea/nanoevents/methods/vector.py
```diff
@@ -1061,6 +912,52 @@ def negative(self):
...
+LorentzVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
+LorentzVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821
+
+PtEtaPhiMLorentzVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
+PtEtaPhiMLorentzVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
+PtEtaPhiMLorentzVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821
+
+PtEtaPhiELorentzVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
+PtEtaPhiELorentzVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
+PtEtaPhiELorentzVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821
```
Saransh-cpp: Should this be? -
```suggestion
TwoVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
TwoVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
TwoVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821

PolarTwoVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
PolarTwoVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
PolarTwoVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821

ThreeVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
ThreeVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
ThreeVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821

SphericalThreeVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
SphericalThreeVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
SphericalThreeVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821

LorentzVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
LorentzVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
LorentzVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821

PtEtaPhiMLorentzVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
PtEtaPhiMLorentzVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
PtEtaPhiMLorentzVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821

PtEtaPhiELorentzVectorArray.ProjectionClass2D = TwoVectorArray  # noqa: F821
PtEtaPhiELorentzVectorArray.ProjectionClass3D = ThreeVectorArray  # noqa: F821
PtEtaPhiELorentzVectorArray.ProjectionClass4D = LorentzVectorArray  # noqa: F821
```

Saransh-cpp (commented):
I tried removing the `add` and `subtract` methods locally and noticed something peculiar (notice the types and the coordinate names) -

## This branch
```py
In [3]: a = ak.zip(
   ...:             {
   ...:                 "pt": [10.0, 20.0, 30.0],
   ...:                 "eta": [0.0, 1.1, 2.2],
   ...:                 "phi": [0.1, 0.9, -1.1],
   ...:                 "mass": [1.0, 1.0, 1.0],
   ...:             },
   ...:             with_name="PtEtaPhiMLorentzVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [4]: b = ak.zip(
   ...:             {
   ...:                 "rho": [10.0, 20.0, 30.0],
   ...:                 "theta": [0.3, 0.6, 1.1],
   ...:                 "phi": [-3.0, 1.1, 0.2],
   ...:             },
   ...:             with_name="SphericalThreeVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [5]: c = ak.zip(
   ...:             {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
   ...:             with_name="PolarTwoVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [6]: a + b
Out[6]: <ThreeVectorArray [{x: 0.0501, y: -0.413, ...}, ...] type='3 * ThreeVector[...'>

In [7]: a + c
Out[7]: <TwoVectorArray [{x: 6.51, y: -8.39}, {...}, {...}] type='3 * TwoVector[x: ...'>

In [8]: c + a
Out[8]: <TwoVectorArray [{x: 6.51, y: -8.39}, {...}, {...}] type='3 * TwoVector[x: ...'>

In [9]: b + a
Out[9]: <ThreeVectorArray [{x: 0.0501, y: -0.413, ...}, ...] type='3 * ThreeVector[...'>
```

## Removing the `add` and `subtract` methods
```py
In [3]: a = ak.zip(
   ...:             {
   ...:                 "pt": [10.0, 20.0, 30.0],
   ...:                 "eta": [0.0, 1.1, 2.2],
   ...:                 "phi": [0.1, 0.9, -1.1],
   ...:                 "mass": [1.0, 1.0, 1.0],
   ...:             },
   ...:             with_name="PtEtaPhiMLorentzVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [4]: b = ak.zip(
   ...:             {
   ...:                 "rho": [10.0, 20.0, 30.0],
   ...:                 "theta": [0.3, 0.6, 1.1],
   ...:                 "phi": [-3.0, 1.1, 0.2],
   ...:             },
   ...:             with_name="SphericalThreeVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [5]: c = ak.zip(
   ...:             {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
   ...:             with_name="PolarTwoVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [6]: a + b
Out[6]: <ThreeVectorArray [{x: 0.0501, y: -0.413, ...}, ...] type='3 * ThreeVector[...'>

In [7]: b + a
Out[7]: <SphericalThreeVectorArray [{x: 0.0501, y: -0.413, ...}, ...] type='3 * Sph...'>

In [8]: a + c
Out[8]: <TwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * TwoVector[...'>

In [9]: c + a
Out[9]: <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
```

## Removing `add` and `subtract` and incorporating the suggestion below
```py
In [3]: a = ak.zip(
   ...:             {
   ...:                 "pt": [10.0, 20.0, 30.0],
   ...:                 "eta": [0.0, 1.1, 2.2],
   ...:                 "phi": [0.1, 0.9, -1.1],
   ...:                 "mass": [1.0, 1.0, 1.0],
   ...:             },
   ...:             with_name="PtEtaPhiMLorentzVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [4]: b = ak.zip(
   ...:             {
   ...:                 "rho": [10.0, 20.0, 30.0],
   ...:                 "theta": [0.3, 0.6, 1.1],
   ...:                 "phi": [-3.0, 1.1, 0.2],
   ...:             },
   ...:             with_name="SphericalThreeVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [5]: c = ak.zip(
   ...:             {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
   ...:             with_name="PolarTwoVector",
   ...:             behavior=vector.behavior,
   ...:         )

In [6]: a = b

In [7]: a + b
Out[7]: <ThreeVectorArray [{rho: 20, phi: -3, ...}, ..., {...}] type='3 * ThreeVect...'>

In [8]: b + a
Out[8]: <ThreeVectorArray [{rho: 20, phi: -3, ...}, ..., {...}] type='3 * ThreeVect...'>

In [9]: a + c
Out[9]: <TwoVectorArray [{rho: 17.2, phi: -2.46}, ..., {...}] type='3 * TwoVector[r...'>

In [10]: c + a
Out[10]: <TwoVectorArray [{rho: 17.2, phi: -2.46}, ..., {...}] type='3 * TwoVector[r...'>
```

I'm not sure which one of these is correct, could you please let me know which one is? All of this is most probably on coffea's side and not on vector's side.

src/coffea/nanoevents/methods/vector.py
nsmith-: So obviously whatever solution we pick should make sure the array type matches the available coordinates.
That said, there is a choice here, whether:
- `SphericalThreeVector + ThreeVector = ThreeVector`, or
- `SphericalThreeVector + ThreeVector = SphericalThreeVector`

and similarly whether transposing the arguments maintains this casting.

Personally, and I've mentioned this to @jpivarski in the past, I would have had everything convert to cartesian on addition, even as far as to have `SphericalThreeVector + SphericalThreeVector = ThreeVector` (as coffea does now). The motivation is that adding is simplest in cartesian coordinates and I fear that doing it in other coordinates may inflate the effects of numerical imprecision in some cases. This would need to be carefully studied of course.

For now, I would advocate for add/subtract to return cartesian coordinates.

jpivarski: In Vector, longitudinal and temporal coordinates each convert to Cartesian (`z` and `t`, respectively); it's only the polar azimuthal coordinates that stay polar if both addends are already polar. That's [this function](https://github.com/scikit-hep/vector/blob/e4c3f536d8d3833ba8b31453dc7313b5b3cc5f56/src/vector/_compute/planar/add.py#L46-L54):

```python
# specialized
def rhophi_rhophi(lib, rho1, phi1, rho2, phi2):
    diff = phi2 - phi1
    u = rho2 * lib.cos(diff)
    v = rho2 * lib.sin(diff)
    return (
        lib.sqrt((rho1 + u) ** 2 + v**2),
        rectify(lib, phi1 + lib.arctan2(v, rho1 + u)),
    )
```

If this is bad (no study was performed—what metric would define "goodness"?[^1]), then it can be replaced by

```python
def rhophi_rhophi(lib, rho1, phi1, rho2, phi2):
    return xy_xy(
        lib,
        x.rhophi(lib, rho1, phi1),
        y.rhophi(lib, rho1, phi1),
        x.rhophi(lib, rho2, phi2),
        y.rhophi(lib, rho2, phi2),
    )
```

and similarly for [subtract](https://github.com/scikit-hep/vector/blob/e4c3f536d8d3833ba8b31453dc7313b5b3cc5f56/src/vector/_compute/planar/subtract.py#L46-L54).

[^1]: Arguably, it could just be the count of transcendental functions. The `x.rhophi` and `y.rhophi` functions each have one `sin` and `cos`, so that's four, and the specialized function has a `cos`, `sin`, `sqrt`, and `arctan2`, but also the `rectify`, which uses `%` to keep the phi from wrapping around. I bet the conversion to Cartesian is faster for all inputs. To quantify round-off error, it might depend on what the inputs are (a 4D space of two pairs of azimuthal coordinates).

Saransh-cpp: Thanks @nsmith-  and @jpivarski! I'd be happy to change the way things are handled in vector or keep them as it is, but in any case, I think we should get rid off `add`, `subtract`, and other methods in coffea's vector classes. Please let me know if vector should always convert azimuthal coordinates to cartesian.

lgray:
@nsmith- @Saransh-cpp from the looks of it we are now close to being able to merge this change. Please provide reasons not to do so yet, or any further tests that must be done before proceeding.

src/coffea/nanoevents/methods/vector.py
nsmith-: Can you try an experiment? Sample randomly two Lorentz vectors in cartesian space (set `t^2=x^2+y^2+z^2+m^2` for some randomly sampled m), then compare the result of: adding them; converting them to azimuthal coordinates and adding them? The sampling distribution could just be gaussian with a large (~50) width.

jpivarski: Accuracy and/or speed?

For accuracy, note that among Python, NumPy, and Awkward Array, there are either 2 or 3 summation algorithms that differ in the least significant bits, so choice of backend matters.

For speed, naturally big arrays are more representative.

**Edit:** I'm hooked. I'll give it a try.

Since this is a question about how only the azimuthal coordinates are handled, I'm going to do the test in x and y only. That will focus on the part that might or might not be changed.

jpivarski: Accuracy with NumPy:

```python
>>> import numpy as np
>>> import awkward as ak
>>> import vector
>>> vector.register_awkward()
>>> 
>>> source_ax = np.random.normal(0, 1, 1000000)
>>> source_ay = np.random.normal(0, 1, 1000000)
>>> source_bx = np.random.normal(0, 1, 1000000)
>>> source_by = np.random.normal(0, 1, 1000000)
>>> 
>>> cartesian_a = vector.array({"x": source_ax, "y": source_ay})
>>> cartesian_b = vector.array({"x": source_bx, "y": source_by})
>>> cartesian_sum = cartesian_a + cartesian_b
>>> 
>>> polar_a = vector.array({"rho": cartesian_a.rho, "phi": cartesian_a.phi})
>>> polar_b = vector.array({"rho": cartesian_b.rho, "phi": cartesian_b.phi})
>>> polar_sum = polar_a + polar_b
>>> 
>>> abs(cartesian_sum - polar_sum).min()
0.0
>>> abs(cartesian_sum - polar_sum).mean()
4.489121610871359e-16
>>> abs(cartesian_sum - polar_sum).max()
5.748513452393584e-15
```

Accuracy with Awkward:

```python
>>> cartesian_a = ak.zip({"x": source_ax, "y": source_ay}, with_name="Vector2D")
>>> cartesian_b = ak.zip({"x": source_bx, "y": source_by}, with_name="Vector2D")
>>> cartesian_sum = cartesian_a + cartesian_b
>>> 
>>> polar_a = ak.zip({"rho": cartesian_a.rho, "phi": cartesian_a.phi}, with_name="Vector2D")
>>> polar_b = ak.zip({"rho": cartesian_b.rho, "phi": cartesian_b.phi}, with_name="Vector2D")
>>> polar_sum = polar_a + polar_b
>>> 
>>> ak.min(abs(cartesian_sum - polar_sum))
0.0
>>> ak.mean(abs(cartesian_sum - polar_sum))
4.489131490391888e-16
>>> ak.max(abs(cartesian_sum - polar_sum))
5.748513452393584e-15
```

Accuracy with Python:

```python
>>> cartesian_a = [vector.obj(x=x, y=y) for x, y in zip(source_ax, source_ay)]
>>> cartesian_b = [vector.obj(x=x, y=y) for x, y in zip(source_bx, source_by)]
>>> cartesian_sum = [a + b for a, b in zip(cartesian_a, cartesian_b)]
>>> 
>>> polar_a = [vector.obj(rho=a.rho, phi=a.phi) for a in cartesian_a]
>>> polar_b = [vector.obj(rho=b.rho, phi=b.phi) for b in cartesian_b]
>>> polar_sum = [a + b for a, b in zip(polar_a, polar_b)]
>>> 
>>> tmp = [abs(car - pol) for car, pol in zip(cartesian_sum, polar_sum)]
>>> np.min(tmp)
0.0
>>> np.mean(tmp)
4.489131490397885e-16
>>> np.max(tmp)
5.748513452393584e-15
```

Although there were a million elementwise `a`, `b` pairs, these accuracy measurements represent 1 numerical step. The error would grow if this is compounded in a loop. (But when are you going to have coordinates changing between Cartesian and polar in a loop? Once some operation puts it in Cartesian, it will remain in Cartesian; that's an absorbing state.)

**Edit:** do-over with the same random numbers in all three studies, to see if any algorithms differ. It seems that they don't! (Okay, Awkward and NumPy only differ in repeated summations, as in `ak.sum`, scikit-hep/awkward#1250. It's not a surprise that NumPy and Python agree, since `vector.obj` functions [are computed with](https://github.com/scikit-hep/vector/blob/6ee3a1f81f02ce8013c6a766860ed05009d672fe/src/vector/backends/object.py#L336) `lib=numpy`, rather than `math` or something.)

jpivarski: As for time, performing the calculation in polar is only 2% slower than converting to Cartesian, then adding.

```python
>>> import numpy as np
>>> import vector
>>> import time
>>> source_ax = np.random.normal(0, 1, 100000000)
>>> source_ay = np.random.normal(0, 1, 100000000)
>>> source_bx = np.random.normal(0, 1, 100000000)
>>> source_by = np.random.normal(0, 1, 100000000)
>>> 
>>> cartesian_a = vector.array({"x": source_ax, "y": source_ay})
>>> cartesian_b = vector.array({"x": source_bx, "y": source_by})
>>> polar_a = vector.array({"rho": cartesian_a.rho, "phi": cartesian_a.phi})
>>> polar_b = vector.array({"rho": cartesian_b.rho, "phi": cartesian_b.phi})
>>> 
>>> t0 = time.perf_counter(); tmp = polar_a + polar_b; time.perf_counter() - t0
6.290486244004569
>>> t0 = time.perf_counter(); tmp = polar_a + polar_b; time.perf_counter() - t0
6.331587403001322
>>> t0 = time.perf_counter(); tmp = polar_a + polar_b; time.perf_counter() - t0
6.302776351003558
>>> 
>>> t0 = time.perf_counter(); tmp = vector.array({"x": polar_a.x, "y": polar_a.y}) + vector.array({"x": polar_b.x, "y": polar_b.y}); time.perf_counter() - t0
6.182807716999378
>>> t0 = time.perf_counter(); tmp = vector.array({"x": polar_a.x, "y": polar_a.y}) + vector.array({"x": polar_b.x, "y": polar_b.y}); time.perf_counter() - t0
6.175768633998814
>>> t0 = time.perf_counter(); tmp = vector.array({"x": polar_a.x, "y": polar_a.y}) + vector.array({"x": polar_b.x, "y": polar_b.y}); time.perf_counter() - t0
6.201813601001049
```

By comparison, if it had been Cartesian to begin with, that's 8.8× faster:

```python
>>> t0 = time.perf_counter(); tmp = cartesian_a + cartesian_b; time.perf_counter() - t0
0.7100841389983543
>>> t0 = time.perf_counter(); tmp = cartesian_a + cartesian_b; time.perf_counter() - t0
0.710351111003547
>>> t0 = time.perf_counter(); tmp = cartesian_a + cartesian_b; time.perf_counter() - t0
0.6980755539989332
```

An argument could be made to convert to Cartesian before adding, not to speed up the addition, but to speed up any _subsequent_ operations. It has long-range performance repercussions, not short-range.

But leaving polar coordinates in polar was motivated by the principle of least surprise. An expert user might know that converting to Cartesian is better for their particular computation in the long run, but it's best to leave that decision to them.

jpivarski: These are all of the constituent parts, the cost of each operation (ignoring memory access):

```python
>>> numbers = abs(source_ax)
>>> t0 = time.perf_counter(); tmp = numbers + numbers; time.perf_counter() - t0
0.11195918399607763
>>> t0 = time.perf_counter(); tmp = numbers * numbers; time.perf_counter() - t0
0.11245346799842082
>>> t0 = time.perf_counter(); tmp = numbers**2; time.perf_counter() - t0
0.11356563799927244
>>> t0 = time.perf_counter(); tmp = np.sin(numbers); time.perf_counter() - t0
0.8857823760044994
>>> t0 = time.perf_counter(); tmp = np.cos(numbers); time.perf_counter() - t0
0.9174221430002945
>>> t0 = time.perf_counter(); tmp = np.sqrt(numbers); time.perf_counter() - t0
0.16433826299908105
>>> t0 = time.perf_counter(); tmp = np.arctan2(numbers, numbers); time.perf_counter() - t0
1.1808518600009847
>>> def rectify(lib, phi):
...     return (phi + lib.pi) % (2 * lib.pi) - lib.pi
... 
>>> t0 = time.perf_counter(); tmp = rectify(np, numbers); time.perf_counter() - t0
0.5352489489960135
```

The additions and multiplications are likely dominated by memory access, so 0.11 seconds is the floor. `np.arctan2` is a large part of the 6.3 seconds.

Saransh-cpp (approved):
This looks good to me in this state!

src/coffea/nanoevents/methods/vector.py
nsmith-: Ok, thanks for checking. I'm ok then with leaving it to the user to decide when to change coordinates. This does mean that the current coffea vector behavior of `PtEtaPhiMLorentzVector + PtEtaPhiMLorentzVector = LorentzVector`will change. We are opting in to this now.

nsmith- (dismissed):

lgray:
This fails on a real user analysis with:
```
File "/Users/lgray/coffea-dev/ewkcoffea/ewkcoffea/modules/selection_wwz.py", line 334, in get_mt2
    rest_WW = w_lep0 + w_lep1 + misspart
...
TypeError: no numpy.add overloads for custom types: Candidate, PtEtaPhiMLorentzVector
```

Please update and harden tests before we merge. This appears to be on the coffea side of things.

@Saransh-cpp @nsmith-

nsmith-:
Ah, this is going to be an issue. We have a class hierarchy that in principle would all need dispatch updates, so fixing it for `Candidate` just means that now it will fail for `Electron` or `Jet`. I'm kind of puzzled why awkward cannot do the type resolution automatically.

nsmith-:
So #1063 is just pushing the issue one level down.

Saransh-cpp:
Oh yes, I missed that, thanks! I don't think there is a better way to do this? Should I add dispatch updates for `Electron` and `Jet` in a similar way?

Saransh-cpp:
More tests are failing, and I think there are more classes for which we'll have to do the same - `Particle`, `Muon`, `MasslessParticle`, `Photon`, `Track`, `Tower`, `Vertex`, `MissingET` - everything inheriting the coffee vector classes.

nsmith-:
@jpivarski is there a way to have awkward dispatch using the method resolution order of the mixin types, rather than having to register every subtype explicitly?

jpivarski:
The `ak.behavior` → Python class object mapping (that is, through the `__record__` parameter name) has to map to the concrete classes, but those concrete classes can share superclasses. If we're talking about getting methods and properties on class instances, those methods and properties can be inherited onto the concrete class, but the concrete class must be the one that `ak.behavior` points to.

For ufunc and binary operator overloading, it's still the concrete classes that have to be named, but those a free functions, rather than methods and properties, so there isn't an easy way for inheritance to work. If you want to make `np.add` call function `f` for all subclasses of `T`, you need to put signatures with each multiple dispatch combination into the `ak.behavior`. That's why Vector has for loops like this:

https://github.com/scikit-hep/vector/blob/06a733784d8df77d2fe4ad0f87051f725fd6a831/src/vector/backends/awkward.py#L1519-L1546

@agoose77 and I brought this up several times—how to describe inheritance in `ak.behavior` so that a single line

```python
ak.behavior[np.add, "Vector", "Vector"] = lambda v1, v2: v1.add(v2)
```

would effectively do

```python
for left in [names of all subclasses of Vector]:
    for right in [names of all subclasses of Vector]:
        ak.behavior[np.add, left, right] = lambda v1, v2: v1.add(v2)
```

The problem is that the interface of `ak.behavior` gets more complicated: we'd have to tell it what the names of all subclasses of some common class is independently of Python's own class inheritance mechanism—having two class inheritance mechanisms, Python's and Awkward Array's, which can get out of sync, could lead to an amazingly messy situation. These names, `"Vector"`, `"Vector2D"`, etc., are distinct from Python class objects because parameters have to be JSON to be serialized in Forms, passed from C++, to and from Julia, etc.

Having to build these things with for loops is cumbersome, but I'm afraid that having two parallel inheritance mechanisms would be worse in practice.

Saransh-cpp:
Thank you for the explanation! I will add the loops for every vector class in coffee. Additionally, should coffea also have distinct and not the class names in the behavior dict?

jpivarski:
> Additionally, should coffea also have distinct and not the class names in the behavior dict?

I'm not sure what you mean. Is it the distinction between putting behaviors in a global `ak.behaviors` versus putting them in each `ak.Array.behaviors`?

The fact that Coffea used `ak.Array.behaviors` was great for debugging that feature, and `ak.Array` should have that feature. However, when I present Vector, I always do `vector.register_awkward()` once globally after import because then I don't have to show special constructors—Awkward Arrays with Vectors are just ordinary Awkward Arrays. Coffea hasn't had _this_ problem because the data come from files—there has to be a special constructor anyway.

But maybe some of Coffea's metadata handling costs come from passing `behaviors` from one `ak.Array` to another. (The same costs would have been present in Coffea 0.7, but maybe they're exacerbated by Dask.) If the `behaviors` are being passed around as an unexamined dict, then there's no appreciable cost, but if the dicts need to be copied or manipulated, it might add up.

Maybe I'm misunderstanding @Saransh-cpp's question, but either way, that might be something to consider: switching to `coffea.register_awkward()` (if not doing it automatically with `import coffea`).

lgray:
How are we doing on the dependencies to this PR? @Saransh-cpp @nsmith-

lgray:
@Saransh-cpp @nsmith- ping.

Saransh-cpp:
Hi @lgray, sorry for the delay. I have created https://github.com/scikit-hep/awkward/pull/3137 that should simplify registering ufuncs for coffea vectors. I am still writing tests for that PR, but as soon as it is merged and released, the tests in #1063 should start passing.

lgray:
@Saransh-cpp multiple test failures:
```
FAILED tests/test_dataset_tools.py::test_apply_to_fileset[proc_and_schema1] - AttributeError: no field named 'MomentumClass'

    numpy.add.__call__(
        <MuonArray-typetracer [...] type='## * var * Muon[dxy: float32[para...'>
        <MuonArray-typetracer [...] type='## * var * Muon[dxy: float32[para...'>
    )
```

lgray:
@nsmith- This is finally passing - please give it a reasonably thorough review.

jpivarski (approved):
This is such a clean swap: impressive!

lgray:
@nsmith- ping on a review!

src/coffea/nanoevents/methods/candidate.py
```diff
@@ -70,4 +70,23 @@ class PtEtaPhiECandidate(Candidate, vector.PtEtaPhiELorentzVector):
     pass
 
 
+awkward.behavior.update(
```
nsmith-: Why are we updating the global awkward behavior and not our module-local `behavior` object defined on L13?

src/coffea/nanoevents/methods/delphes.py
```diff
@@ -130,6 +130,14 @@ def z(self):
 
 
 _set_repr_name("Vertex")
+awkward.behavior.update(
```
nsmith-: Same comment as before, we are registering in the global namespace when there already is a local namespace that's wired into the NanoEvents

src/coffea/nanoevents/methods/delphes.py
```diff
@@ -130,6 +130,14 @@ def z(self):
...
+    awkward._util.copy_behaviors(vector.LorentzVector, Vertex, behavior)
+)
+
+VertexArray.ProjectionClass2D = vector.TwoVectorArray  # noqa: F821
+VertexArray.ProjectionClass3D = vector.ThreeVectorArray  # noqa: F821
+VertexArray.ProjectionClass4D = VertexArray  # noqa: F821
```
nsmith-, line 138: Is this a way to signal that there is no 4D interpretation of VertexArray? (which is true)

src/coffea/nanoevents/methods/candidate.py
```diff
@@ -70,4 +70,23 @@ class PtEtaPhiECandidate(Candidate, vector.PtEtaPhiELorentzVector):
...
+    awkward._util.copy_behaviors(vector.LorentzVector, Candidate, behavior)
+)
+
+CandidateArray.ProjectionClass2D = vector.TwoVectorArray  # noqa: F821
+CandidateArray.ProjectionClass3D = vector.ThreeVectorArray  # noqa: F821
+CandidateArray.ProjectionClass4D = vector.LorentzVectorArray  # noqa: F821
+CandidateArray.MomentumClass = CandidateArray  # noqa: F821
```
nsmith-, line 78: This block of items is commonly repeated throughout. It would be good to create a utility function to set these projections

tests/test_nanoevents_vector.py
```diff
@@ -519,22 +534,22 @@ def test_inherited_method_transpose(lcoord, threecoord, twocoord):
...
+    assert_allclose(a.delta_phi(b), -b.delta_phi(a))
+    assert_allclose(a.delta_phi(c), -c.delta_phi(a))
+    assert_allclose(b.delta_phi(c), -c.delta_phi(b))
 
-    assert record_arrays_equal(a - b, -(b - a))
-    assert record_arrays_equal(a - c, -(c - a))
-    assert record_arrays_equal(b - c, -(c - b))
+    assert_record_arrays_equal((a.like(b) - b), -(b - a.like(b)), check_type=True)
+    assert_record_arrays_equal((a.like(c) - c), -(c - a.like(c)), check_type=True)
+    assert_record_arrays_equal((b.like(c) - c), -(c - b.like(c)), check_type=True)
```
nsmith-, line 560: I would suggest also testing we error on incompatible types:
```python
with pytest.raises(ValueError):
    a - b == -(b - a)
```

nsmith- (changes requested):
There are a few cases where we are adding to the `ak.behavior` global namespace. In general we try to keep everything in our own namespace. For example, it would break the awkward namespace if someone had a Delphes and NanoAOD file open at the same time since both schemas end up defining a `Jet` object mixin that has a different implementation. I am wondering how tests do not pick up on this, as the copy_behaviors output would not end up in the schema's behavior and I don't think the `ak.behavior` is queried at all for arrays built using NanoEventsFactory.

There are a few more minor comments related to code duplication, and I'm mildly concerned how the tests are all approximate now. In how many cases did we actually lose exact agreement?

lgray:
@Saransh-cpp when you have time can you take care of these review comments? Thanks!

src/coffea/nanoevents/methods/delphes.py
Saransh-cpp: I can just remove L139, but vector would throw an error saying something like "ProjectionClass4D is not defined for VertexArray" if a user tries an operation that needs converting the array to a 4D vector. The error does not tell the user what they are doing wrong, but I am hoping that not a lot of users will try converting it to 4D?

Saransh-cpp:
Thanks for the review!

> In how many cases did we actually lose exact agreement?

I am not sure about this but I can revert the tests to their original state and manually check which ones are failing. I'll report them here.

Saransh-cpp:
Six tests fail if the approximations are removed -

```py
=========================== short test summary info ============================
FAILED tests/test_nanoevents_vector.py::test_polar_two_vector - assert False
FAILED tests/test_nanoevents_vector.py::test_inherited_method_transpose[PolarTwoVector-ThreeVector-PtEtaPhiMLorentzVector] - assert False
FAILED tests/test_nanoevents_vector.py::test_inherited_method_transpose[PolarTwoVector-ThreeVector-PtEtaPhiELorentzVector] - assert False
FAILED tests/test_nanoevents_vector.py::test_inherited_method_transpose[PolarTwoVector-SphericalThreeVector-LorentzVector] - assert False
FAILED tests/test_nanoevents_vector.py::test_inherited_method_transpose[PolarTwoVector-SphericalThreeVector-PtEtaPhiMLorentzVector] - assert False
FAILED tests/test_nanoevents_vector.py::test_inherited_method_transpose[PolarTwoVector-SphericalThreeVector-PtEtaPhiELorentzVector] - assert False
================== 6 failed, 26 passed, 21 warnings in 17.52s ==================
```

<details>
<summary>Trace:</summary>

```console
=================================== FAILURES ===================================
____________________________ test_polar_two_vector _____________________________

    def test_polar_two_vector():
        a = ak.zip(
            {
                "rho": [[1, 2], [], [3], [4]],
                "phi": [[0.3, 0.4], [], [0.5], [0.6]],
            },
            with_name="PolarTwoVector",
            behavior=vector.behavior,
        )
    
>       assert_record_arrays_equal(
            a * 2,
            ak.zip({"rho": [[2, 4], [], [6], [8]], "phi": [[0.3, 0.4], [], [0.5], [0.6]]}),
        )

tests/test_nanoevents_vector.py:77: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PolarTwoVectorArray [[{rho: 2, phi: 0.3}, {...}], ...] type='4 * var * Pol...'>
b = <Array [[{rho: 2, phi: 0.3}, {...}], ...] type='4 * var * {rho: int64, phi:...'>
check_type = False

    def assert_record_arrays_equal(a, b, check_type=False):
        if check_type:
            assert type(a) is type(b)
        assert ak.fields(a) == ak.fields(b)
>       assert all([ak.all(a[f] == b[f]) for f in ak.fields(a)])
E       assert False
E        +  where False = all([True, False])

tests/test_nanoevents_vector.py:15: AssertionError
_ test_inherited_method_transpose[PolarTwoVector-ThreeVector-PtEtaPhiMLorentzVector] _

lcoord = 'PtEtaPhiMLorentzVector', threecoord = 'ThreeVector'
twocoord = 'PolarTwoVector'

    @pytest.mark.parametrize(
        "lcoord", ["LorentzVector", "PtEtaPhiMLorentzVector", "PtEtaPhiELorentzVector"]
    )
    @pytest.mark.parametrize("threecoord", ["ThreeVector", "SphericalThreeVector"])
    @pytest.mark.parametrize("twocoord", ["TwoVector", "PolarTwoVector"])
    def test_inherited_method_transpose(lcoord, threecoord, twocoord):
        if lcoord == "LorentzVector":
            a = ak.zip(
                {
                    "x": [10.0, 20.0, 30.0],
                    "y": [-10.0, 20.0, 30.0],
                    "z": [5.0, 10.0, 15.0],
                    "t": [16.0, 31.0, 46.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiMLorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "mass": [1.0, 1.0, 1.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiELorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "energy": [11.0, 21.0, 31.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        if threecoord == "ThreeVector":
            b = ak.zip(
                {
                    "x": [-10.0, 20.0, -30.0],
                    "y": [-10.0, -20.0, 30.0],
                    "z": [5.0, -10.0, 15.0],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        elif threecoord == "SphericalThreeVector":
            b = ak.zip(
                {
                    "rho": [10.0, 20.0, 30.0],
                    "theta": [0.3, 0.6, 1.1],
                    "phi": [-3.0, 1.1, 0.2],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        if twocoord == "TwoVector":
            c = ak.zip(
                {"x": [-10.0, 13.0, 15.0], "y": [12.0, -4.0, 41.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
        elif twocoord == "PolarTwoVector":
            c = ak.zip(
                {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
    
        assert_record_arrays_equal(a.like(b) + b, b + a.like(b), check_type=True)
>       assert_record_arrays_equal(a.like(c) + c, c + a.like(c), check_type=True)

tests/test_nanoevents_vector.py:543: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
b = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
check_type = True

    def assert_record_arrays_equal(a, b, check_type=False):
        if check_type:
            assert type(a) is type(b)
        assert ak.fields(a) == ak.fields(b)
>       assert all([ak.all(a[f] == b[f]) for f in ak.fields(a)])
E       assert False
E        +  where False = all([True, False])

tests/test_nanoevents_vector.py:15: AssertionError
_ test_inherited_method_transpose[PolarTwoVector-ThreeVector-PtEtaPhiELorentzVector] _

lcoord = 'PtEtaPhiELorentzVector', threecoord = 'ThreeVector'
twocoord = 'PolarTwoVector'

    @pytest.mark.parametrize(
        "lcoord", ["LorentzVector", "PtEtaPhiMLorentzVector", "PtEtaPhiELorentzVector"]
    )
    @pytest.mark.parametrize("threecoord", ["ThreeVector", "SphericalThreeVector"])
    @pytest.mark.parametrize("twocoord", ["TwoVector", "PolarTwoVector"])
    def test_inherited_method_transpose(lcoord, threecoord, twocoord):
        if lcoord == "LorentzVector":
            a = ak.zip(
                {
                    "x": [10.0, 20.0, 30.0],
                    "y": [-10.0, 20.0, 30.0],
                    "z": [5.0, 10.0, 15.0],
                    "t": [16.0, 31.0, 46.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiMLorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "mass": [1.0, 1.0, 1.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiELorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "energy": [11.0, 21.0, 31.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        if threecoord == "ThreeVector":
            b = ak.zip(
                {
                    "x": [-10.0, 20.0, -30.0],
                    "y": [-10.0, -20.0, 30.0],
                    "z": [5.0, -10.0, 15.0],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        elif threecoord == "SphericalThreeVector":
            b = ak.zip(
                {
                    "rho": [10.0, 20.0, 30.0],
                    "theta": [0.3, 0.6, 1.1],
                    "phi": [-3.0, 1.1, 0.2],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        if twocoord == "TwoVector":
            c = ak.zip(
                {"x": [-10.0, 13.0, 15.0], "y": [12.0, -4.0, 41.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
        elif twocoord == "PolarTwoVector":
            c = ak.zip(
                {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
    
        assert_record_arrays_equal(a.like(b) + b, b + a.like(b), check_type=True)
>       assert_record_arrays_equal(a.like(c) + c, c + a.like(c), check_type=True)

tests/test_nanoevents_vector.py:543: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
b = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
check_type = True

    def assert_record_arrays_equal(a, b, check_type=False):
        if check_type:
            assert type(a) is type(b)
        assert ak.fields(a) == ak.fields(b)
>       assert all([ak.all(a[f] == b[f]) for f in ak.fields(a)])
E       assert False
E        +  where False = all([True, False])

tests/test_nanoevents_vector.py:15: AssertionError
_ test_inherited_method_transpose[PolarTwoVector-SphericalThreeVector-LorentzVector] _

lcoord = 'LorentzVector', threecoord = 'SphericalThreeVector'
twocoord = 'PolarTwoVector'

    @pytest.mark.parametrize(
        "lcoord", ["LorentzVector", "PtEtaPhiMLorentzVector", "PtEtaPhiELorentzVector"]
    )
    @pytest.mark.parametrize("threecoord", ["ThreeVector", "SphericalThreeVector"])
    @pytest.mark.parametrize("twocoord", ["TwoVector", "PolarTwoVector"])
    def test_inherited_method_transpose(lcoord, threecoord, twocoord):
        if lcoord == "LorentzVector":
            a = ak.zip(
                {
                    "x": [10.0, 20.0, 30.0],
                    "y": [-10.0, 20.0, 30.0],
                    "z": [5.0, 10.0, 15.0],
                    "t": [16.0, 31.0, 46.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiMLorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "mass": [1.0, 1.0, 1.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiELorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "energy": [11.0, 21.0, 31.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        if threecoord == "ThreeVector":
            b = ak.zip(
                {
                    "x": [-10.0, 20.0, -30.0],
                    "y": [-10.0, -20.0, 30.0],
                    "z": [5.0, -10.0, 15.0],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        elif threecoord == "SphericalThreeVector":
            b = ak.zip(
                {
                    "rho": [10.0, 20.0, 30.0],
                    "theta": [0.3, 0.6, 1.1],
                    "phi": [-3.0, 1.1, 0.2],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        if twocoord == "TwoVector":
            c = ak.zip(
                {"x": [-10.0, 13.0, 15.0], "y": [12.0, -4.0, 41.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
        elif twocoord == "PolarTwoVector":
            c = ak.zip(
                {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
    
        assert_record_arrays_equal(a.like(b) + b, b + a.like(b), check_type=True)
        assert_record_arrays_equal(a.like(c) + c, c + a.like(c), check_type=True)
>       assert_record_arrays_equal(b.like(c) + c, c + b.like(c), check_type=True)

tests/test_nanoevents_vector.py:544: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PolarTwoVectorArray [{rho: 17.2, phi: -2.46}, ..., {...}] type='3 * PolarT...'>
b = <PolarTwoVectorArray [{rho: 17.2, phi: -2.46}, ..., {...}] type='3 * PolarT...'>
check_type = True

    def assert_record_arrays_equal(a, b, check_type=False):
        if check_type:
            assert type(a) is type(b)
        assert ak.fields(a) == ak.fields(b)
>       assert all([ak.all(a[f] == b[f]) for f in ak.fields(a)])
E       assert False
E        +  where False = all([False, False])

tests/test_nanoevents_vector.py:15: AssertionError
_ test_inherited_method_transpose[PolarTwoVector-SphericalThreeVector-PtEtaPhiMLorentzVector] _

lcoord = 'PtEtaPhiMLorentzVector', threecoord = 'SphericalThreeVector'
twocoord = 'PolarTwoVector'

    @pytest.mark.parametrize(
        "lcoord", ["LorentzVector", "PtEtaPhiMLorentzVector", "PtEtaPhiELorentzVector"]
    )
    @pytest.mark.parametrize("threecoord", ["ThreeVector", "SphericalThreeVector"])
    @pytest.mark.parametrize("twocoord", ["TwoVector", "PolarTwoVector"])
    def test_inherited_method_transpose(lcoord, threecoord, twocoord):
        if lcoord == "LorentzVector":
            a = ak.zip(
                {
                    "x": [10.0, 20.0, 30.0],
                    "y": [-10.0, 20.0, 30.0],
                    "z": [5.0, 10.0, 15.0],
                    "t": [16.0, 31.0, 46.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiMLorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "mass": [1.0, 1.0, 1.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiELorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "energy": [11.0, 21.0, 31.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        if threecoord == "ThreeVector":
            b = ak.zip(
                {
                    "x": [-10.0, 20.0, -30.0],
                    "y": [-10.0, -20.0, 30.0],
                    "z": [5.0, -10.0, 15.0],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        elif threecoord == "SphericalThreeVector":
            b = ak.zip(
                {
                    "rho": [10.0, 20.0, 30.0],
                    "theta": [0.3, 0.6, 1.1],
                    "phi": [-3.0, 1.1, 0.2],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        if twocoord == "TwoVector":
            c = ak.zip(
                {"x": [-10.0, 13.0, 15.0], "y": [12.0, -4.0, 41.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
        elif twocoord == "PolarTwoVector":
            c = ak.zip(
                {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
    
        assert_record_arrays_equal(a.like(b) + b, b + a.like(b), check_type=True)
>       assert_record_arrays_equal(a.like(c) + c, c + a.like(c), check_type=True)

tests/test_nanoevents_vector.py:543: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
b = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
check_type = True

    def assert_record_arrays_equal(a, b, check_type=False):
        if check_type:
            assert type(a) is type(b)
        assert ak.fields(a) == ak.fields(b)
>       assert all([ak.all(a[f] == b[f]) for f in ak.fields(a)])
E       assert False
E        +  where False = all([True, False])

tests/test_nanoevents_vector.py:15: AssertionError
_ test_inherited_method_transpose[PolarTwoVector-SphericalThreeVector-PtEtaPhiELorentzVector] _

lcoord = 'PtEtaPhiELorentzVector', threecoord = 'SphericalThreeVector'
twocoord = 'PolarTwoVector'

    @pytest.mark.parametrize(
        "lcoord", ["LorentzVector", "PtEtaPhiMLorentzVector", "PtEtaPhiELorentzVector"]
    )
    @pytest.mark.parametrize("threecoord", ["ThreeVector", "SphericalThreeVector"])
    @pytest.mark.parametrize("twocoord", ["TwoVector", "PolarTwoVector"])
    def test_inherited_method_transpose(lcoord, threecoord, twocoord):
        if lcoord == "LorentzVector":
            a = ak.zip(
                {
                    "x": [10.0, 20.0, 30.0],
                    "y": [-10.0, 20.0, 30.0],
                    "z": [5.0, 10.0, 15.0],
                    "t": [16.0, 31.0, 46.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiMLorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "mass": [1.0, 1.0, 1.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        elif lcoord == "PtEtaPhiELorentzVector":
            a = ak.zip(
                {
                    "pt": [10.0, 20.0, 30.0],
                    "eta": [0.0, 1.1, 2.2],
                    "phi": [0.1, 0.9, -1.1],
                    "energy": [11.0, 21.0, 31.0],
                },
                with_name=lcoord,
                behavior=vector.behavior,
            )
        if threecoord == "ThreeVector":
            b = ak.zip(
                {
                    "x": [-10.0, 20.0, -30.0],
                    "y": [-10.0, -20.0, 30.0],
                    "z": [5.0, -10.0, 15.0],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        elif threecoord == "SphericalThreeVector":
            b = ak.zip(
                {
                    "rho": [10.0, 20.0, 30.0],
                    "theta": [0.3, 0.6, 1.1],
                    "phi": [-3.0, 1.1, 0.2],
                },
                with_name=threecoord,
                behavior=vector.behavior,
            )
        if twocoord == "TwoVector":
            c = ak.zip(
                {"x": [-10.0, 13.0, 15.0], "y": [12.0, -4.0, 41.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
        elif twocoord == "PolarTwoVector":
            c = ak.zip(
                {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
                with_name=twocoord,
                behavior=vector.behavior,
            )
    
        assert_record_arrays_equal(a.like(b) + b, b + a.like(b), check_type=True)
>       assert_record_arrays_equal(a.like(c) + c, c + a.like(c), check_type=True)

tests/test_nanoevents_vector.py:543: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
b = <PolarTwoVectorArray [{rho: 10.6, phi: -0.911}, ..., {...}] type='3 * Polar...'>
check_type = True

    def assert_record_arrays_equal(a, b, check_type=False):
        if check_type:
            assert type(a) is type(b)
        assert ak.fields(a) == ak.fields(b)
>       assert all([ak.all(a[f] == b[f]) for f in ak.fields(a)])
E       assert False
E        +  where False = all([True, False])

tests/test_nanoevents_vector.py:15: AssertionError
```

</details>

src/coffea/nanoevents/methods/delphes.py
nsmith-: I think it might make sense to raise an error if a user tries to write a code that requires implicit upcast from 3D point to 4D

Saransh-cpp: Thanks! I will open an issue in vector for this.

lgray:
@Saransh-cpp looks like a few things broke... there are missing behavior functions according to the tests.

Saransh-cpp:
Weirdly, I am not able to reproduce any of the failures locally. I guess that's why I gave this PR a green flag in the last meeting.

Saransh-cpp:
I am having no luck in reproducing these locally, even with new environments. @lgray could you or someone else check if this is reproducible locally?

The GH Actions error looks unrelated (something wrong with pytest's installation in `main`).

lgray:
I'll give it a spin soon here to see if I can reproduce.

lgray:
@Saransh-cpp so this appears to be some windows-specific issues with paths in new images github produced.

There's still another error, and we'll see if that shows up without the changes in this PR.

lgray:
@Saransh-cpp unfortunately it appears to be something with this PR. The tests with just awkward 2.6.7 over in https://github.com/CoffeaTeam/coffea/pull/1149 are passing fine.

lgray:
@Saransh-cpp any luck?

Saransh-cpp:
Hi @lgray, I tried the tests in a fresh environment as well as on another system (windows), but I really don't know why I am not seeing the errors locally. I am a bit clueless here.

lgray:
@Saransh-cpp so I reproduced this on my laptop:
```
>>> events.GenPart.parent
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.12/site-packages/dask_awkward/lib/core.py", line 1578, in __getattr__
    cls_method = getattr_static(self._meta, attr)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.12/inspect.py", line 1882, in getattr_static
    raise AttributeError(attr)
AttributeError: parent

During handling of the above exception, another exception occurred:

  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.12/site-packages/dask_awkward/lib/core.py", line 1580, in __getattr__
    raise AttributeError(f"{attr} not in fields.")
AttributeError: parent not in fields.
```

```
>>> events.GenPart._meta
<PtEtaPhiMLorentzVectorArray-typetracer [...] type='## * var * GenParticle[...'>
```

So somehow we are resolving to the wrong type! It should be a GenParticleArray-typetracer.

@jpivarski

lgray:
The layout seems to be correct though:
```
<ListOffsetArray len='##'>
    <offsets><Index dtype='int64' len='##'>
        [## ... ##]
    </Index></offsets>
    <content><RecordArray is_tuple='false' len='##'>
        <parameter name='__record__'>'GenParticle'</parameter>
        <parameter name='__doc__'>'interesting gen particles '</parameter>
        <parameter name='collection_name'>'GenPart'</parameter>
```

Saransh-cpp:
Hi @lgray. I was able to reproduce this on the third system 😬 

I think I know why this is happening. Working on it right now.

lgray:
@nsmith- finally passing tests except for what looks like a spurious failure on arm linux. I've re-requested your review some time ago, but now it looks like you can properly do one. I think most of what you asked last time has been answered.

Saransh-cpp:
Two minor missing suggestions from the review, which are more of discussion topics (maybe in tomorrow's meeting) -
- https://github.com/scikit-hep/vector/issues/488
- https://github.com/scikit-hep/awkward/pull/3177#issuecomment-2238777329

tests/test_nanoevents_vector.py
```diff
@@ -519,22 +534,36 @@ def test_inherited_method_transpose(lcoord, threecoord, twocoord):
         )
     elif twocoord == "PolarTwoVector":
         c = ak.zip(
-            {"r": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
+            {"rho": [-10.0, 13.0, 15.0], "phi": [1.22, -1.0, 1.0]},
```
nsmith-, line 537: This renaming of `r` to `rho` is something we'll need to remember to include in the release notes.
Perhaps a migration guide is in order? Are there any other user-facing changes we'll have to advertise?

nsmith- (approved):
Looks good!

tests/test_nanoevents_vector.py
lgray: This the only one that isn't resolved by reading vector's documentation.

I think this one will be a fairly low hit-rate, since most people would specify a polar two vector using pT and phi, which is the same in both cases.

PR #993, opened 2024-01-14
refactor: significantly simplify dask version of LumiList

lgray:
Turns out you can `np.unique(<rectilinear awkward array>, axis=0)` and it just works!


lgray:
@nsmith- see here

PR #997, opened 2024-01-18
feat: deprecate coffea.nanoevents.methods.vector

lgray:
Since we are looking into changing nanoevents to use https://github.com/scikit-hep/vector, see #991,  and it seems very possible, we are deprecating nanoevents.methods.vector very much in advance.

### DIFF
--- a/src/coffea/nanoevents/methods/vector.py
+++ b/src/coffea/nanoevents/methods/vector.py
@@ -56,7 +56,7 @@ from coffea.util import deprecate
 _cst = pytz.timezone("US/Central")
 _depttime = _cst.localize(datetime(2024, 6, 30, 11, 59, 59))
 deprecate(
-    "coffea.nanoevents.methods.vector will be removed and replaced with scikit-hep vector. Consider using that package!",
+    "coffea.nanoevents.methods.vector will be removed and replaced with scikit-hep vector. Nanoevents schemas internal to coffea will be migrated. Otherwise please consider using that package!",
     version="2024.7.0",
     date=str(_depttime),
     category=FutureWarning,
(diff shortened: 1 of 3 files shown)


lgray:
@nsmith- any comment?

nsmith-:
When does this warning message show up in the following?
```python
import coffea
import coffea.nanoevents
f = NanoEventsFactory.from_root({"tests/samples/nano_dy.root": "Events"})
f.events()
```

lgray:
It should be showing up when it imports `NanoAODSchema` in `nanoevents/factory.py`

lgray:
It does in the tests: https://github.com/CoffeaTeam/coffea/actions/runs/7633632610/job/20796243571?pr=997#step:10:269

nsmith- (approved):
So we warn users on import of NanoEvents-related things, which makes sense as these all currently depend on coffea-vector and will need to be migrated eventually. So it's a good scope.

src/coffea/nanoevents/methods/vector.py
```diff
@@ -43,12 +43,25 @@
...
 import numpy
+import pytz
 from dask_awkward import dask_method
 
+from coffea.util import deprecate
+
+_cst = pytz.timezone("US/Central")
+_depttime = _cst.localize(datetime(2024, 6, 30, 11, 59, 59))
+deprecate(
+    "coffea.nanoevents.methods.vector will be removed and replaced with scikit-hep vector. Consider using that package!",
```
nsmith-: Maybe we should say that the coffea schemas that are based on vector will internally be migrated to use vector?

PR #999, opened 2024-01-19
chore: cleanup VCS versioning

Saransh-cpp:
Ran into a few developer issues while installing locally and decided to inspect `pyproject.toml` -

- `hatch-vcs` uses `setuptools_scm` underneath, so it is not required to list down `setuptools_scm`'s config in `pyproject.toml` (this was redundant anyways, given that the `hatch-vcs` config was given preference)
- The path of the version file is automatically picked up by `hatch-vcs`
- The version file (`_version.py`) should not be tracked by version control
- `setuptools_scm` suggests using the name `_version.py` and not `version.py`
- `version.pyi` defines the structure of the version (this is optional and can be removed, but it is always nice to have one)

I hope these things were not intentional.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,6 @@ venv.bak/
 .idea
 .servicex
 dask-worker-space/
+
+# VCS versioning
+src/coffea/_version.py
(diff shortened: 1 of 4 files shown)


lgray:
Not intentional, thanks for cleaning it up!

lgray:
@Saransh-cpp if you're done here I'll turn on auto-merge. Please let me know either way.

Saransh-cpp:
This should be ready to merge!

PR #1000, opened 2024-01-19
fix: update Reading data with coffea NanoEvents notebook

Saransh-cpp:
Making minor fixes as I learn coffea and dask🙂 

I'm not a big fan of the new jupyter display of the awkward arrays, plus I'm not sure if I should be calling `compute()` in each cell for the calculations to run or if there is a better way to do this.

### DIFF
--- a/binder/nanoevents.ipynb
+++ b/binder/nanoevents.ipynb
@@ -36,10 +36,12 @@
     "import awkward as ak\n",
     "from coffea.nanoevents import NanoEventsFactory, NanoAODSchema\n",
     "\n",
+    "NanoAODSchema.warn_missing_crossrefs = False\n",
+    "\n",
     "fname = \"https://raw.githubusercontent.com/CoffeaTeam/coffea/master/tests/samples/nano_dy.root\"\n",
     "events = NanoEventsFactory.from_root(\n",
-    "    fname,\n",
-    "    schemaclass=NanoAODSchema.v6,\n",
+    "    {fname: \"Events\"},\n",
+    "    schemaclass=NanoAODSchema,\n",
     "    metadata={\"dataset\": \"DYJets\"},\n",
     ").events()"
    ]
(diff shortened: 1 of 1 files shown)


binder/nanoevents.ipynb
```diff
@@ -31,15 +31,24 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/saransh/Code/HEP/coffea/src/coffea/nanoevents/schemas/nanoaod.py:243: RuntimeWarning: Missing cross-reference index for FatJet_genJetAK8Idx => GenJetAK8\n",
```
Saransh-cpp: I think I should suppress this warning.

lgray: set `NanoAODSchema.warn_missing_crossrefs = False` before you make nanoevents.

lgray:
If you want to see the data in the awkward array you need to `.compute()` the array, otherwise it's all just representative of the task graph, rather than the data itself.

binder/nanoevents.ipynb
Saransh-cpp: Thanks!

lgray:
Took at look at the notebook and LGTM. Thanks!

lgray:
@Saransh-cpp anything else you want to add to this one?

binder/nanoevents.ipynb
```diff
@@ -126,7 +152,29 @@
...
+       "Partitioned, lazy, and parallel Awkward Array Dask collection.\n",
+       "\n",
+       "The class constructor is not intended for users. Instead use\n",
+       "factory functions like :py:func:`~dask_awkward.from_parquet`,\n",
+       ":py:func:`~dask_awkward.from_json`, etc.\n",
+       "\n",
+       "Within dask-awkward the ``new_array_object`` factory function is\n",
+       "used for creating new instances.\n"
+      ]
+     },
```
Saransh-cpp, line 173: Thanks for the review!

One minor question and then this should be ready to go.

This part is supposed to output information about `Generator.id1` -
```
Type:            Array
String form:     [1, -1, -1, 21, 21, 4, 2, -2, 2, 1, 3, 1, ... -1, -1, 1, -2, 2, 1, 2, -2, -1, 2, 1]
Length:          40
File:            ~/src/awkward-1.0/awkward1/highlevel.py
Docstring:       id of first parton
Class docstring: ...
```
but instead, it outputs the docs of the Dask compatible Array class. Is there a new way to display or access the information of a `TBranch`?

lgray: I don't think so since we never have access to the opened root file, just dask mapping to it, so the dask-array for the class docstring is actually correct (if confusing).

Saransh-cpp: Oh, I see, thanks! This should be ready then.

PR #1004, opened 2024-01-22
fix(docs): update Running inference tools notebook

Saransh-cpp:
I have updated the notebook and have additionally removed all the awkward/dask-awkward comparisons (if I am not wrong, the `permit_dask` argument was removed and now there is no way to turn dask "off" in coffee?)

Updated notebook: https://github.com/Saransh-cpp/coffea/blob/update-mltools-notebook/binder/mltools.ipynb


lgray:
Thanks @Saransh-cpp - having a look.

lgray:
@Saransh-cpp 

`permit_dask=True` is now `delayed=True` and is set default True. There are still cases where people want to play with concrete arrays through nanoevents, but we do not intend for it to be the common case.

Another thing you can remove in the notebook are lines like:
```python3
ak = self.get_awkward_lib(return_array, events)
```

This is no longer needed since now you can call `ak.some_function` on a `dask_awkward.Array` and it will call the corresponding `dask_awkward` function. You should be able to remove those lines and re-evaluate the notebook.

Saransh-cpp:
Thanks for the review!

> permit_dask=True is now delayed=True and is set default True. There are still cases where people want to play with concrete arrays through nanoevents, but we do not intend for it to be the common case.

Oh, I see, nice!

I've removed the occurrences of `self.get_awkward_lib` from the notebook.

lgray:
@Saransh-cpp Anything else you'd like to add here?

Saransh-cpp:
Nope, looks ready to me!

PR #1005, opened 2024-01-22
fix(docs): rewrite Coffea Processors notebook

Saransh-cpp:
I started experimenting with the `processor` notebook to replace the `Candidate` classes' behavior with Scikit-HEP/vector, but I bumped into a lot of old code.

I have rewritten the notebook with the latest coffea version. The plots look bad because I could not run the code for larger chunks on my local machine. It might be a good idea for someone else to run this notebook with larger chunks and push the code here.

On the brighter side, I was able to switch this notebook entirely to Scikit-HEP/vector locally, and everything, including the numba function, worked well. I have not pushed vector changes but I will once coffea deprecates `nanoevents.vector`.

cc: @lgray @nsmith- @jpivarski

---

I might have also discovered a bug in vector's awkward-numba backend along with its workaround (will create an issue in vector) -
```python
In [1]: import vector; import numba as nb

In [2]: vec = vector.zip({"pt": [1, 2 ,3], "phi": [1, 2, 3], "eta": [1, 2, 3], "
   ...: mass": [1, 2, 3], "charge": [1, 2, 3]})

In [3]: @nb.njit
   ...: def return_charge(vec):
   ...:     return vec.charge

In [4]: return_charge(vec)  # works fine
Out[4]: <Array [1, 2, 3] type='3 * int64'>

In [5]: @nb.njit
   ...: def return_charge(vec):
   ...:     return vec[0].charge 

In [6]: return_charge(vec)  # errors out
TypingError                               Traceback (most recent call last)
<ipython-input-8-74cedf343dae> in <cell line: 0>()
----> 1 return_charge(vec)  # errors out

/opt/homebrew/lib/python3.11/site-packages/numba/core/dispatcher.py in _compile_for_args(self, *args, **kws)
    466                 e.patch_message(msg)
    467 
--> 468             error_rewrite(e, 'typing')
    469         except errors.UnsupportedError as e:
    470             # Something unsupported is present in the user code, add help info

/opt/homebrew/lib/python3.11/site-packages/numba/core/dispatcher.py in error_rewrite(e, issue_type)
    407                 raise e
    408             else:
--> 409                 raise e.with_traceback(None)
    410 
    411         argtypes = []

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Unknown attribute 'charge' of type MomentumObject4DType(AzimuthalObjectRhoPhi(int64 x 2), LongitudinalObjectEta(int64 x 1), TemporalObjectTau(int64 x 1))

File "<ipython-input-7-37c9d458f596>", line 3:
def return_charge(vec):
    return vec[0].charge
    ^

During: typing of get attribute at <ipython-input-7-37c9d458f596> (3)

File "<ipython-input-7-37c9d458f596>", line 3:
def return_charge(vec):
    return vec[0].charge
    ^

In [7]: @nb.njit
   ...: def return_charge(vec):
   ...:     return vec.charge[0] 

In [8]: return_charge(vec)  # works fine
Out[8]: 1
```


lgray:
@Saransh-cpp ah! you're calling `.compute()` in the processor! That works but is extremely inefficient!

Instead of:
```python3
        # ... processor code
        cut = (ak.num(muons) == 2) & (ak.sum(muons.charge, axis=1) == 0)
        # add first and second muon in every event together
        dimuon = muons[cut][:, 0] + muons[cut][:, 1]
        h_mass.fill(sign="opposite", mass=dimuon.mass.compute())

        cut = (ak.num(muons) == 2) & (ak.sum(muons.charge, axis=1) != 0)
        dimuon = muons[cut][:, 0] + muons[cut][:, 1]
        h_mass.fill(sign="same", mass=dimuon.mass.compute())

        return {
            dataset: {
                "entries": len(events.compute()),
                "mass": h_mass,
            }
        }

```
do:
```python3
        # ... processor code
        cut = (ak.num(muons) == 2) & (ak.sum(muons.charge, axis=1) == 0)
        # add first and second muon in every event together
        dimuon = muons[cut][:, 0] + muons[cut][:, 1]
        h_mass.fill(sign="opposite", mass=dimuon.mass)

        cut = (ak.num(muons) == 2) & (ak.sum(muons.charge, axis=1) != 0)
        dimuon = muons[cut][:, 0] + muons[cut][:, 1]
        h_mass.fill(sign="same", mass=dimuon.mass)

        return {
            dataset: {
                "entries": ak.count(events, axis=0),
                "mass": h_mass,
            }
        }
```
and then later:
```python3
p = MyProcessor()
out = p.process(events)
print(out)
computed, = dask.compute(out)
print(computed)
```

That's the pattern we want to teach people in the documentation (it's also probably why you're having memory problems ;-)).

Then replicate this further down in the notebook.

You'll want to have a look at `coffea.dataset_tools` and specifically the `apply_to_fileset` function (check the tests). This function more or less takes up relevant functionality of the older Runner/Executor interface.

Saransh-cpp:
Thanks for the review! Not calling `compute()` while filling the histogram gives me -
```python
NotImplementedError                       Traceback (most recent call last)
Cell In[23], line 11
      4 events = NanoEventsFactory.from_root(
      5     {filename: \"Events\"},
      6     entry_stop=10000,
      7     metadata={\"dataset\": \"DoubleMuon\"},
      8     schemaclass=BaseSchema,
      9 ).events()
     10 p = MyProcessor()
---> 11 out = p.process(events)
     12 computed, = dask.compute(out)
     13 print(out)

Cell In[22], line 42, in MyProcessor.process(self, events)
     40 # add first and second muon in every event together
     41 dimuon = muons[cut][:, 0] + muons[cut][:, 1]
---> 42 h_mass.fill(sign=\"opposite\", mass=dimuon.mass)
     44 cut = (ak.num(muons) == 2) & (ak.sum(muons.charge, axis=1) != 0)
     45 dimuon = muons[cut][:, 0] + muons[cut][:, 1]

File ~/Code/HEP/coffea/.env/lib/python3.11/site-packages/hist/basehist.py:246, in BaseHist.fill(self, weight, sample, threads, *args, **kwargs)
    240     raise TypeError(
    241         \"All axes must be accounted for in fill, you may have used a disallowed name in the axes\"
    242     )
    244 data = (data_dict[i] for i in range(len(args), self.ndim))
--> 246 return super().fill(*args, *data, weight=weight, sample=sample, threads=threads)

File ~/Code/HEP/coffea/.env/lib/python3.11/site-packages/boost_histogram/_internal/hist.py:498, in Histogram.fill(self, weight, sample, threads, *args)
    495     self._variance_known = False
    497 # Convert to NumPy arrays
--> 498 args_ars = _fill_cast(args)
    499 weight_ars = _fill_cast(weight)
    500 sample_ars = _fill_cast(sample)

File ~/Code/HEP/coffea/.env/lib/python3.11/site-packages/boost_histogram/_internal/hist.py:78, in _fill_cast(value, inner)
     75     return value  # type: ignore[return-value]
     77 if not inner and isinstance(value, (tuple, list)):
---> 78     return tuple(_fill_cast(a, inner=True) for a in value)  # type: ignore[misc]
     80 if hasattr(value, \"__iter__\") or hasattr(value, \"__array__\"):
     81     return np.asarray(value)

File ~/Code/HEP/coffea/.env/lib/python3.11/site-packages/boost_histogram/_internal/hist.py:78, in <genexpr>(.0)
     75     return value  # type: ignore[return-value]
     77 if not inner and isinstance(value, (tuple, list)):
---> 78     return tuple(_fill_cast(a, inner=True) for a in value)  # type: ignore[misc]
     80 if hasattr(value, \"__iter__\") or hasattr(value, \"__array__\"):
     81     return np.asarray(value)

File ~/Code/HEP/coffea/.env/lib/python3.11/site-packages/boost_histogram/_internal/hist.py:81, in _fill_cast(value, inner)
     78     return tuple(_fill_cast(a, inner=True) for a in value)  # type: ignore[misc]
     80 if hasattr(value, \"__iter__\") or hasattr(value, \"__array__\"):
---> 81     return np.asarray(value)
     83 return value

File ~/Code/HEP/coffea/.env/lib/python3.11/site-packages/dask_awkward/lib/core.py:1614, in Array.__array__(self, *_, **__)
   1613 def __array__(self, *_, **__):
-> 1614     raise NotImplementedError

NotImplementedError:
```

Should I be using `dask-histogram` here?

lgray:
@Saransh-cpp you can use `import hist.dask as hda` instead of `import hist`. It's a drop-in replacement. correspondingly, `hist.dask` histograms can be `dask.compute`'d and receive dask collections as input.

Saransh-cpp:
Thank you for all the suggestions, @lgray! The plots look really nice now. I could not get rid of 2 `.compute()` calls from `FancyDimuonProcessor`. One is on an argument that is being passed to the JIT function and the second one is on a variable being indexed using list comprehension.

Could you please review this once more and let me know if any changes are required? Thanks!

The rendered notebook: https://github.com/Saransh-cpp/coffea/blob/coffea-processors-nb/binder/processor.ipynb

lgray:
This statement no longer applies in the dask_awkward context, there are no empty arrays since the dask_awkward array represents the whole dataset, you can safely remove this code:
```python3
        if ak.all(ak.num(fourmuon) == 0):
            # skip processing as it is an EmptyArray
            return {
                'nMuons': h_nMuons,
                'cutflow': {dataset: cutflow},
            }
```

Ah! The first use of ArrayBuilder in this context, nice!

For this will need to wrap it in a `dask_awkward.map_partitions` call with a light function shim.
It exposes some internals to users but they are reasonable enough in this case.

It'll look something like this:
```python3
@numba.njit
def find_4lep_kernel(events_leptons, builder):
     # the function def of what's currently in find_4lep()

def find_4lep(events_leptons):
    if ak.backend(events_leptons) = "typetracer":
        # here we fake the output of find_4lep_kernel since
        # operating on length-zero data returns the wrong layout!
        ak.typetracer.length_zero_if_typetracer(events_leptons.charge) # force touching of the necessary data
        return ak.Array(ak.Array([[(0,0,0,0)]]).layout.to_typetracer(forget_length=True))
    return find_4lep_kernel(events_leptons, ak.ArrayBuilder()).snapshot()

        # in the processor code later...
        # find all candidates with helper function
        fourmuon = dak.map_partitions(find_4lep, muons)

        # ... more code...

        # you can now also get rid of the compute here
        fourmuon = [muons[fourmuon[idx]] for idx in "0123"]
        
        # ... etc ....
```

nsmith-:
Is 
```python
template = ak.Array([[(0, 0, 0, 0)]])
ak.Array(template.layout.to_typetracer(forget_length=True))
```
the recommended pattern to build a `meta` over supplying a `meta=` argument to map_partitions? Is there a reason why one would be more or less preferred than the other?
Can we expect dask-awkward+numba support in the future? (I was actually about to make an issue for it because I ran into the same problem in converting the TTGamma DAS exercise code)

lgray:
@nsmith- it supports numba automatically when you don't use the ArrayBuilder, that's the thing breaking the nice compatibility. Otherwise it functions out of the box. Essentially ArrayBuilder has no delayed equivalent and that is what is breaking conceptually here.

If you could do this without the ArrayBuilder no map_partitions is needed. 

Maybe @agoose77 / @jpivarski / @douglasdavis know something that I don't here and there's something slick to do?

lgray:
On the issue of supply meta or cook a typetracer in the function, I guess supplying the meta is a better way to do it but it looks less clean in terms of flow of the code (since you have to make the meta array by hand and give it to map_partitions).

Just having it change what it does in the function leads to cleaner code and better compartmentalization of responsibilities.

However, that's just opinion.

lgray:
@Saransh-cpp as a matter of style don't do:
```python3
out = dask.compute(to_compute)[0]
```
but rather
```python3
(out, ) = dask.compute(to_compute)
```

@nsmith- pointed this out to me a bit ago, it accomplishes same thing but makes the semantics more clear, IMO.

lgray:
@Saransh-cpp  the error in the notebook in your recent commit is caused by the omission of a line:
```python3
        # find all candidates with helper function
        fourmuon = dask_awkward.map_partitions(find_4lep, muons)

        fourmuon = ak.zip({
            "z1": ak.zip({
                "lep1": fourmuon[0],
                "lep2": fourmuon[1],
                "p4": fourmuon[0] + fourmuon[1],
            }),
            "z2": ak.zip({
                "lep1": fourmuon[2],
                "lep2": fourmuon[3],
                "p4": fourmuon[2] + fourmuon[3],
            }),
        })
```

should instead be:
```python3
        # find all candidates with helper function
        fourmuon = dask_awkward.map_partitions(find_4lep, muons)
        fourmuon = [muons[fourmuon[idx]] for idx in "0123"]
        fourmuon = ak.zip({
            "z1": ak.zip({
                "lep1": fourmuon[0],
                "lep2": fourmuon[1],
                "p4": fourmuon[0] + fourmuon[1],
            }),
            "z2": ak.zip({
                "lep1": fourmuon[2],
                "lep2": fourmuon[3],
                "p4": fourmuon[2] + fourmuon[3],
            }),
        })
```

Saransh-cpp:
Edit: oops, I deleted a line by mistake.

lgray:
@Saransh-cpp I posted the fix before you posted the problem 🤣, I think the post I just made above should set you on the right path.

Saransh-cpp:
Haha yes, I just read your comment. I did not even realize that I had somehow deleted a line. I should've really pushed my code the moment I got the error.

lgray:
Cool - no more interim computes! Perfect.

In the first cell you'll want to add:
```python3
client = Client()
```

So that you get a proper parallel computing environment for the subsequent computes.

A few comments on the data processing:
```python3
    {filename: "Events"},
    entry_stop=10000,
    metadata={"dataset": "DoubleMuon"},
    schemaclass=BaseSchema,
).events()
```
`entry_stop` doesn't do anything when nanoevents is in `delayed=True` mode. You may omit this option.
and in particular you are making one enormous chunk when you pass the file as-is.

You should instead do something like:
```python3
    {filename: "Events"},
    steps_per_file=2_000,
    metadata={"dataset": "DoubleMuon"},
    schemaclass=BaseSchema,
).events()
```
You can tune the number `2_000` until you get something that's suitably fast.

Later, when you are using `preprocess` you do not specify a `maybe_step_size` which also means that it only makes one chunk per file, affecting parallelism.

Instead of:
```python3
dataset_runnable, dataset_updated = preprocess(
    fileset,
    align_clusters=False,
    files_per_batch=100,
    skip_bad_files=True,
    save_form=False,
)
```
You should do:
```python3
dataset_runnable, dataset_updated = preprocess(
    fileset,
    align_clusters=False,
    maybe_step_size=100_000,
    files_per_batch=1,
    skip_bad_files=True,
    save_form=False,
)
```
which will make 100k event-sized steps through the data.

This will also exercise your `max_chunks(events, 1000)`, since right now your code only creates one chunk.
You might want to use a number smaller than 1000 (like 200 or 300) since each of the datasets is a few million/ten-million-ish events.

Saransh-cpp:
Thanks for the detailed explanation! Pushed the changes.

agoose77:
> @nsmith- it supports numba automatically when you don't use the ArrayBuilder, that's the thing breaking the nice compatibility. Otherwise it functions out of the box. Essentially ArrayBuilder has no delayed equivalent and that is what is breaking conceptually here.
> 
> If you could do this without the ArrayBuilder no map_partitions is needed.
> 
> Maybe @agoose77 / @jpivarski / @douglasdavis know something that I don't here and there's something slick to do?

The problem of declaring a robust schema with array builder is not restricted to dask-awkward. My general approach has been to append a false item to the builder that spans the space of possible inputs. This can then be trimmed at the end. A longer-term solution would be a layout builder, c.f. https://github.com/scikit-hep/awkward/pull/2408

PR #1006, opened 2024-01-22
fix(docs): Update dataset_discovery.ipynb

nsmith-:
The headings for the sub-sections of this notebook are the same level as the overall heading, which breaks the index in the rendered docs:
Screenshot 2024-01-22 at 3 24 19 PM

This should fix it


nsmith-:
FYI @valsdav

lgray:
@nsmith- did you build the docs locally to check?

nsmith-:
No, but I use subheadings in some of the other notebooks I've made in there.

PR #1008, opened 2024-01-23
fix!: improve user interface for preprocess

lgray:
This PR changes the names of some arguments to preprocess so it breaks that interface a little.

In particular:
`maybe_step_size` -> `step_size`
`recalculate_seen_steps` -> `recalculate_steps`

you can also now pass `uproot_options` to the eventual `uproot.open` that happens during `preprocess`.

It also allows preprocess to take lists of strings and simple dicts of file info as inputs.

@nsmith-


lgray:
@nsmith- the memory leak workaround I've put in is delicate but should work for nanoaod consistently while you're taking a look at all of it.

lgray:
Ok with this recent change the workaround is much less delicate and probably suitable for wider use.

PR #1011, opened 2024-01-25
feat: round target step multiplicity so we end up with more even steps on average

lgray:
This should be more or less the old chunking output from coffea 0.7

@nsmith- Tell me what I did wrong.


nsmith- (approved):
Other than the print this looks good. I can't say if it's exactly what we had before but also I don't think its critical unless people want to be able to reproduce runs that previously had a `maxchunks`. Also there may be an argument that this is not as important with dask.

PR #1019, opened 2024-01-31
feat: calculate analysis task graphs in parallel via dask.delayed

lgray:
This is a workaround for the large amount of time it can take to calculate all the task graphs for a large and mature analysis.


lgray:
The memory footprint in dask is particularly odd right now, it's very large. While a single task graph is very manageable in memory for a single thread.

I may switch this to be local-only and use a ProcessPoolExecutor to parallelize the processing.

nsmith-:
Based on our discussion this morning, I'm closing this as stale.
We may consider making an issue to follow up on task graph building time, but we also expect substantial improvements from work by @pfackeldey in the meantime.

PR #1020, opened 2024-02-01
fix: convert sys.exit(1) to ValueError in DataDiscoveryCLI

ikrommyd:
I was getting `NameError: name 'exit' is not defined` from these statments and not a proper exit.

### DIFF
--- a/src/coffea/dataset_tools/dataset_query.py
+++ b/src/coffea/dataset_tools/dataset_query.py
@@ -3,6 +3,7 @@ import gzip
 import json
 import os
 import random
+import sys
 from collections import defaultdict
 from typing import List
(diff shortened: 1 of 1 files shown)


lgray:
Shouldn't this raise an exception instead? `sys.exit` can't be handled by any code that calls this class.

ikrommyd:
Yeah probably. I believe @valsdav was thinking actual cli here. I'll change it to whatever y'all want

valsdav:
Hi! right, better to raise an exception there!

ikrommyd:
Then converting the print statement to a ValueError should be good for both in-script/notebook and cli use

PR #1024, opened 2024-02-02
fix: update`DataDiscoveryCLI` usage of dataset_tools.preprocess

ikrommyd:
fixes breaking changes from https://github.com/CoffeaTeam/coffea/pull/1008 in `DataDiscoveryCLI`


ikrommyd:
I don't know if you wanna do a quick patch release. The `do_preprocess()` method of the cli is broken.

PR #1030, opened 2024-02-07
fix: Revert "fix: use public uproot interfaces when extracting form"

lgray:
Reverts CoffeaTeam/coffea#1013

Unfortunately opening the file twice over xrootd is extremely slow and incurs a significant performance penalty.
So we have to use the internal interfaces to avoid that.

@jpivarski


lgray:
Red herring - dominated by xrootd fluctuations on subsequent tests. Apologies for noise.

PR #1033, opened 2024-02-11
fix(build): pin dask < 2024.2.0 until a fix in awkward is in place

lgray:
Somehow dask 2024.2.0 breaks the column optimization of dask-awkward.


lgray:
c.f. https://github.com/dask-contrib/dask-awkward/issues/468

PR #1040, opened 2024-02-15
chore: test in python 3.12

lgray:


lgray:
Dropping blosc @nsmith- https://github.com/dask/distributed/pull/6027

nsmith-:
I think this was added only for some early NanoEvents / columnservice testing. Fine to remove (I don't think it's used anywhere)

lgray:
OK if we ever return that keep in mind that blosc2 supersedes now, but it's not used anyway in distributed. It was too finicky of a dependency.

PR #1042, opened 2024-02-16
feat: `modifier` argument for `partial_weights` method

rkansal47:
Allows for a modifier when selecting a subset of weights in the `Weights` class.

### DIFF
--- a/src/coffea/analysis_tools.py
+++ b/src/coffea/analysis_tools.py
@@ -366,7 +366,7 @@ class Weights:
             return self._weight / self._modifiers[modifier.replace("Down", "Up")]
         return self._weight * self._modifiers[modifier]
 
-    def partial_weight(self, include=[], exclude=[]):
+    def partial_weight(self, include=[], exclude=[], modifier=None):
         """Partial event weight vector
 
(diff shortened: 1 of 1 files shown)


nsmith- (approved):
Looks good to me! Do we need a 0.7 backport?

lgray:
Can you give an example of usage? This seems niche or possibly confusing as written.

lgray:
Another way to put this is why isn't this covered by appropriate definition of include or exclude? 
A more concise/correct API is better than stapling new parts to it.

rkansal47:
@lgray my use case is separating the variations which should preserve the normalization and those which don't. i.e. I want to select the subset of weights which should preserve the normalization, and then normalize the variations of those weights.

rkansal47:
Is there a way to do this with the current API?

lgray:
@nsmith- since you designed this you know better, but shouldn't this be covered by something like:
```python3
weights.partial_weights(include=["norm_preserving", "modifier"], exclude=["norm_modifying"])
```
?

lgray:
Oh, I see what's going on, yeah this is fine then.

PR #1063, opened 2024-03-20
fix: add correct behaviors and projection classes for candidate classes

Saransh-cpp:


lgray:
@Saransh-cpp @nsmith- this looks like it's missing all the classes derived from candidate. How do we want to proceed?

Saransh-cpp:
I am working on adding a helper function in awkward to make this easy for users (coffea in this case). I will update this PR once that function is in awkward.

lgray:
OK - great just ping when it's ready and I'm happy to move things along.

PR #1066, opened 2024-03-21
fix:Incorrect MET correction method CorrectedMETFactory.py

### LINKED ISSUE #1064: Incorrect MET correction method
Hi @lgray 
### There are two questions about MET correction：

**1. It seems that `CorrectedMETFactory.py` has not been updated according to the official correction formula.** 

In this script, propagate JEC effect to pfMET through the `corrected_polar_met`  function:

$$
\rm E_T^{miss,corr} = E_T^{miss}+\sum_{jets}({p_{T,jet}^{JEC}-p_{T,jet}^{orig})}
$$

```
def corrected_polar_met(
    met_pt, met_phi, jet_pt, jet_phi, jet_pt_orig, positive=None, dx=None, dy=None
):
    sj, cj = numpy.sin(jet_phi), numpy.cos(jet_phi)
    x = met_pt * numpy.cos(met_phi) + awkward.sum((jet_pt - jet_pt_orig) * cj, axis=1)
    y = met_pt * numpy.sin(met_phi) + awkward.sum((jet_pt - jet_pt_orig) * sj, axis=1)
    if positive is not None and dx is not None and dy is not None:
        x = x + dx if positive else x - dx
        y = y + dy if positive else y - dy

    return awkward.zip(
        {"pt": numpy.hypot(x, y), "phi": numpy.arctan2(y, x)}, depth_limit=1
    )
```
Can you illustrate where this correction method was referenced from？

And by checking [this twiki link](https://twiki.cern.ch/twiki/bin/viewauth/CMS/METType1Type2Formulae), I found that the  formula for `MET Type-I correction`(mostly used when processing with NanoAOD) should be like:

$$
\rm E_T^{miss,corr} = E_T^{miss}-\sum_{jets}({p_{T,jet}^{L123}-p_{T,jet}^{L1})}
$$

However, it is evident that in `CorrectedMETFactory.py` , jet_pt is the  ${p_{T,\text{jet}}^{L123}}$, and jet_pt_orig is pTRaw instead of ${p_{T,\text{jet}}^{L1}}$. This will result in incorrect MET correction.

**2. The target object for `CorrectedMETFactory.py` use with is MET, not RawMET.**
According to the following code, it can be seen that this script is applicable to `MET`, not `RawMET`.
Because in `NanoAOD` samples, only MET has branchs related to `UnClusteredEnergyDeltaX/Y`, and RawMET does not have.
```
out_dict["MET_UnclusteredEnergy"] = dask_awkward.map_partitions(
            create_variants,
            MET,
            corrected_jets,
            MET[self.name_map["UnClusteredEnergyDeltaX"]],
            MET[self.name_map["UnClusteredEnergyDeltaY"]],
            label="UnclusteredEnergy_met",
)
```
But in NanoAOD samples, the branch names and what they correspond to are as follows:
- RawMET: raw (i.e., uncorrected) PF MET
- MET: Type-1 corrected PF MET
- RawPuppiMET: raw PUPPI MET
- PuppiMET: Type-1 corrected PUPPI MET

So it is unreasonable to use `MET` because it has already undergone Type-I correction, and `RawMET` should be used instead.
How do you think?

9GaoHong:


lgray:
Please fix the tests your change has cause to fail. 

https://github.com/CoffeaTeam/coffea/actions/runs/8376520077/job/22936434720?pr=1066#step:10:477

src/coffea/jetmet_tools/CorrectedMETFactory.py
```diff
@@ -4,11 +4,11 @@
 
 
 def corrected_polar_met(
-    met_pt, met_phi, jet_pt, jet_phi, jet_pt_orig, positive=None, dx=None, dy=None
+    met_pt, met_phi, jet_pt, jet_phi, positive=None, dx=None, dy=None
 ):
     sj, cj = numpy.sin(jet_phi), numpy.cos(jet_phi)
-    x = met_pt * numpy.cos(met_phi) + awkward.sum((jet_pt - jet_pt_orig) * cj, axis=1)
-    y = met_pt * numpy.sin(met_phi) + awkward.sum((jet_pt - jet_pt_orig) * sj, axis=1)
+    x = met_pt * numpy.cos(met_phi) + awkward.sum(jet_pt * cj, axis=1)
```
lgray: Shouldn't this be calculating a difference in the corrected momentum with respect to each jet? That's what it says in the equation you quote, that's what the code was doing before. Now it does not. Please justify this change.

That or create a test that compares to outputs from some other tool.

src/coffea/jetmet_tools/CorrectedMETFactory.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, name_map):
 
         self.name_map = name_map
 
-    def build(self, in_MET, in_corrected_jets):
+    def build(self, in_MET, type1_MET, in_corrected_jets):
```
lgray, line 39: Couldn't you just set the Type1 MET as the MET that is input to the build function? This seems odd.

src/coffea/jetmet_tools/CorrectedMETFactory.py
9GaoHong: Sorry, the `+` here should be the `-`, I have updated it.
I have attached the example code in the [issue](https://github.com/CoffeaTeam/coffea/issues/1064).
The Type-I correction calculation formula has been implemented in `corrected_jets`：
```
def corrected_jets(self, jets, event_rho, lazy_cache):
        jets_L123 = self.jec_factory.build(
            add_jme_variables(jets, event_rho),
            lazy_cache #events.caches[0]
        )
        jets_L1 = self.jec_factory_L1.build(
            add_jme_variables(jets, event_rho),
            lazy_cache #events.caches[0]
        )
        emFraction = jets_L123.chEmEF + jets_L123.neEmEF
        mask_jec = (jets_L123.pt > 15) & (emFraction <= 0.9)
        selected_jets_L123 = ak.mask(jets_L123, mask_jec)
        jets_jec = selected_jets_L123
        jets_jec['pt'] = selected_jets_L123.pt - jets_L1.pt
        return jets_jec
```
Then propagate the difference value calculation of jet_pt corrected by L123 and L1 to MET through the following code:
```
jets_col = self._jmeu.corrected_jets(jets_col, rho, cache)
met  = self._jmeu.corrected_met(event.RawMET,event.MET, jets_col, event.fixedGridRhoFastjetAll, event.caches[0])
```
Does this explain that my updated code is correct?

9GaoHong:
> Please fix the tests your change has cause to fail.
> 
> https://github.com/CoffeaTeam/coffea/actions/runs/8376520077/job/22936434720?pr=1066#step:10:477

src/coffea/jetmet_tools/CorrectedMETFactory.py
9GaoHong: - In order to obtain the corresponding uncertainty values of MET after correction for `JEC `and `JER`, we must use `RawMET` as input for correction.RawMET means uncorrected PF MET, while MET refers to Type-1 corrected PF MET. If using MET as input, it means that it has been corrected twice.

- But if we want to obtain the `UnclusterdEnergy` uncertainty value of MET, you must use MET as input, as in `NanoAOD`, only MET has the branch for `MetUnclusterEnUpDeltaX/Y`, while RawMET does not. Therefore, it is necessary to use MET to obtain the `UnclusteredEnergy` uncertainty value.

So I have to input both `in_MET` and `type1_MET` meanwhile, they represent RawMET and MET respectively, for example:
`met  = self._jmeu.corrected_met(event.RawMET,event.MET, jets_col, event.fixedGridRhoFastjetAll, event.caches[0])`

src/coffea/jetmet_tools/CorrectedMETFactory.py
lgray: Ah this is a very clear to implement this since it now becomes unclear what the user is getting as a physics object. They expect to get the jets corrected by the stack that they put in, not something else that is suitable for MET correction.

Instead, it would be best to create two CorrectedJetsFactories:
```python3
jets_l123 = CorrectedJetFactory(config_L123).build(events.Jet) #add rho and whatnot to events.Jet
jets_l1 = CorrectedJetFactory(config_L1).build(events.Jet)

#then set ptRaw in the JECStack to jets_l1.pt
corr_met = CorrectedMetFactory(config_l123_l1).build(type1_met, jets_l123)
```

This would require no code changes to my understanding.

src/coffea/jetmet_tools/CorrectedMETFactory.py
lgray: Perhaps a better way to put this is I think this can be solved by changing the flexibility in the configuration as opposed to changing the meaning of the physics objects that people get out from correctors.

src/coffea/jetmet_tools/CorrectedMETFactory.py
9GaoHong: I agree that set `ptRaw` to `jets_L1.pt`,it's a good alternative.
But the plus signs `+` here should be changed to `-`. Just like:
`x = met_pt * numpy.cos(met_phi) - awkward.sum((jet_pt - jet_pt_orig) * cj, axis=1)`

src/coffea/jetmet_tools/CorrectedMETFactory.py
9GaoHong: But it is obvious that the `CorrectedMETFactory` function here cannot implement the correction of `RawMET` as input. Because `RawMET` does not have the `MetUnclusterEnUpDeltaX/Y` branch, a bug will occur. How can I modify the configuration file to solve this problem? At the same time, I also need to obtain the uncertainty of `MET_UnclusterdEnergy_up/down`

lgray: This would be better solved by making a new met object with the appropriate substitutions.
```python3
raw_met = events.RawMET
met_to_correct = events.MET
met_to_correct["pt"] = raw_met.pt
met_to_correct["phi"] = raw_met.phi

# then go on to correct met_to_correct, adjusting configuration as appropriate
```

lgray: Though I'm not sure if this spoils the validity of the systematic variations.

9GaoHong: I know what you mean! Great! 
This will not spoils the validity of the systemic variations,  the uncertainty of `MET_UnclusteredEnergy` is calculated using the `corrected MET_pt` value through `corrected_polar_met`, rather than RawMET.pt. 
You are incredibly smart, thank you so much for your patient and thoughtful answers.

nsmith-:
@9GaoHong if you can take a look at my comment in the associated issue https://github.com/scikit-hep/coffea/issues/1064#issuecomment-2738375395 and confirm you agree that would be useful to move this forward.

PR #1076, opened 2024-04-12
feat: Request all arrays from uproot at once inside dask task

nsmith-:
So we can do more IO optimizations inside uproot's source (e.g. coalesced reads, more concurrent decompression)

FYI @jpivarski  while working on this I found that uproot 5.3.3 broke `test_nanoevents_delphes.py` and `test_nanoevents_treemaker.py`.
https://github.com/scikit-hep/uproot5/compare/v5.3.2...v5.3.3 doesn't suggest anything obvious

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -86,6 +86,18 @@ class _TranslatedMapping:
         return self._mapping[self._func(index)]
 
 
+class _OnlySliceableAs:
+    """A workaround for how PreloadedSourceMapping works"""
+    def __init__(self, array, expected_slice):
+        self._array = array
+        self._expected_slice = expected_slice
+
+    def __getitem__(self, s):
+        if s != self._expected_slice:
+            raise RuntimeError(f"Mismatched slice: {s} vs. {self._expected_slice}")
+        return self._array
+
+
 class _map_schema_uproot(_map_schema_base):
     def __init__(
         self, schemaclass=BaseSchema, metadata=None, behavior=None, version=None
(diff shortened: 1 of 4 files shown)


nsmith-:
In the longer term we should really refactor `nanoevents/mapping`. It is a bit convoluted mostly due to being designed to work with https://github.com/CoffeaTeam/columnservice with globally unique form keys. I think we will find a different implementation of column caching. Then, more of the form key logic can be left to uproot (as it already is implemented for the sake of uproot.dask) and we can simplify the mapping to just implement the DSL for things like counts2offsets, parents, etc.

jpivarski:
> @jpivarski while working on this I found that uproot 5.3.3 broke `test_nanoevents_delphes.py` and `test_nanoevents_treemaker.py`.
> [scikit-hep/uproot5@v5.3.2...v5.3.3](https://github.com/scikit-hep/uproot5/compare/v5.3.2...v5.3.3) doesn't suggest anything obvious

Does Delphes have any `std::bitvector` TBranches? Before, they would have been skipped; now we'd be trying to interpret them. Maybe that's going wrong?

lgray:
closes #1077

src/coffea/nanoevents/factory.py
```diff
@@ -141,17 +154,23 @@ def load_buffers(
...
             decompression_executor=decompression_executor,
             interpretation_executor=interpretation_executor,
+            how=dict,
         )
-        mapping.preload_column_source(partition_key[0], partition_key[1], tree)
+        source_arrays = {
+            k: _OnlySliceableAs(v, slice(start, stop)) for k, v in arrays.items()
+        }
+        mapping = PreloadedSourceMapping(
+            PreloadedOpener(uuidpfn), start, stop, access_log=None
```
lgray, line 170: Shouldn't this be slicing on, instead, `[0, stop-start]` since we've already sliced the array down to `start:stop` in `tree.arrays` and so any further slicing of the chunk needs to index from zero?

In `preloaded.py` the mapping shouldn't be given a full array usually.

I'm not quite sure how the tests are passing if this is the case though, but this seems off to me.

nsmith-: It appears that the preloading mapping is given the full array because in the `extract_column` function we have `return columnhandle[start:stop]` so I just wrapped it to leave the behavior the same. To me I would think we would rather not have that and refactor the preloaded mapping to always assume the arrays are ready to go in the correct shape

lgray: Oh sorry - I wasn't reading the code right. Yeah your work around is fine. Unless you want to refactor now I'll go ahead and and merge?

nsmith-: I'd save the refactor for later since I'd like to remove most of the mapping stuff anyway

lgray (approved):

PR #1077, opened 2024-04-13
build: skip uproot 5.3.3

lgray:


lgray:
@nsmith- do you want to over this in coffea or is it a legitimate problem in uproot that needs to be fixed there?

nsmith-:
I will take a closer look, but proposing to merge #1076 and then fix in separate PR

PR #1079, opened 2024-04-14
feat: use tree reduction to aggregate files in preprocessing

### LINKED ISSUE #1078: Maximum recursion depth during preprocessing
**Describe the bug**
Running preprocessing on a large number of files (750 is enough in a test case I've seen) results in errors like

```pytb
RecursionError: maximum recursion depth exceeded while calling a Python object

    ak.concatenate(
        [<Array [{file: ..., ...}] type='1 * {file: string, object_path: stri...
    )
```

**To Reproduce**
I don't have an easy local way to reproduce this without much compute / scaling needed. I found that `sys.setrecursionlimit(200)` makes this fail with 300 files already. Otherwise nothing special that I'm aware of: run `dataset_tools.preprocess` over a few hundred files.

**Expected behavior**
Successful execution without error.

**Output**
see above

```
awkward: 2.6.2
dask-awkward: 2024.3.0
uproot: 5.3.2
coffea: 2024.3.0
```

**Additional context**
As advised by @lgray, the solution is presumably applying the same tree reduction from https://github.com/dask-contrib/dask-awkward/pull/479 to https://github.com/CoffeaTeam/coffea/blob/750b96dc53ad8b80480181e67e0f898ccf2b6d4d/src/coffea/dataset_tools/preprocess.py#L317.

alexander-held:
Preprocessing used to fail with `RecursionError` when enough files were included (#1078). Following helpful advice from @lgray, this applies the exact same method as https://github.com/dask-contrib/dask-awkward/pull/479 to resolve the problem by using a tree reduction strategy.

This seems to have worked as a rather clean copy/paste with no real changes apart from the paths to all the imported objects and functions. Perhaps that is not so surprising given that the approach is generic, but it did make me slightly suspicious at first. I did not find anything wrong with it myself so far. I was also testing with `sys.setrecursionlimit(200)` to more quickly hit the error. I don't _think_ that this is an issue, but wanted to mention it to be sure.

I am not very familiar with the internals here and only have a rough understanding of everything happening. In addition, my test case is running distributed on a facility (which works since the relevant code only is needed on the head node). I can confirm that these changes do resolve the issue in my test setup, but a critical look / other tests would be very welcome.

resolves #1078

### DIFF
--- a/src/coffea/dataset_tools/preprocess.py
+++ b/src/coffea/dataset_tools/preprocess.py
@@ -314,7 +314,20 @@ def preprocess(
         dak_norm_files = dask_awkward.from_awkward(
             ak_norm_files, math.ceil(len(ak_norm_files) / files_per_batch)
         )
-        files_to_preprocess[name] = dask_awkward.map_partitions(
+
+        concat_fn = partial(
+            awkward.concatenate,
+            axis=0,
+        )
+
+        split_every = 8
+
+        files_trl_label = f"{name}"
+        files_trl_token = dask.base.tokenize(dak_norm_files, concat_fn, split_every)
+        files_trl_name = f"{files_trl_label}-{files_trl_token}"
+        files_trl_tree_node_name = f"{files_trl_label}-tree-node-{files_trl_token}"
+
+        files_part = dask_awkward.map_partitions(
             get_steps,
             dak_norm_files,
             step_size=step_size,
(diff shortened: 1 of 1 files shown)


src/coffea/dataset_tools/preprocess.py
```diff
@@ -314,7 +314,20 @@ def preprocess(
...
+            awkward.concatenate,
+            axis=0,
+        )
+
+        split_every = 8
+
+        files_trl_label = f"{name}"
+        files_trl_token = dask.base.tokenize(dak_norm_files, concat_fn, split_every)
+        files_trl_name = f"{files_trl_label}-{files_trl_token}"
+        files_trl_tree_node_name = f"{files_trl_label}-tree-node-{files_trl_token}"
```
alexander-held: These use essentially the fileset keys to identify the names, which looked fine to me in the task graph but I'm happy for these to change as desired.

alexander-held:
The errors seen in CI come from Delphes and treemaker nanoevents and seem to also happen in nightly CI, so I believe they are unrelated and I'm guessing resolved by #1077?

lgray:
Can you get an image of the task graph for preprocess before and after this change? Just for posterity?

alexander-held:
Before the change:

before_change

After the change:
after_change

This is for a fileset with key `ttbar` and 30 files. For full context, produced with optimization:
```python
files_to_preprocess[name].visualize(filename=..., optimize_graph=True)
```

lgray:
We'll merge this after #1076

Issue #1041, opened 2024-02-15
Stop supporting/testing against python 3.8 on 14 Oct 2024


lgray:
Following end of python security support for this release.

PR #1085, opened 2024-04-25
ci: drop python 3.8 (images not consistently available)

lgray:


lgray:
This is partially in the direction of #1041. We still test python 3.8 in cirrus ci for now.
We'll fully drop support when scipy drops it.

PR #1092, opened 2024-05-10
fix: add zero photon charge and fix their cutBased id

ikrommyd:
This adds a zero Photon charge like we do with mass so that operations like diphoton invariant mass `(photon1 + photon2).mass` don't run into
```
AttributeError: no field named 'charge'

    numpy.add.__call__(
        <PhotonArray-typetracer [...] type='## * Photon[seediEtaOriX: int8[...'>
        <PhotonArray-typetracer [...] type='## * Photon[seediEtaOriX: int8[...'>
    )
```

Also fixes the cutBased ID of photons since it was changed in later NanoAOD versions and it is not caught by tests.

### DIFF
--- a/src/coffea/nanoevents/methods/nanoaod.py
+++ b/src/coffea/nanoevents/methods/nanoaod.py
@@ -292,12 +292,25 @@ _set_repr_name("Tau")
 class Photon(candidate.PtEtaPhiMCandidate, base.NanoCollection, base.Systematic):
     """NanoAOD photon object"""
 
-    LOOSE = 0
-    "cutBasedBitmap bit position"
-    MEDIUM = 1
-    "cutBasedBitmap bit position"
-    TIGHT = 2
-    "cutBasedBitmap bit position"
+    @classmethod
+    def _initialize_cutbased(cls, fields):
+        """Initialize cut-based selection values based on the fields"""
+        if "cutBased" in fields:
+            cls.FAIL = 0
+            """cutBased selection minimum value"""
+            cls.LOOSE = 1
+            """cutBased selection minimum value"""
+            cls.MEDIUM = 2
+            """cutBased selection minimum value"""
+            cls.TIGHT = 3
+            """cutBased selection minimum value"""
+        else:
+            cls.LOOSE = 0
+            """cutBased bit position"""
+            cls.MEDIUM = 1
+            """cutBased bit position"""
+            cls.TIGHT = 2
+            """cutBased bit position"""
 
     @property
     def mass(self):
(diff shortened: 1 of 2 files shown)


ikrommyd:
Will add a test soon!

ikrommyd:
I think this is also worth backporting to 0.7

src/coffea/nanoevents/methods/nanoaod.py
```diff
@@ -303,19 +303,29 @@ class Photon(candidate.PtEtaPhiMCandidate, base.NanoCollection, base.Systematic)
...
         return 0.0 * self.pt
 
+    @property
+    def charge(self):
+        return 0.0 * self.pt
+
     @property
     def isLoose(self):
         """Returns a boolean array marking loose cut-based photons"""
+        if "cutBased" in self.fields:
```
nsmith-, line 316: Can you add a note that this is for v9 +

ikrommyd: Yup will do. I’m chatting a bit with E/Gamma ID if this is the correct way to access the bitmask. I’m surprised a bit because it returns zero tight photons over like a half a million event file.

ikrommyd:
@nsmith- I think it's no longer a bitmap and the documentation is just very bad.
I'm getting zero tight photons within two root files with 1150994 events and 1768508 photons. This isn't expected at all.
```python
In [53]: ak.sum(events.Photon.cutBased & (1 << 2) != 0).compute()
Out[53]: 0

In [54]: ak.sum(events.Photon.isTight).compute()
Out[54]: 0
```
This is also the distribution of the cutbased id values in those files:
image
and this is the distribution in our `nano_dy.root` sample:
image
The documentation for nanoaod v10 says:
image
And for v12 it says:
image
and in v13 they dropped the second cutbased id. I think it was a bitmap when it was documented in the form of "2^(0 : loose ....)".

ikrommyd:
I don't know if this is the best way to initialize the cutbased attributes in the Photon class. Please let me know about requested changes. The downside with my implementation is that the `LOOSE`, `MEDIUM`, `TIGHT` attributes are not available before a cutbased id method is called. That's only a problem in eager mode though as they are not accessible in dask mode anyways. Doing `events.Electron.TIGHT` fails with `TypeError: meta must be an instance of an Awkward Array, not <class 'int'>.`

Another thing I thought of is to start with
```
    FAIL = 0
    "cutBased selection minimum value"
    VETO = 1
    "cutBased selection minimum value"
    LOOSE = 2
    "cutBased selection minimum value"
    MEDIUM = 3
    "cutBased selection minimum value"
    TIGHT = 4
    "cutBased selection minimum value"
```
by default and if `cutBased` is not in fields, then we switch to bitmasking on the spot without changing those attributes. No one is gonna use that old of a NanoAOD version anymore anyways.

edit: I actually did it like that in my commit below.

PR #1106, opened 2024-06-12
fix: UpRoot's KeyInFileError Handling for Preprocessor's File Exception Parameter; Fixes #1107

### LINKED ISSUE #1107: Preprocessor file_exception Parameter Not Catching uproot's KeyInFileError
**Description**
I noticed that when using the preprocessor's file_exception parameter, uproot's KeyInFileError was not always caught. This happens because the TTree is being accessed outside of the try-except statement where the file_exception parameter is being used. So while the file may be opened correctly, the try-except statement does not capture errors regarding the reading of TTrees within the file.

**Steps to Reproduce**

1. Import UpRoot's KeyInFileError and specify it in the preprocessor's file_exception parameter.
2. Try to access a file in which a given key is not present, i.e a root file without any TTrees
3. Observe that the KeyInFileError is not caught by the try-except statement.

**Expected Behavior**
The KeyInFileError should be caught by the try-except statement when using the file_exception parameter.

**Actual Behavior**
The KeyInFileError is not caught because the TTree is being accessed outside of the try-except statement where the file_exception parameter is being used.

**Fix Implemented**
I have implemented a fix for this bug and added a test to ensure this bug cannot recur.

**Recommendation**
Allowing the preprocessor to skip over files that give user-specified exceptions can save the user a lot of headaches, especially if the user wants to ignore a small number of bad files in an otherwise large batch of working files. However, it may be worth giving some type of warning to the user in the future.

SamKelson:
Hi all,

I noticed that when using the `preprocessor`'s `file_exception` parameter,` uproot`'s `KeyInFileError` was not always caught. This happens because the `TTree` is being accessed outside of the try-except statement where the `file_exception` parameter is being used. So while the file may be opened correctly, the try-except statement does not capture errors regarding the reading of TTrees within the file.

I am unsure if this is the desired behavior or not. I have implemented a fix for the bug below and added a test to make sure this bug cannot creep up again.

Allowing the `preprocessor` to skip over files that give user-specified exceptions can save the user a lot of headaches, especially if the user wants to ignore a small number of bad files in an otherwise large batch of working files. However, it may be worth giving some type of warning to the user in the future.


SamKelson:
Fixes #1107

PR #1115, opened 2024-06-17
fix: remove upper limit on numpy 2, update tests

### LINKED ISSUE #1112: n-minus-one cuts and cutflow generators are failing tests with numpy 2.0
@ikrommyd could you please have a look at this? It fails on all architectures and there are visible differences in the output arrays if you look at the failed CI runs.

image

image

Given that these are the first tests with numpy 2, somehow the problem stems from there. So this problem could easily be deeper in awkward, or in dask-awkward.

@jpivarski FYI

lgray:
Fixes #1112


pyproject.toml
```diff
@@ -48,7 +48,7 @@ dependencies = [
   "fsspec-xrootd>=0.2.3",
   "matplotlib>=3",
   "numba>=0.58.1",
-  "numpy<2,>=1.22.0",
+  "numpy>=2",
```
henryiii: This is going to be harsh for a while, any package setting `<2` won't be able to install. Also, if you do set this, you can then set new limits for everything else based on when they added numpy 2 support, like numba 0.60+. Though maybe this is jus testing?

lgray: I found a better workaround that relaxes this.

ikrommyd:
@lgray FYI the correctionlib pin is gonna cause problems in lcg releases by the way. LCG releases can’t have correctionlib 2.4.0 yet because they want to include zfit as well.

lgray:
Then they should move their pins, old correctionlib yields incorrect answers with numpy2 and that has to be fixed.

PR #1130, opened 2024-07-11
fix: updates from review

Saransh-cpp:
XRef #991 

- a new util function to register projection classes
- do not modify the global awkward behavior
- tests for geometric dimension errors

cc: @nsmith- @lgray

### DIFF
--- a/src/coffea/nanoevents/methods/candidate.py
+++ b/src/coffea/nanoevents/methods/candidate.py
@@ -9,6 +9,7 @@ import awkward
 import numpy
 
 from coffea.nanoevents.methods import vector
+from coffea.util import register_projection_classes
 
 behavior = dict(vector.behavior)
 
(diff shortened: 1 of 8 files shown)


src/coffea/nanoevents/methods/candidate.py
```diff
@@ -70,23 +71,29 @@ class PtEtaPhiECandidate(Candidate, vector.PtEtaPhiELorentzVector):
...
+    CandidateArray,  # noqa: F821
+)
+register_projection_classes(
+    PtEtaPhiMCandidateArray,  # noqa: F821
+    vector.PolarTwoVectorArray,
+    vector.SphericalThreeVectorArray,
+    vector.LorentzVectorArray,
+    PtEtaPhiMCandidateArray,  # noqa: F821
+)
+register_projection_classes(
```
nsmith-, line 90: Ok this is my misunderstanding, I hadn't realized each case is somewhat different so the function doesn't in the end save any code duplication. Though at least for many of these the first 3 arguments are the same

nsmith-: I think in this case what you had before was more readable code, is it ok to revert? No need to make separate PR

lgray: arg, sorry, I thought convo was done, my bad

Saransh-cpp: I'll have to create a separate "revert" PR to fix this. There are some changes in this PR that should go in, like the extra tests and not modifying the global awkward behavior.

nsmith-: Feel free to revert the subset of changes directly on the other PR and push, or make a revert PR if you prefer

PR #1134, opened 2024-07-19
feat: Integration of `correction-lib` within CorrectedJets/METFactory

anpicci:
This PR is intended to track the progress on integrating `correction-lib` within the coffea JME tools.

The very first contribution is the calculation of the JECs using inputs from the `correction-lib` tool.

FYI @lgray @nsmith-


coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -98,8 +98,21 @@ def getfunction(layout, depth):
...
+    """
+    input_values = [
+        awkward.flatten(jets[name_map[inp.name]]) for inp in corr_obj.inputs
+    ]
+    return input_values
+
+
 class CorrectedJetsFactory(object):
-    def __init__(self, name_map, jec_stack):
+    def __init__(self, name_map, jec_stack, jec_names, jec_year):
```
lgray: Please do not alter the signature of this function (especially with new non-defaulted arguments). We want to maintain backwards compatibility for the time being. 

Could you try implementing this such that the code detects correction-lib inputs coming in from the JEC stack, and then follow a different code path in the case of correctionlib inputs?

anpicci: @lgray thank you for pointing out this. Would it work if `jec_stack` is a list or a dict if one wants to use `correction-lib`? IMU, the JECStack class (as well as the auxiliary ones) is not needed when using `correction-lib`, as the latter already has functionalities to get the right numbers from the formulas that are part of the JME corrections/

lgray: The JEC stack also defines the mapping from names in your data to the arguments of the functions to evaluate. Sure, you can have the `jec_stack` be something much simpler (like a list) if correctionlib does a bunch of the bookkeeping for you.

Though I think you'll need to do some arranging yourself for the JES uncertainties.

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -171,6 +191,29 @@ def build(self, jets, lazy_cache):
         out_dict = dict(in_dict)
 
         # take care of nominal JEC (no JER if available)
+        ## General setup to use correction-lib
+        clib_json = self.jec_year
+        json_path = f"/cvmfs/cms.cern.ch/rsync/cms-nanoAOD/jsonpog-integration/POG/JME/{clib_json}/jet_jerc.json.gz"
```
lgray: we do not put hard dependencies on cvmfs paths in coffea.

lgray: This should be configurable.

I'd also be careful of building new objects every time you call `uncertainties`, this can become very slow.

garvitaa: Also, the json file name would change depending on if it's AK8 or AK4. So, I would also recommend against hard-coding this and instead keeping this in the jec_stack.

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -140,7 +156,11 @@ def __init__(self, name_map, jec_stack):
 
         self.real_sig = [v for k, v in name_map.items()]
         self.name_map = name_map
-        self.jec_stack = jec_stack
+        self.jec_stack = jec_stack  ## to be removed after the porting is completed
+        self.jec_names = jec_names
+        self.jec_year = jec_year
+        ## the last element of jec_names has to be a boolean, specifying if the use wants to save the different correction levels
+        self.separated = self.jec_names.pop()
```
garvitaa: This could be a useful debug feature, but from the JME point of view, providing just the name of the JEC tag for the required step should suffice. I don't foresee a need for analyses to store the result of each step, but I'm open to hearing about any exceptions.

anpicci: @garvitaa I would suggest keeping it to provide users more flexibility, without the need for them to put the hands on the coffee scripts

anpicci:
@lgray @nsmith- @garvitaa I have pushed a commit to address your comments, as well as to extend the correction-lib implementation to JER and JEC uncertainties

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -182,17 +259,85 @@ def build(self, jets, lazy_cache):
...
+        elif self.tool == "clib":
+            corrections_list = []
+
+            for lvl in self.jec_names:
+                if "Uncertainty" in lvl:
+                    continue
+
+                cumCorr = None
+                if len(corrections_list) > 0:
+                    ones = numpy.ones_like(corrections_list[-1], dtype=numpy.float32)
```
anpicci: Try without `ones`

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -220,28 +365,80 @@ def build(self, jets, lazy_cache):
...
 
         # in jer we need to have a stash for the intermediate JEC products
-        has_jer = False
-        if self.jec_stack.jer is not None and self.jec_stack.jersf is not None:
+        if self.tool == "jecstack":
+            has_jer = False
+            if self.jec_stack.jer is not None and self.jec_stack.jersf is not None:
+                has_jer = True
+        elif self.tool == "clib":
             has_jer = True
```
anpicci: Add a sanity check if the input contains jer or not

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -220,28 +365,80 @@ def build(self, jets, lazy_cache):
...
+                }
+                out_dict[
+                    "jet_energy_resolution_scale_factor"
+                ] = self.jec_stack.jersf.getScaleFactor(
+                    **jersfargs, form=_JERSF_FORM, lazy_cache=lazy_cache
+                )
+            elif self.tool == "clib":
+                ## needed to attach to jets the JECs
+                jer_out_parms = out.layout.parameters
+                jer_out_parms["corrected"] = True
```
anpicci, line 514: Consider to add a method here:
https://github.com/CoffeaTeam/coffea/blob/master/src/coffea/nanoevents/methods/nanoaod.py#L499

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -313,12 +522,24 @@ def build(self, jets, lazy_cache):
...
+                        out_dict[jer_name_map["JetPt"]], numpy.float32
+                    ),
+                    awkward.values_astype(
+                        out_dict[jer_name_map["JetEta"]], numpy.float32
+                    ),
+                    awkward.values_astype(
+                        out_dict["jet_energy_resolution"], numpy.float32
+                    ),
+                    awkward.values_astype(
+                        out_dict["jet_resolution_rand_gauss"], numpy.float32
```
anpicci, line 655: Try to use the `correction-lib` random values

anpicci:
Following the in-person discussion with @nsmith- , I will implement his suggestions (that I left here and there as comments to my commits), as well as work to make jec_stack a data class when using `correction-lib` in place of `JECStack`

green-cabbage:
Hello, may I ask for the status of this pull request? I am looking for a method where I could ask for compounded JEC from a given list, something like [L1FastJet, L2Relative, L3Absolute, L2L3Residual], and then also get JEC and JER uncertainties in a clean manner. This pull request seems to do that, but perhaps I am wrong?

anpicci:
Hi @green-cabbage , that's correct, feel free to use it and provide any type of feedback!
This is an example of how to use this version of the jetmet-tools:
https://github.com/TopEFT/topeft/blob/master/topeft/modules/corrections.py#L1425-L1462

green-cabbage:
@anpicci Noted! On the surface, this seems to work with coffea 0.7, which is convenient for me.

green-cabbage:
Dear @anpicci and @lgray ,

Quick question. Has this code been validated with coffea 202X ? It seems like from [this code](https://github.com/TopEFT/topeft/blob/faa85f6f93d4f84f9cb29c83d36c0d750983f833/analysis/topeft_run2/analysis_processor.py#L457) that I have to add lazy_cache to get the Factory. Unfortunately, `events.caches[0]` isn't available in the newest coffea afaik. I am tempted to just copy paste the `CorrectedJetsFactory` code from the newest coffea, but I am not so sure if that's a wise choice, given there must be some reason why `CorrectedJetsFactory` has also been touched from coffea 0.7

Thank you!

anpicci:
Hi @green-cabbage , I have validated this PR only with coffea 0.7

green-cabbage:
I see. Would it be theoretically difficult to migrate this method to coffea 202X? I am willing to put in some of my own time if I could get some test cases I could validate with. @anpicci @lgray

nsmith-:
Is this ready for review now?

anpicci:
Hi @nsmith-, from my side yes
@green-cabbage at the moment, I cannot dedicate effort to the coffea calendar implementation, but feel free to work on it on top of my branch!

lgray:
@anpicci tests fail due to absence of `topcoffea`. We cannot depend on packages that depend themselves on coffea, please fix this!

coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -114,53 +150,65 @@ def __init__(self, name_map, jec_stack):
...
-            warnings.warn(
-                'Input JaggedCandidateArray must have "ptGenJet" in order to apply hybrid JER smearing method. Stochastic smearing will be applied.'
-            )
-            self.forceStochastic = True
-
-        self.real_sig = [v for k, v in name_map.items()]
-        self.name_map = name_map
-        self.jec_stack = jec_stack
-
-    def uncertainties(self):
```
lgray, line 145: why was this function removed?

Justify and alter tests or please add it back!

NJManganelli:
New JERC tutorials with reference implementation. Unfortunately not a package which we can use to factorize out code, but we may want to keep it in mind, or test against it directly...
https://gitlab.cern.ch/cms-analysis/jme/jerc-application-tutorial/-/blob/master/ApplyOnNanoAOD/applyJercAndJvm.py?ref_type=heads

ikrommyd:
Since we have https://github.com/scikit-hep/coffea/pull/1521 merged, we can close this.

PR #1136, opened 2024-07-22
fix: use the new type signature for copy_behaviors

Saransh-cpp:
Wait for the new awkward release. See https://github.com/scikit-hep/awkward/pull/3177.


lgray:
anything more we need for this?

Saransh-cpp:
Just a new awkward release; the fix is currently present in main.

TODO:
- [ ] bump awkward's minimum required version

src/coffea/nanoevents/methods/nanoaod.py
```diff
@@ -312,7 +312,7 @@ def matched_photon(self, dask_array):
 
 _set_repr_name("LowPtElectron")
 awkward.behavior.update(
```
Saransh-cpp:
```suggestion
behavior.update(
```

PR #1137, opened 2024-07-24
feat: Sync the DatasetDiscoveryCLI do_preprocess options with preprocess function

NJManganelli:
This lets the user pass through save_form and other options to the preprocess function when using the DataDiscoveryCLI tool.

### DIFF
--- a/src/coffea/dataset_tools/dataset_query.py
+++ b/src/coffea/dataset_tools/dataset_query.py
@@ -10,7 +10,7 @@ import yaml
 from dask.distributed import Client
 from rich import print
 from rich.console import Console
-from rich.prompt import Confirm, IntPrompt, Prompt
+from rich.prompt import Confirm, IntPrompt, FloatPrompt, Prompt
 from rich.table import Table
 from rich.tree import Tree
 
@@ -528,10 +528,10 @@ Some basic commands:
         step_size=None,
         align_to_clusters=None,
         scheduler_url=None,
-        recalculate_steps=False,
-        files_per_batch=1,
+        recalculate_steps=None,
+        files_per_batch=None,
         file_exceptions=(OSError,),
-        save_form=False,
+        save_form=None,
         uproot_options={},
         step_size_safety_factor=0.5,
     ):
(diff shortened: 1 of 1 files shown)


src/coffea/dataset_tools/dataset_query.py
```diff
@@ -528,6 +528,12 @@ def do_preprocess(
         step_size=None,
         align_to_clusters=None,
         scheduler_url=None,
+        recalculate_steps=False,
+        files_per_batch=1,
+        file_exceptions=(OSError,),
+        save_form=False,
+        uproot_options={},
```
ikrommyd, line 535: I think `urpoot_options=None` and then setting it to empty dict within the function is best here due to the known problems of mutable default args in python.
Other than that I don't see a problem here.

NJManganelli: In general mutable default options are bad, but as this is in a library where we don't modify it between this point and uproot.open, maybe it can stay (and this is a duplicate of coffea's default in preprocess). @lgray should that be changed to None -> {}? 
https://github.com/CoffeaTeam/coffea/blob/master/src/coffea/dataset_tools/preprocess.py#L68
Trying to unpack `None` would give an error here, so we can't permit passing through a `None` option

ikrommyd: Ah, I didn't notice what is already there in the coffea code. I always do `None` and then `if x is None: x = {}` in my code so that no linters complain.

NJManganelli: @lgray quick ping, should I make the uproot_options default mutability-safe, or leave it as-is?

lgray: `**kwargs` will be a `dict` even if no kwargs are passed.

from the python 3.11 repl:
```
>>> def fun(**kwargs):
...     print(type(kwargs))
...
>>> fun()
<class 'dict'>
```

and preprocess also declares the type as `dict` with no nullability.

lgray: hence, it should always be a dict.

lgray:
otherwise LGTM - just make sure these tests pass.

PR #1144, opened 2024-07-31
fix: Add dataset discovery tools to coffea.dataset_tools's __all__ and to docs

rpsimeon34:
The dataset discovery tools were not previously included in the `dataset_tools` sub package's `__all__`, resulting in them not appearing in the docs. This PR adds objects to `__all__` and makes some edits so those objects are documented more clearly by Sphinx.

### DIFF
--- a/src/coffea/analysis_tools.py
+++ b/src/coffea/analysis_tools.py
@@ -25,7 +25,7 @@ class WeightStatistics:
     """
     Container for statistics about the weight, including the sum of squared weights
     and number of entries.
-    
+
     Parameters
     ----------
         sumw: float
@@ -39,6 +39,7 @@ class WeightStatistics:
         n: int
             The number of entries
     """
+
     def __init__(self, sumw=0.0, sumw2=0.0, minw=numpy.inf, maxw=-numpy.inf, n=0):
         self.sumw = sumw
         self.sumw2 = sumw2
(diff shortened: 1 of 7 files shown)


lgray:
including this package in `__all__` will cause rucio to be a hard dependency of coffea, which will cause all testing to fail.

rpsimeon34:
I managed to get testing to pass by installing the `dev` and `rucio` dependency collections. For some reason I also had to manually pip install `dogpile.cache` (a dependency of rucio), but then pytest passed successfully. Are you opposed to me either
a. modifying the `dev` dependency collection to include `rucio` and `dogpile.cache`, or
b. creating a separate `test` dependency collection for testing?

lgray:
@rpsimeon34 can you use `rucio-client` as opposed to `rucio`. The latter is a much heftier/nasty dependency. It's completely fine with me if it goes in dev, so long as it works on all the versions of python we test against and all the OS's.

ikrommyd:
@lgray I don't think rucio is available on windows. If so, it can't be a strict dependency.

lgray:
We're talking about the `[dev]` install extra, not the base dependencies. However, I think this doesn't actually get around the issue that importing coffea would cause a hard dependency on `rucio` or `rucio-client`. 

I think the better fix here is to manually hint these files to sphinx, on top of the auto discovery.

rpsimeon34:
Apologies, I did mean `rucio-client` as the necessary package (I was referring to the `[rucio]` optional dependencies, and should have been clearer).

It looks to me like Windows isn't explicitly supported by `rucio-clients` (per [this issue](https://github.com/rucio/rucio/issues/6336), which I don't think has been moved on at this point). So, I agree that addressing the dataset discovery files to Sphinx is probably safer than adding to `__all__`.

PR #1151, opened 2024-08-08
feat: Add ScoutingNanoAODSchema

### LINKED ISSUE #1148: Add a Schema for Scouting NanoAOD
For the scouting data sets in CMS, there are several fields with unique names. "ScoutingJet", "ScoutingElectron", so on and so forth. If you use NanoAODSchema to open these files, these objects won't be recognized as four vectors and won't have the nice methods like DeltaR and what not. 

I suggest that there should be a Schema for these Scouting data sets. At first it would be pretty simple to just have these fields inherit from there non scouting counter-parts, but if the scouting people want to add custom functionality to these fields it would then be pretty easy. 

A proposed solution would look something like this:

```python
class ScoutingNanoAODSchema(NanoAODSchema):

    mixins = {
        **NanoAODSchema.mixins,
        "ScoutingJet": "Jet",
         # all the other fields that have nice counter-parts
    }   
```

This is something I am interested in writing myself. I was hoping for feedback on if this is the correct approach and if this dataset warrants its own schema.

jslawless:
This adds a ScoutingNanoAODSchema in src/coffea/nanoevents/schemas/nanoaod.py. Currently this only matches collections in ScoutingNANO to existing mixins in NanoAODSchema. This makes it a bit easier for coffea to work out of the box with scouting data and mc. In the future, the scouting groups could add more scouting specific functionality to this schema.

As a question for the maintainers, should I add a test class along with this new schema in this commit?


lgray:
Please add tests!

jslawless:
I added a test file "test_nanoevents_scoutingnano.py". I'm not very experienced with pytest so I basically copied what was done in "test_nanoevents_pfnano.py". I had to add a sample file tests/samples/scouting_nano.root. The file I added was ~3.5 MB since I just grabbed the smallest root file I had on hand. This didn't seem to be too far from some of the other root files in that directory but let me know if it needs to be smaller. Is there any guidelines for what should be in sample file or is just a random root file fine?

lgray:
If you can pare down the test file to ~40 events or something like that it'll keep the repository more trim.

lgray:
And if you could rebase all your commits into one the larger file won't remain in the repository history.

lgray:
@jslawless if this is ready to go please mark it as ready for review :-)

jslawless:
I changed the tests/sample to only be 157 KB instead of ~3.5 MB. I rebased the commit history so I think the old file is no longer present as well.

lgray:
Looks nice and tidy, thanks!

lgray:
@jslawless is there anything else you would like to add to this PR?

jslawless:
This was all I had planned. ScoutingNanoAOD's features aren't frozen yet, so there will be more things to add in the coming months, particularly cross references, but I will add those once the exact format is frozen.

PR #1155, opened 2024-08-09
fix: correct coffea version in docs

Saransh-cpp:
Right now the docs say 0.1 as the git history (tags) are not fetched when the repository is checked out -

image

image


lgray:
thanks!

PR #1156, opened 2024-08-09
fix: Physlite elementlink update 8/4/24

### LINKED ISSUE #975: Change `CaloCalTopoClusters` to `egammaClusters` for PHYSLITE
**Describe the bug**
There has been some evolution on ATLAS PHYSLITE data format. The branch `CaloCalTopoClusters` has changed to `egammaClusters`. This means the `caloClusters` methods of the [`physlite.py`](https://github.com/CoffeaTeam/coffea/blob/master/src/coffea/nanoevents/methods/physlite.py) need to be updated. The relevant methods can be found here: [1](https://github.com/CoffeaTeam/coffea/blob/master/src/coffea/nanoevents/methods/physlite.py#L250) and [2](https://github.com/CoffeaTeam/coffea/blob/master/src/coffea/nanoevents/methods/physlite.py#L256).

**To Reproduce**
If you try to request the `caloClusters` on an object, e.g.:
```python
events.Electrons.caloClusters
```
it will fail with error:
`FieldNotFoundError: no field 'CaloCalTopoClusters' in record with 1 fields`

**Expected behavior**
I would expect to look for the branch `egammaClusters` instead of `CaloCalTopoClusters`.

 - OS: macOS
 - Version: 14.2.1
 - Coffea version: 2023.12.0

**Additional context**
Tagging @nikoladze , @matthewfeickert

### LINKED ISSUE #1075: nanoevents/methods/physlite needs to distinguish between `dak.Array` and `dak.Array._meta`
This is following up from scikit-hep/awkward#3077.

In

https://github.com/CoffeaTeam/coffea/blob/750b96dc53ad8b80480181e67e0f898ccf2b6d4d/src/coffea/nanoevents/methods/physlite.py#L93-L122

`load_column` can have type `dask_awkward.Array` past the first if-guard if the `event_index` is not also `dask_awkward.Array`. Then some things are done with it that shouldn't be done with a `dask_awkward.Array`.

The first is `ak.backend`, a top-level Awkward function that hasn't had a dask-awkward overload (until PR dask-contrib/dask-awkward#498) because it would always return the same thing, `"typetracer"`. But useless functions for the sake of conformity are worthwhile, so I'm in favor of adding `dak.backend`.

The next is `ak.typetracer.touch_data`, which is not a high-level function—not the sort of thing that we should be overriding with dask-awkward. It's a public function so that libraries that depend on Awkward (like Coffea) can use them, but it's not intended for data analysts, and that's where dask-awkward's coverage of the Awkward interface should stop. To avoid a confusing downstream error message, I've added PR scikit-hep/awkward#3079 to make Awkward _complain_ about passing a `dak.Array` into `ak.typetracer.touch_data`, which means that the issue still has to be fixed.

The Coffea code should detect whether `load_column` is a `dak.Array`, and if so, pass `load_column._meta` into these hidden-but-public Awkward functions. Where, exactly, you want to do that depends on the logic of this function; I'm not going to suggest a change. I noticed that on line 115, you're again assuming that `load_column` is not a `dak.Array` (which doesn't have a `layout`), so someone knowledgeable about it, such as @nikoladze (the original author), should take a look.

SamKelson:
Fixes #1075
Throws an error if the user tries to add an element link without supporting a Dask version of the link method. Also updated names of objects to match the newest PHYSLITE standard.

Update uproot version to 5.3.10 because of [this bug](https://github.com/scikit-hep/uproot5/pull/1245).

Added more PHYSLITE tests and disabled some due to [uproot bug](https://github.com/scikit-hep/uproot5/issues/1267). 

Uploaded newer PHYSLITE testing file. Is it okay to delete the old testing file?


lgray:
@SamKelson it should be fine to remove the old physlite testing file if you've updated all the tests to use the new one. Please go ahead!

SamKelson:
Fixes #975 
This PR also fixes the naming issue however it would be nice for the PHYSLITE schema to read metadata from files automatically.

lgray:
@SamKelson there's nothing that stops you from opening the first file in a dataset to extract additional metadata, it only costs time. Also you have to assume that all files in the dataset are structured the same way, which is a reasonable assumption that I've seen broken by PHYSLITE... I think it's reasonable to enforce it in the future though, uniformity helps with data processing.

tests/test_nanoevents_physlite.py
```diff
@@ -26,6 +29,9 @@ def test_load_single_field_of_linked(events):
         events.Electrons.caloClusters.calE.compute()
 
 
+@pytest.mark.skip(
+    reason="temporarily disabled because of uproot issue #1267 https://github.com/scikit-hep/uproot5/issues/1267"
+)
```
nsmith-, line 34: Do we want to wait for an uproot release with the fix?

nsmith- (approved):

tests/test_nanoevents_physlite.py
lgray: @SamKelson please create a tracking issue so we reactivate this test when an uproot release is made!

matthewfeickert: @jpivarski Can we poke you for a patch release for uproot with https://github.com/scikit-hep/uproot5/pull/1269 in it?

tests/test_nanoevents_physlite.py
```diff
@@ -39,3 +45,91 @@ def test_electron_track_links(events, do_slice):
                     event.GSFTrackParticles[track_index].z0
                     == trackParticles[i][j][link_index].z0
                 )
+
+
+def mock_empty(form, behavior={}):
```
matthewfeickert, line 50: @SamKelson You don't want mutable defaults in function definitions (google "python mutable defaults" to see why). These things should be written like

```python
def mock_empty(form, behavior=None):
    behavior = behavior if behavior is not None else {}
    ...
```

tests/samples/PHYSLITE_example.root
matthewfeickert, line 1: @ekourlit as @SamKelson said that you gave him this file, can you tell us where it came from so that we know what test files are where so we can update/remove them when we add things to `scikit-hep-testdata`?

matthewfeickert (commented):
Nice PR, Sam. Just some comments for follow up.

tests/test_nanoevents_physlite.py
SamKelson: Okay, thanks! I see my mistake haha. [example for myself](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

SamKelson: I will add this to a new PR in addition to making an issue for bumping the uproot version.

tests/test_nanoevents_physlite.py
SamKelson: @lgray added a tracking issue, see issue #1164

tests/test_nanoevents_physlite.py
matthewfeickert: Yeah, everyone has done this at some point in their life (it is one of those surprising behavior of Python that the first time you hit it you do a lot of head scratching) so no worries. :) 

> I will add this to a new PR in addition to making an issue for bumping the uproot version.

SGTM.

tests/samples/PHYSLITE_example.root
ekourlit: @matthewfeickert we used to have a public PHYSLITE file (Sherpa 2.2.11 Zee) under your EOS (`eos/user/f/feickert/physlite_public_testing/`) for technical collaborations with external to ATLAS developers. I guess this is what has been used here.
Since then we have released [Open Data PHYSLITE files for research](https://opendata.cern.ch/record/80020), so I reckon these can be better used for testing purposes. We released both read and simulated data.

PR #1160, opened 2024-08-13
fix: remove lru_cache from Weights

### LINKED ISSUE #1158: Cannot add more weights to coffea.analysis_tools.Weights once self.weight() is called, but self.weightStatistics says otherwise
Hello, I am currently working on the H-> mu mu analysis, and one of our weight implementation (ie btag shape) takes current nominal weight value as input in order to evaluate its new weights, which I would like to ideally add back to the `coffea.analysis_tools.Weights` instance. However, it seems that once the nominal weights are evaluated, `coffea.analysis_tools.Weights` instance looks to be incapable of adding new weights. To give a simple reproducer:

```
import dask_awkward as dak
import awkward as ak
import numpy as np
from coffea.analysis_tools import Weights

file_dict = {"root://eos.cms.rcac.purdue.edu:1094///store/mc/RunIISummer20UL18NanoAODv9/DYJetsToLL_M-100to200_TuneCP5_13TeV-amcatnloFXFX-pythia8/NANOAODSIM/106X_upgrade2018_realistic_v16_L1v1-v1/40000/AA6F89B0-EDAA-3942-A3BB-BC3709722EB4.root": {"object_path": "Events", "steps": [[0, 10], [10, 20]], "uuid": "b16b5dea-fbcd-11ed-bae7-a2a0b8bcbeef"}}

    file_dict,
    schemaclass=NanoAODSchema,
).events()


weights = Weights(None, storeIndividual=True)
weights.add("wgt1", weight=ak.ones_like(events.genWeight))

# get weights that depends on the current weight (ie btag shape weight)
old_wgt = weights.weight()
other_wgt = old_wgt * 2
weights.add("wgt2", weight=other_wgt)
new_wgt = weights.weight()
print(np.allclose(old_wgt.compute(), new_wgt.compute())) # this should print False, but gives True, because once weights.weight() is called, its values are frozen and seemingly cannot add more weights
```
From the simple reproducer above, what I would like the code to do is for `new_wgt` to give an array of twos, but at the current state, it returns an array of ones because any new weights added after `weights.weight()` seems to be ignored. Deceptively however, when I execute `print(weights.weightStatistics.keys())`, it returns `dict_keys(['wgt1', 'wgt2'])`, suggesting that it is keeping track of wgt2 when it in fact isn't. Please let me know if I didn't make anything unclear!

Thanks in advance, 

Hyeon-Seo

lgray:
Fixes #1158 

We added an lru_cache a long time ago to keep the taskgraph from exploding. More recent dask has deterministic task graphs and so we can drop this workaround, which had some unintended side-effects.

### DIFF
--- a/src/coffea/analysis_tools.py
+++ b/src/coffea/analysis_tools.py
@@ -6,6 +6,7 @@ but have been migrated and updated to be compatible with awkward-array 1.0
 
 import warnings
 from collections import namedtuple
+from functools import lru_cache
 
 import awkward
 import dask.array
@@ -1209,6 +1210,7 @@ class PackedSelection:
         for name, selection in selections.items():
             self.add(name, selection, fill_value)
 
+    @lru_cache
     def require(self, **names):
         """Return a mask vector corresponding to specific requirements
 
--- a/src/coffea/analysis_tools.py
+++ b/src/coffea/analysis_tools.py
@@ -6,7 +6,6 @@ but have been migrated and updated to be compatible with awkward-array 1.0
 
 import warnings
 from collections import namedtuple
-from functools import lru_cache
 
 import awkward
 import dask.array
@@ -347,7 +346,6 @@ class Weights:
         elif isinstance(weight, dask_awkward.Array):
             self.__add_variation_delayed(name, weight, weightUp, weightDown, shift)
 
-    @lru_cache
     def weight(self, modifier=None):
         """Current event weight vector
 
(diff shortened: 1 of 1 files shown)


lgray:
@green-cabbage please check that this fixes your issue.

green-cabbage:
> @green-cabbage please check that this fixes your issue.

Just confirmed that the fix works on my end. Thanks!

PR #1166, opened 2024-08-19
feat: Change Jet and FatJet Vector behaviour to be compatible with other lepton classes

green-cabbage:
changing Jet and FatJet's base vector class from vector.PtEtaPhiMLorentzVector to candidate.PtEtaPhiMCandidate with introduction of charge property to match operations with lepton classes (ie Muon and Electron)

### DIFF
--- a/src/coffea/nanoevents/methods/nanoaod.py
+++ b/src/coffea/nanoevents/methods/nanoaod.py
@@ -509,7 +509,7 @@ class Jet(candidate.PtEtaPhiMCandidate, base.NanoCollection, base.Systematic):
     @property
     def charge(self):
         return 0.0 * self.pt
-        
+
     @property
     def isLoose(self):
         """Returns a boolean array marking loose jets according to jetId index"""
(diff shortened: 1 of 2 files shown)


src/coffea/nanoevents/methods/nanoaod.py
```diff
@@ -506,6 +506,10 @@ class Jet(vector.PtEtaPhiMLorentzVector, base.NanoCollection, base.Systematic):
     TIGHTLEPVETO = 2
     "jetId bit position"
 
+    @property
+    def charge(self):
+        return 0.0 * self.pt
```
lgray: can you try using `ak.zeros_like(self.pt)`, it will read less data. There are a few other places in the file that could use this optimization as well, please fix them.

ikrommyd: Does the dtype matter btw? Is there a reason it's `0.0 * self.pt` to return a float specifically. If yes, it should be set in `zeros_like` as well.

lgray: zeros_like respects the dtype of the input.

ikrommyd:
@green-cabbage please also replace the other `0.0 * self.pt` in the Delphes schema while you're at it: https://github.com/CoffeaTeam/coffea/blob/246f13446271dbc0d7d1a74f9af170099be11e86/src/coffea/nanoevents/methods/delphes.py#L197 while you're at it.

src/coffea/nanoevents/methods/nanoaod.py
green-cabbage: I have made the relevant changes to the nanoaod.py. I notice that `0.0 * self.pt` code is also present in delphs.py. Is there I way I could add delphs.py to this pull request? If not, I can always just make a new pull request

green-cabbage: ok NVM, the pull request automatically does it for you once you commit to the same branch

PR #1193, opened 2024-10-15
docs: Add some missing docstrings and include dataset discovery tools in docs

rpsimeon34:
Fill in some docstrings in source code so that the API Reference Guide has more complete documentation for classes, functions, methods, etc. This PR also fixes some reStructuredText formatting issues, and includes the Dataset Discovery Tools documentation (without making those tools part of the central `coffea` namespace, as in #1144 ).


lgray:
Whoa, thanks a lot, this looks great!

PR #1195, opened 2024-10-16
ci: use python 3.13

lgray:


lgray:
Seems a few things aren't quite there yet for even regular 3.13! Numba is missing, in addition to the torch stuff that's causing problems in the CI.

lgray:
@jpivarski FYI https://github.com/numba/numba/pull/9682

So 3.13 compatibility is merged in, but waiting on movement to make the next numba.

lgray:
@jpivarski at least with the rc version of numba 0.61 things still seem to at least pass tests, but I'm rather surprised at the long execution time in the python 3.13 case. Maybe it's just an issue of using an oversubscribed machine, hopefully not actually something with numba and recent python (though that'll get fixed quick if it is the case, I imagine).

PR #1204, opened 2024-11-07
feat: add `mapfilter` decorator

pfackeldey:
This PR adds a new decorator for embarrassingly parallel computations to be collapsed into a **single node** in the high-level dask graph. 

This is basically a curried decorator version of `dask_awkward.map_partitions`. Thus, this decorator comes with the same caveats as `dask_awkward.map_partitions` regarding the mapped function:
- The mapped function must use pure eager awkward inside (no delayed operations).
- The mapped function must return a single argument, i.e. an awkward array.
- The mapped function must be emberassingly parallel.

In addition, the decorator accepts two more arguments to ease the pain of writing `if ak.backend(array) == "typetracer": ...` conditions. The arguments are `needs` and `out_like`. 

- `needs` takes a dictionary that maps arguments of the function signature (which are awkward arrays) to an utterable of slices that should be touched respectively.
- `out_like` skips the function execution during typetracing pass and "mocks" the output array as whatever `out_like` is.

An example is given with this untraceable function (it's untraceable because of the `numpy` conversion):
```python
from coffea.processor import mapfilter

@partial(
  mapfilter,
  needs={"muons": ["pt"]},
  out_like=ak.Array([0.0]),
)
def untraceable_fun(muons):
  # a non-traceable computation for ak.typetracer
  # which needs "pt" column from muons and returns a 1-element array
  return ak.Array([np.sum(ak.to_numpy(muons.pt[0:1]))])
```

Here, `muons.pt` will be forcefully touched during typetracing and the output is mocked as `ak.Array([0.0])`. 

In general, these two arguments are not needed as typetracing should work in most cases. Only in cases where it doesn't they might come in handy.

### DIFF
--- a/src/coffea/processor/__init__.py
+++ b/src/coffea/processor/__init__.py
@@ -3,8 +3,10 @@
 
 """
 
+from .decorator import mapfilter
 from .processor import ProcessorABC
 
 __all__ = [
     "ProcessorABC",
+    "mapfilter",
 ]
--- a/src/coffea/processor/decorator.py
+++ b/src/coffea/processor/decorator.py
@@ -1,0 +1,182 @@
+from dataclasses import dataclass
+from functools import wraps
+import typing as tp
+import inspect
+
+import awkward as ak
+
+
+@dataclass
+class mapfilter:
+    """Map a callable across all partitions of any number of collections.
+    This decorator is a convenience wrapper around the `dak.map_partitions` function.
+
+    It serves the following purposes:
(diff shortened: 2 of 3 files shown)


pfackeldey:
I put this PR in draft mode for now, as I plan to add documentation with examples in addition.

pfackeldey:
I'm closing this PR as I'm moving this feature directly into [dask-awkward](https://github.com/dask-contrib/dask-awkward)

PR #1221, opened 2024-11-21
fix(docs): change path to logo for deployment

lgray:


lgray:
clearly the best way to use quotes is to not use them at all

PR #1229, opened 2024-12-12
chore: Update project URLs to use 'scikit-hep/coffea'

matthewfeickert:
* Applied with

```
git grep --name-only "CoffeaTeam/coffea" | xargs sed -i 's|CoffeaTeam/coffea|scikit-hep/coffea|g'
```

where `sed` is `gsed`.


lgray:
re-running CI and then merging! Thanks!

PR #1232, opened 2024-12-16
ci: pin dask < 2024.12.0 for taskvine tests

lgray:
Checking to see if we pass tests with this version of dask-awkward or if there are residual problems.


lgray:
@btovar It looks like there may be some growing pains with the most recent dask and daskvine!

btovar:
Mmm, can confirm it works with ndcctools 7.14.0 and  dask, dask-awkward < 2024.12.0, so you are probably right.

btovar:
That new `dask._task_spec.Task` is confusing the dag processing. Seems to be equivalent to a tuple call like `(f, arg)`?

lgray:
Yeah the old tuples are now handled using Tasks and TaskRefs.

lgray:
Here's the PR where they introduced it: https://github.com/dask/dask/pull/11568

btovar:
Got it, thanks!  Need to think about this one since we really relied on the graph being tuple-based for processing, and we have code that manually generates graphs with tuples, so we would like to preserve both.

PR #1245, opened 2025-01-15
feat: EDM4HEPSchema and Newstyle FCCSchema

prayagyadav:
@lgray @davidlange6 @gomber

Here is a clean draft for the EDM4HEPSchema (edm4hep1) and FCCSchema based on the same. I have not yet managed to add many comments and descriptions, but I plan to add them eventually.

**Workings:**

- The EDM4HEPSchema reads the `edm4hep.yaml` file from the assets directory. I felt this was necessary to add maximum functionality to the schema. Reading the specifications of all the 'components' and 'datatypes' from the yaml file helps to identify the 'members' (example, `energy` is a member of `edm4hep::ReconstructedParticle` datatype) and which members correspond to the various types of cross-branch relations in EDM4HEP: vector members, OneToOneRelations, OneToManyRelations and Links.

- The Schema fetches the comments in the `edm4hep.yaml` file and assigns them as docstrings to the relevant branches.

- The EDM4HEPSchema supports all these relations (With Links needing some manual boilerplate code from the user).

- The version of the `edm4hep.yaml` file used is [here](https://github.com/key4hep/EDM4hep/blob/c6bc8f4428aaba54c621de3ae4c4b1c791018aca/edm4hep.yaml). Please note that the way Links are represented in EDM4HEP has changed in the latest commit. @tmadlener can comment more on this. In any case, it seems necessary to find a way to track the changes from edm4hep.yaml, so that the COFFEA EDM4HEPSchema does not become obsolete after a few version changes.

**Link to example Notebooks:**
- [Here is the link to a notebook that describes the features of the EDM4HEPSchema.](https://github.com/prayagyadav/coffea/blob/master/dev/EDM4HEP-Showcase.ipynb)
- [Here is the link to a notebook that describes the features of the FCCSchema that is based on the EDM4HEPSchema.](https://github.com/prayagyadav/coffea/blob/master/dev/FCC-EDM4HEP-Showcase.ipynb)

**Tests:**
- The test sample for EDM4HEP was obtained from [here](https://github.com/key4hep/EDM4hep/blob/main/scripts/createEDM4hepFile.py) (which, I believe, was pointed out by @tmadlener).
- The test sample for the Newstyle (edm4hep1) FCC was provided by @davidlange6
- Both the schema pass the basic tests (the tests are similar to the ones that I wrote for the oldstyle (pre-edm4hep1) FCCSchema)

**Other comments:**
- I understand that this commit includes a lot of code, and its purpose may not be immediately clear, but I can provide a detailed explanation for specific snippets if needed.
- While I have tried to add most of the functionality to the schema, I cannot figure out how to add the `ExtraCode` sections mentioned in `edm4hep.yaml`. They appear to be declarations for C++ methods specific to certain collections.


tests/samples/p8_ee_WW_ecm240_edm4hep.root
lgray, line 1: is there any way this file can be made smaller? I'd prefer not to take a 12MB file into the repository.

tests/test_nanoevents_edm4hep.py
```diff
@@ -0,0 +1,87 @@
...
+    ],
+)
+def test_field_is_present(eager_events, delayed_events, field):
+    eager_fields = eager_events.fields
+    delayed_fields = delayed_events.fields
+    assert field in eager_fields
+    assert field in delayed_fields
+
+
+def test_MC_daughters(delayed_events):
```
lgray: these tests should also check correctness!

tests/test_nanoevents_fcc_edm4hep1.py
```diff
@@ -0,0 +1,138 @@
...
+    ],
+)
+def test_field_is_present(eager_events, delayed_events, field):
+    eager_fields = eager_events.fields
+    delayed_fields = delayed_events.fields
+    assert field in eager_fields
+    assert field in delayed_fields
+
+
+def test_MC_daughters(delayed_events):
```
lgray: these tests should also check correctness!

tests/test_nanoevents_fcc_edm4hep1.py
```diff
@@ -0,0 +1,138 @@
...
+    p = delayed_events.Particle.Map_Relation("parents", "Particle").compute()
+    assert isinstance(p, awkward.highlevel.Array)
+    assert p.layout.branch_depth[1] == 3
+
+
+# Todo: Add Link tests
+#
+
+
+def test_KaonParent_to_PionDaughters_Loop(eager_events):
```
lgray, line 331: ah ok, I guess correctness is checked here in the end.

Perhaps some more basic checks in the prior tests would still be useful?

tests/samples/p8_ee_WW_ecm240_edm4hep.root
prayagyadav: Sure, made smaller : 12 M --> 186 K

tests/test_nanoevents_fcc_edm4hep1.py
prayagyadav: Yeah, test_KaonParent_to_PionDaughters_Loop attempts to check the correctness of Parents and Daughters relations. Sure, I can add more basic tests

lgray:
can you rebase all the commits into 1 so that that 12MB file isn't into the repo history at all? Thanks!

prayagyadav:
> can you rebase all the commits into 1 so that that 12MB file isn't into the repo history at all? Thanks!

Done

src/coffea/nanoevents/transforms.py
```diff
@@ -282,20 +281,494 @@ def index_range(stack):
...
+    )
+    return form
+
+
+def get_index_ranges(begin, end):
+    begin_end = awkward.concatenate(
+        (begin[:, :, numpy.newaxis], end[:, :, numpy.newaxis]), axis=2
+    )
+
+    @numba.njit
```
lgray: Having this be a local function seems like a possible source of slowdown / confusion, please move this into the module namespace!

src/coffea/nanoevents/transforms.py
```diff
@@ -282,20 +281,494 @@ def index_range(stack):
...
+
+    ranges = get_index_ranges_kernel(begin_end, awkward.ArrayBuilder()).snapshot()
+
+    offset1 = ranges.layout.offsets
+    offset2 = ranges.layout.content.offsets
+    return ranges, offset1, offset2
+
+
+def get_array_from_indices(indices, target):
+    @numba.jit
```
lgray: same thing here - no reason to have this locally scoped

src/coffea/nanoevents/transforms.py
```diff
@@ -282,20 +281,494 @@ def index_range(stack):
...
+            builder.begin_list()
+            for j in range(len(indices[ev])):
+                builder.begin_list()
+                for k in indices[ev][j]:
+                    builder.real(target[ev][k])
+                builder.end_list()
+            builder.end_list()
+        return builder
+
+    @numba.jit
```
lgray: scope issue here as well.

src/coffea/nanoevents/transforms.py
```diff
@@ -282,20 +281,494 @@ def index_range(stack):
...
+        key, "!content", "!content"
+    )  # Axis 3 offsets
+    form["content"]["content"]["content"]["form_key"] = concat(
+        key, "!content", "!content", "!content"
+    )  # Content
+    return form
+
+
+def get_array_from_indices_xyzrecord_target(indices, target):
+    @numba.jit
```
lgray: another local scope + numba issue

lgray:
@prayagyadav is there anything more you want to do on this PR?

prayagyadav:
yeah, gotta sort out some features which I missed. Also, I have to add more comments ...

lgray:
OK - ping me when you are done! It's otherwise looking good so far.

prayagyadav:
Hi, @lgray. I am sorry for the radio silence. 
- There is an issue with EDM4HEPSchema that we stumbled upon: I have hard-coded the mixins, which is not desirable if I want this schema to recognize any general EDM4HEP-based root file. For example, this schema will work for the specific edm4hep sample file I have used but will fall apart for any other EDM4HEP-based root file (eg. FCC fullsim samples) unless I have the root-file specific mixin dictionary for the said root file.
- The solution that we came up with: Build the mixin dictionary at run time by inferring the C++ typenames. Infering the C++ typenames is supported in eager mode, but unsupported in delayed mode. After a brief discussion, we found that we need to add that functionality to uproot.dask.
- We have [this](https://github.com/scikit-hep/uproot5/pull/1375) PR at uproot5 to add the typename functionality. I have tested the changes and how they translate in coffea. [Here is the branch that has the necessary changes](https://github.com/prayagyadav/coffea/tree/test-extend-ak-add-doc)
- So, I believe, we have to wait for uproot to merge the change and also coffea to bump the uproot dependency

lgray:
@prayagyadav you should be able to determine the C++ just from the file metadata right? If so if you dig through uproot a bit you can extract it without needing to resort to materializing an array.

Otherwise let's wait for that PR to go through and an uproot release.

Also @nsmith- had some interesting things to say about dealing with streamers when we were talking yesterday. May help this out.

prayagyadav:
@lgray Using [this](https://github.com/HEP-FCC/AuxTools/blob/main/Podio/collInfo) script that I found, I can list out all the relevant metadata for a given EDM4HEP based file.

I found out that some generations of FCC samples have the typename info in the `podio_metadata/events___CollectionTypeInfo._1` location in the `podio_metadata` tree.

Unfortunately, the older generations like the Spring2021 campaign do not have the typename info in metadata. 

Another issue is that the collection typenames are incorrect! Maybe it's just the problem with this script, but I am not sure. 

In the long term, it would be very beneficial to get access to the Collection IDs which are stored in the metadata. Is there a way to access the `podio_metadata` tree at runtime, within a schema?

image

tmadlener:
> I found out that some generations of FCC samples have the typename info in the podio_metadata/events___CollectionTypeInfo._1 location in the podio_metadata tree.

Regarding the format of this metadata: I would like to point out that they are technically not part of the guarantees we give for stability as we do in other places, see also [the documentation](https://key4hep.web.cern.ch/podio/storage_details.html), or a concrete example of how we want to change this https://github.com/AIDASoft/podio/pull/711

If you want to have the ground truth of the generation you can get to the actual full definition of the datamodel as JSON encoded string from `podio_metadata.EDMDefinitions`.

> Another issue is that the collection typenames are incorrect! Maybe it's just the problem with this script, but I am not sure.

Maybe this is because the script gives you the names of the classes in the user-layer (see [documentation](https://key4hep.web.cern.ch/podio/design.html)), while what actually lands in the files comes from the POD layer. So if you see `vector<XYZData>` instead of `XYZCollection` that is most likely the reason for that.

davidlange6:
@tmadlener - instead the output of the script seem shuffled? I guess ```Particle``` is not really a ```UserDataCollection<float> ``` under the hood, is it?

tmadlener:
Ah right, didn't catch that, sorry. From a quick look, I think this `zip` [here](https://github.com/HEP-FCC/AuxTools/blob/f0e7842949454e34e2fd13a5688b35c09c7d130b/Podio/collInfo#L28) is the issue, because there is no guarantee that the order in the `idTable` that links collection ids and collection names is the same as the one in `CollectionTypeInfo`. However, the `CollectionTypeInfo` has a field with the collection ID, so it can be matched. The tuple is defined [here](https://github.com/AIDASoft/podio/blob/724aa5523743ff66519e786245877fe8186a85df/include/podio/utilities/RootHelpers.h#L21).

So Looking at `events___CollectionTypeInfo._0` and matching that with the collection ID retrieved from `events___idTable` should give you the correct type from `events___CollectionTypeInfo._1`.

lgray:
It seems with the latest uproot we're working rather well?

prayagyadav:
Yeah. I think I have everything in place: the improved tests, the docstrings and the comments. Seems ready for a merge.

lgray:
how large are the yaml files that you've included in `assets`?

lgray:
Ah - only seems to be 10s of kb. Not worth compressing.

src/coffea/nanoevents/assets/__init__.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+import yaml
+
+root_dir = os.path.dirname(os.path.abspath(__file__))
```
prayagyadav: aah , this section needs improvement, I guess. Maybe there is a way to load only the yaml version that I need at runtime, instead of loading every version.

prayagyadav: @lgray I would like to fix this before merge.

lgray: Sure, you can keep a dictionary of versions to functions that load the info just in time, instead of pre-loading everything like you have here. That should achieve what you want.

I am also somewhat unsure if we want to have non-python assets be tied into the module structure, may warrant a small reorganization.

lgray: something like:
```python3
versions  = [
    # as they are now
]

def _load_edm4hep_version(yamlfile):
    with open(yamlfile) as f:
         return yaml.safe_load(f)

edm4hep_ver  = {
    k: partial(
        _load_edm4hep_version,
        yamlfile="/".join([root_dir, f"edm4hep_v{k}.yaml"]))
    ) for k in versions
}
```
Then later in calling code: `edm4hep_ver[version]()`

prayagyadav: aah, thats smart. Will implement it.

lgray: You should also use `os.path.join(root_dir, f"edm4hep_v{k}.yaml")`rather than `"/".join(...)`

nsmith-: The safest way to locate the data would be with `importlib.resources.files("coffea.nanoevents.assets")` (see e.g. https://github.com/cms-nanoAOD/correctionlib/blob/master/src/correctionlib/util.py)

src/coffea/nanoevents/schemas/edm4hep.py
```diff
@@ -193,6 +193,7 @@ def version(cls, ver="latest"):
             raise ValueError(
                 "The given version {ver} is not found. Available versions are : 00.99.01,00.99.00,00.10.05,00.10.04,00.10.03,00.10.02 and 00.10.01"
```
lgray: The version list can be a `','.join(...)` on the keys of the version dict.

prayagyadav:
@lgray I have incorporated the changes suggested by you and @nsmith- . Please let me know if you find more issues in the code. Thanks

prayagyadav:
@lgray Hi, Do you have an update for the PR? The current status of the PR looks good to me. Let me know if some section of the code needs improvement. Thanks

lgray:
Hi @prayagyadav I've been taking care of issues elsewhere in the ecosystem, sorry! I'll try to review your PR today or tomorrow. A quick look seems to be in decent shape, but let me see.

lgray (approved):

lgray:
Code itself looks in good shape, is commented well as to function when it is unclear and is well tested. 

Thanks for the contribution!

PR #1250, opened 2025-01-21
ci: try out fix for typetracers in vector

lgray:
Trying https://github.com/scikit-hep/vector/pull/558 in CI


lgray:
@pfackeldey looks like that PR fixes things up! Please go ahead and make a release (and probably yank 1.6.0)

PR #1265, opened 2025-02-01
feat: Added length-0 fallback method for ML tools

yimuchen:
Certain pytorch methods fail with length-0 inputs. Length-0 inputs are potentially common for low efficiency event selection, so here we provide a way for the user to specify what to return for length-0 inputs.


ikrommyd:
Apart from the test failure which I'm sure you'll fix, I suggest that it's explicit in the docstring that `None` in the shape refers to the batch dimension.

I have one question though, why only do this for pytorch? How can we be sure that tensorflow, xgboost and triton are all fine with length-zero batches?

yimuchen:
For the other wrappers, there are alternate methods for obtaining the output dimensions rather than having the user specifying those, I'm looking into how to use those. 

Ideally, the various libraries should handle the length-0 situations themselves, so we probably just want a flag the enables fall back behavior is something fails.

yimuchen:
Very close to being done, the only caveat is that the tensorflow model testing currently requires explicit unfolding to test (so it is an issue with data mangling for length-0 chunks). Still need to think a bit how best to add a proper test case for this. I am otherwise done with the code changes to the main package.

ikrommyd:
xgboost is fine with lengh-zero input?

yimuchen:
> xgboost is fine with lengh-zero input?

Yes, I've added a test case for xgboost, and pytest runs fine with the xgboost test case.

ikrommyd:
> > xgboost is fine with lengh-zero input?
> 
> Yes, I've added a test case for xgboost, and pytest runs fine with the xgboost test case.

I guess that's good enough because I wouldn't think the structure of an xgboost model to matter in this.

yimuchen:
Test for tensorflow length-0 added, if the CI test cases pass (I did on my machine), I don't have any code to add from my side.

ikrommyd (approved):

PR #1273, opened 2025-02-11
feat: Re-enable eager parquet reading

NJManganelli:
WIP!

Eager parquet reading had a mismatch between the uproot-like shims and the parquet mappings, so `allow_missing` is inserted into the appropriate methods, with a copy of the uproot IndexedOptionArray creation added (I hope appropriately... let me know if not **Q1**). Running only the test_nanoevents.py locally the tests succeed, but I'm not sure where the dask/delayed version gets (would get) tested **Q2** (that should still fail due to lack of form_mapping in dask-awkward). 

The test for a column being available dives through the file, arrow_schema, etc. which may be more direct than desired **Q3**

I'm not expecting `allow_missing` is tested for parquet, nor if there are files already available for this... **Q4** if needed, I could probably make some... both HLT and GenModel branches potentially

The old `nano_dy.parquet` file triggers an error related to awkward extension type:
```
src/coffea/nanoevents/factory.py:557: in from_parquet
    base_form = mapping._extract_base_form(table_file.schema_arrow)
src/coffea/nanoevents/mapping/parquet.py:155: in _extract_base_form
    form = arrow_schema_to_awkward_form(schema)
src/coffea/nanoevents/mapping/parquet.py:84: in arrow_schema_to_awkward_form
    dtype = schema.to_pandas_dtype()()
pyarrow/types.pxi:406: in pyarrow.lib.DataType.to_pandas_dtype
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   NotImplementedError: ak:uint32
```
c.f. https://github.com/scikit-hep/awkward/issues/3393

However... newer parquet files made with hepconvert, even with extensionarray=True, pass. Might be able to make use of accessing the `storage_type` to bypass the error, but I didn't have a complete implementation before I swapped to using newer parquet files. **Q5**

The new parquet files use hepconvert's default compression, which is [zstd](https://github.com/scikit-hep/hepconvert/blob/main/src/hepconvert/root_to_parquet.py#L28). Okay? Better to use another? **Q6**

The equivalent of preprocess for parquet files would be a nice addition, and for reasons related to consistency (and related work on experimental column-joining, which benefits from similar interfaces + building unionform pre-emptively), I've inserted a dict-like filespec handling ala uproot5 interface. Improvements can certainly be made.

@lgray this will need quite some work still, but if you could approve/veto some of the choices made  (or give thoughts on open questions above) that would be greatly appreciated!


NJManganelli:
@lgray If this is useful to do, I could use your (hopefully really quick) input on the scattered questions in the opening message. Not time critical, as I'm prioritizing unit tests for the weighted cutflows (+ 2 other features that are equally substantial actually), then other parts of column-joining.

lgray:
Oh - sure I can take a look. I'll go through this next week!

lgray:
On first pass I'm surprised it was that easy to setup! Thanks! Looking into the rest as the day goes.

NJManganelli:
status: ruff must be appeased. Using a non-public preprocessing akin to uproot to generate mirrored information; but using parquet dataset might be able to more concisely build the dataset form over changing columns. The input normalization again mirrors the methods for uproot, but might be better off dropped.

The union form is needed for column-joining and might or might not be needed for the dask-awkward backend (mirroring the `uproot.dask` path), which now includes a NotImplementedError

So todos: better metadata handling if possible, lots of tests, drop unnecessary code

NJManganelli:
I've rebased things and I'll open a new PR when it's ready to go (probably not for a few weeks), instead of leaving this draft here

ikrommyd:
Hey @NJManganelli is this branch the rebased code https://github.com/NJManganelli/coffea/tree/datafactory_parquet?
I'd like to have a look. if not, could you please push it somewhere? I'd really love to see eager/virtual parquet reading.

NJManganelli:
That branch is the column-join codevelopment branch, it won’t go in
directly as the PRs (2) for the parquet and datafactory features. If you
need to test something, it has parquet preprocessing and fixes the issue
with schema application for eager (hopefully also virtual) arrays. I will
factorize and push PRs sometime this week, if nobody dumps a dumpster fire
in my lap

Best,
Nick

PR #1277, opened 2025-02-18
feat: add virtual arrays

pfackeldey:
marked as draft because:
- how should the API look like?
- need to wait for https://github.com/scikit-hep/awkward/pull/3364 

If you want to try it with the corresponding awkward branch:
```python

events = NanoEventsFactory.from_root({"tests/samples/nano_dy.root": "Events"}, mode="virtual").events()

events.Muon.pt
# <Array [??, ??, ??, ??, ..., ??, ??, ??, ??] type='40 * var * float32[param...'>

events.Jet.eta
# <Array [??, ??, ??, ??, ..., ??, ??, ??, ??] type='40 * var * float32[param...'>

events.run
# <Array [??, ??, ??, ??, ..., ??, ??, ??, ??] type='40 * uint32[parameters={...'>

import awkward as ak

events.Muon.pt.layout
# <ListOffsetArray len='40'>
#     <offsets><Index dtype='int64' len='41'>[## ... ##]</Index></offsets>
#     <content><NumpyArray dtype='float32' len='##'>
#         <parameter name='__doc__'>'pt'</parameter>
#         [## ... ##]
#     </NumpyArray></content>
# </ListOffsetArray>

ak.materialize(events.Muon.pt).layout
# <ListOffsetArray len='40'>
#     <offsets><Index dtype='int64' len='41'>
#         [ 0  0  0  0  0  2  3  5  5  5  5  6  6  6  6  7  7  7  7  7  7  7  7
#           7  9 11 11 13 13 15 15 16 17 17 17 18 18 18 18 18 18]
#     </Index></offsets>
#     <content><NumpyArray dtype='float32' len='18'>
#         <parameter name='__doc__'>'pt'</parameter>
#         [76.75332   20.131409  31.038704  50.641342  14.330103  16.724983
#          13.908063  46.498886  40.667812  51.43576   39.59279   38.89543
#          33.712822  17.082792  14.526995   4.3605423 10.117709  17.949919 ]
#     </NumpyArray></content>
# </ListOffsetArray>
```


pfackeldey:
I've deprecated `delayed` and added a new argument called `mode`. `mode` can be `"eager"` (loading all columns), `"virtual"` (creating a virtual array that loads on demand), or `"dask"` which creates a dask-awkward array. If `mode` is `None` (default), `delayed=True` will correspond to the `"dask"` case, and `delayed=False` to the `"virtual"` case. If `mode` is given, it will take precedence over `delayed`. I added a deprecation for `delayed` with an unknown deprecation date.

The only thing missing now is a new awkward version that contains virtual arrays. Otherwise, this PR is ready.

lgray:
Looks like we have an actual failure in `tests/test_nanoevents_pdune.py::test_listify`.

pfackeldey:
> Looks like we have an actual failure in `tests/test_nanoevents_pdune.py::test_listify`.

fixed by https://github.com/scikit-hep/awkward/pull/3434

ikrommyd:
@lgray I noticed you merged this. I would just like to point out that we're far from a release yet. We need to think more about the caches which are probably useful during opening the file, not allowing the schema run the kernels for the global indexes and attaching those to the objects instead, and also more testing in ci.

lgray:
I can revert it out if we'd rather keep working here... The last real commits were > 2 months ago so I thought this PR was relatively complete.

Likewise merging it lets us start up the processor/executor work in parallel.

Everything you're mentioning as to-dos are performance optimizations, and although important, they have lower impact on chunked workflows since you don't hit some of the nastier scaling (unless the file is pathologically basketed).

ikrommyd:
No no keep it so we have something that "can read" using virtual arrays in master. Even for chunked workflows of let's say 500_000 events, reading remotely is gonna be annoying. It's fine from your own disk (less than 10 seconds typically) but remotely might be more annoying. I'll open up an issue to keep track of the TODOs.

PR #1304, opened 2025-03-25
ci: Use astral-sh/setup-uv to setup Python

matthewfeickert:
* Use the https://github.com/astral-sh/setup-uv GitHub Action to setup Python, which allows for automatically creating and detecting the default virtual environment so that the `--system` flag is no longer needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


matthewfeickert (commented):
@henryiii is mentioning that

> setup-python is faster if the Python version you request is in the runner image (and relocatable python has some drawbacks)

so I think he might suggest that this PR isn't helpful and should get closed.

I'm not well informed on the drawbacks of `uv`'s standalone Python.

andrzejnovak:
Looks a little faster, why not. Thanks @matthewfeickert

PR #1314, opened 2025-04-24
feat: nanoevents without domain-specific mini-language

maxymnaumchyk:
The idea of the project:
1. **I/O-layer:** NanoAODschema should be constructed directly from the output of `uproot.dask`, `uproot.open`
2. Rewrite schema building by using “no-data-loading/touching” zipping operations (using parts of the code from ak.zip_no_broadcast) of the inputs coming from the **I/O layer**, instead of building awkward forms.
3. Propagate all schema specifics to the rearranged array, e.g., particle collection cross-references for NanoAOD. This will allow to entirely remove some of the features implemented through a DSL.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -740,7 +740,7 @@ class NanoEventsFactory:
             base_form["parameters"]["metadata"] = metadata
         if not callable(schemaclass):
             raise ValueError("Invalid schemaclass type")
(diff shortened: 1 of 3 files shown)


ikrommyd:
Since the ideas of this are being developed in https://github.com/maxymnaumchyk/awkward-zipper, this can be closed yeah?

maxymnaumchyk:
Yes, you are right.

PR #1321, opened 2025-05-06
fix: do not create new virtual arrays in coffea 0.7.x object systematics to avoid memory leaks

ikrommyd:
I'm a bit hesitant about removing laziness for the jetmet tools, but I'm certain that it should be at least removed from the up/down object systematics. It's creating crazy memory leaks in HiggsDNA. I don't have a real analysis example of jet/met tools though to see if removing laziness is bad. `pytest tests/test_jetmet_tools.py` runs faster on my machine (13s vs 16s) without laziness but that's not indicative of the what happens at scale at all.

@lgray what's your take here? Should we just leave it the object systematics or test with large chunks of data for the jetmet tools as well.

The jetmet tools appear to have some sophisticated caching mechanism that's shared between them but a crazy reference cycle can start taking place if people do `events["Jet"] = corrected_jets` and things like that.

### DIFF
--- a/coffea/analysis_tools.py
+++ b/coffea/analysis_tools.py
@@ -98,8 +98,8 @@ class Weights:
         self._weightStats[name] = WeightStatistics(
             weight.sum(),
             (weight**2).sum(),
-            weight.min(),
-            weight.max(),
+            weight.min(initial=numpy.inf),
+            weight.max(initial=-numpy.inf),
             weight.size,
         )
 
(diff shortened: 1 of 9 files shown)


ikrommyd:
@lgray I'm merging this as is. I had a chat with @nsmith- and it was decided that we can remove virtualness in this case because memory highly depends on what of references the `varying_function` holds and we have no control over that. We should keep virtualness in jetmet tools because memory was taken care of there by @nsmith- and there is also a unit test that checks that the jetmet tools leave no dangling references around. Also for jetmet tools a user may not want all the variations. For object systematics, just don't add one if you're not need it....

PR #1322, opened 2025-05-08
fix: use `client.submit` instead of `client.map` because it doesn't work the same way in the latest distributed

ikrommyd:
Perhaps this is a solution to https://github.com/dask/distributed/issues/9070
I'm not sure yet if this behaves the behavior is the same. Tests pass  on my machine with dask before 2025.4.0 and after though.


ikrommyd:
Not needed as it was fixed by https://github.com/dask/distributed/pull/9071. We just need to avoid 2025.4.0 and 2025.4.1.

PR #1323, opened 2025-05-10
feat: bring back iterative, futures and dask executors

ikrommyd:


lgray:
Looks good to me, thanks for pulling this together, and moving the deprecation process forward!

PR #1325, opened 2025-05-12
feat: bring back parsl executor

ikrommyd:


lgray:
@ikrommyd I think we should put in a lower bound on parsl given that we're making some backwards incompatible changes.
What's a reasonable version to skooch up to?

ikrommyd:
Oh yeah, we can put the same pin you did in https://github.com/scikit-hep/coffea/pull/1241

I assume you had checked then when channels got removed and put `"parsl>=2024.12.09"`. I'm not even sure the parsls removal is a backwards incompatible change even. Maybe previous versions work without the `LocalChannel` too.

ikrommyd:
Yup that's the release that removed channels. I pinned to that.

PR #1328, opened 2025-05-13
fix: do not error or return `None` when calling min/max over length zero chunks in weight statistics, return infinities instead

ikrommyd:


ikrommyd:
To be backported to coffea 0.7 as well

PR #1338, opened 2025-05-22
fix: Improve memory consumption in coffea.util.save

lauridsj:
This PR greatly improves the memory consumption of `coffea.util.save`. 

In our coffea-based framework, we use `coffea.util.save` to save the state of the processor as well as histograms produced to disk. The saving is scheduled to take place every 5 mins so that the processor can be resumed in the event of a crash.
We noticed that this causes a huge spike in memory consumption. After investigating, it seems that the current implementation creates two copies of the object to be saved in memory, due to the following:
- first, since the current implementation calls `dumps`, a copy as a byte-string is stored in python memory;
- second, the underlying Python pickler keeps track of the objects internally through another copy.

The first point is adressed by calling `dump` instead of `dumps` to write the pickle directly to disk. The second can in principle be addressed by using the fast mode of the python pickler (`p.fast = True`). According to the documentation (https://docs.python.org/3/library/pickle.html), this can lead to recursion errors for self-referential objects. In our framework this seems not to be an issue and the fast mode completely reduces the memory overhead; however, for completeness I have added an option to `save` in this PR to disable fast mode.

To give numbers, in a (rather heavy) test case of our framework the peak memory consumption dropped from 8.7 GB to 3.6 GB.


NJManganelli:
Is the pickle import used anywhere, if cloudpickle is what's called on directly? If not, I'd remove the import.

In general, changing to a new default should come with some guard-rails. With the known failure mode in case objects have circular references, can you use a try-catch block so that if `fast` fails for that (set of) error(s), the default mode is attempted instead?

lauridsj:
Hi @NJManganelli , thanks for the input!
The error raised for cyclic objects seems to be `ValueError`. I added a try/except block for this as you suggested. I tested with the following code:

```
import coffea.util

testobj = {"test": "foo"}
coffea.util.save(testobj, "test1.coffea")

testobj = {}
testobj["test"] = testobj
coffea.util.save(testobj, "test2.coffea")
```

The first saves in fast mode, while the second triggers the try/except and is then saved in normal mode.

Regarding the pickle import, it is used to set `protocol=pickle.HIGHEST_PROTOCOL`. This is in principle unrelated to the memory issues but according to the python documentation should give a speedup without drawbacks (since python >= 3.8 is required for coffea anyway). If you prefer, I could replace it by a numerical value instead of the import.

lgray:
FYI cloudpickle uses a `DEFAULT_PROTOCOL` when not passed a protocol:
https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L1311-L1313

and `DEFAULT_PROTOCOL` is `pickle.HIGHEST_PROTOCOL`:
https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L86

lgray:
Just to understand a few things - what's driving the 3.6GB processor memory footprint, once you remove the various copies during pickle?

NJManganelli:
When the fallback is triggered, it would be good to use a warnings.warn to communicate that something (might have) had a circular dependence, triggering the failure mode. This could be after the save command, that way if the nominal path also triggers an error, there's no misleading message

lauridsj:
Hi, indeed the pickle import was redundant, I've removed it. Thanks for catching that. I've also added a warning as you suggested.

Regarding our processor, in this case it is mostly accumulated histograms. I may have gone slightly overboard with this test case, containing ~1200 histograms, each for many categories, datasets, and systematic uncertainties. A more realistic case usually has 2-3 GB after the fix.

lgray:
The other thing I'd offer here is that with the present setup you won't end up with a compressed file at the end of this, and pickles are incredibly compressible for the most part.

It would be beneficial if we could still end up with a compressed file at the end of this. I'm also happy to be shown that the pickles just don't compress well any more.

lauridsj:
Hi @lgray, sorry for only now coming back to this (I got a notification because of your push :D).
I think the current solution still compresses the file as before by passing the lz4 stream to the pickler.

lgray:
No worries - just circling back to old PRs so that we can come to decisions.

OK - let me have a fresh look. Mostly I just want to preserve expectations across versions.

lgray:
Ah, I apparently lost the ability to read momentarily at some point in the past. Looks fine to me.

lgray:
Let's see what CI says but otherwise LGTM. @ikrommyd @nsmith- ?

ikrommyd:
I think it's good. I don't see anything harmful here.

nsmith- (approved):
LGTM

ikrommyd:
We have a problem with this one actually. Lambda functions.
```py
import cloudpickle

with open("tmp.pkl", "wb") as fout:
    output = lambda x: x
    p = cloudpickle.Pickler(fout)
    p.fast = True
    p.dump(output)
```
This gives you
```
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 1303, in dump
    return super().dump(obj)
           ~~~~~~~~~~~~^^^^^
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 1388, in reducer_override
    return self._function_reduce(obj)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 1266, in _function_reduce
    return self._dynamic_function_reduce(obj)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 1251, in _dynamic_function_reduce
    state = _function_getstate(func)
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 727, in _function_getstate
    f_globals_ref = _extract_code_globals(func.__code__)
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 315, in _extract_code_globals
    out_names = _extract_code_globals_cache.get(co)
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/weakref.py", line 452, in get
    return self.data.get(ref(key),default)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded in comparison

The above exception was the direct cause of the following exception:

  File "/Users/iason/t.py", line 7, in <module>
    p.dump(output)
    ~~~~~~^^^^^^^^
  File "/Users/iason/micromamba/envs/coffea-202x/lib/python3.13/site-packages/cloudpickle/cloudpickle.py", line 1307, in dump
    raise pickle.PicklingError(msg) from e
_pickle.PicklingError: Could not pickle object as excessively deep recursion required.
```
This isn't caught by the exception with catches only `ValueError` and secondly, because you need to hit the recursion limit to get the error, it's slow if it's a long recursion.

ikrommyd:
Also from the docs:
```
fast

    Deprecated. Enable fast mode if set to a true value. The fast mode disables the usage of memo, therefore speeding the pickling process by not generating superfluous PUT opcodes. It should not be used with self-referential objects, doing otherwise will cause [Pickler](https://docs.python.org/3/library/pickle.html#pickle.Pickler) to recurse infinitely.

    Use [pickletools.optimize()](https://docs.python.org/3/library/pickletools.html#pickletools.optimize) if you need more compact pickles.
```

ikrommyd:
So, I think we're going to revert this because fast is not recommended. We can use `pickletools.optimize()` to reduce the size but that takes a pickle string as input which means we need to call `dumps` instead of `dump` first to optimize it. @lauridsj which was the main problem for you? The first part (the pickle string copy) or the second? I assume the first part right? I would think the copying would cause the memory spike and not the lack of optimization. It looks like we can't do both at the same time if we wanna use non-deprecated and non-problematic stuff. People will want to pickle lambdas.

PR #1352, opened 2025-06-17
feat: make `column_accumulator` support awkward arrays and add accumulator tests

ikrommyd:
Adds support for `column_accumulator` to concatenate awkward arrays and not just numpy arrays.


ikrommyd:
@lgray I've also added the accumulator tests from coffea 0.7 which were missed during the revival of the executors.

PR #1359, opened 2025-06-28
feat: make `max_chunks` return the first N chunks per dataset (not per file per dataset like it is now)

ikrommyd:
This makes `max_chunks` return the first N chunks per dataset and not per file inside a dataset.

I believe this make it consistent with how it always has been in coffea 0.7 and the executors now, see https://github.com/scikit-hep/coffea/blob/52f2e95d206be06e5401e567d8bd02bbcfd9656a/src/coffea/processor/executor.py#L1306-L1317.
Therefore it's better for people that knew this behavior for many years.
If people want the first N chunks per file, then they can use `slice_chunks` with `bydataset=False`.

I also think it's more intuitive this way because if it's the first N chunks per file then `max_chunks(fileset, 3)` would typically be the whole fileset as most files are less than 3 chunks for 500K event chunks for example (typical number).


ikrommyd:
@lgray I think this makes more sense. Also I remember that `max_chunks` getting the first N chunks per file didn't make sense to @nsmith- when I brought this up in Houston.

`max_chunks` to me is to get "roughly the same number of chunks for each dataset". If it's per file, then that will lead to large discrepancies in the amount of data processed per dataset.

nsmith-:
Best option would be to deprecate `max_chunks` and make a new `max_chunks_per_dataset`

ikrommyd:
> Best option would be to deprecate `max_chunks` and make a new `max_chunks_per_dataset`

I thought about it but `maxchunks` in coffea 0.7 always did max chunks per dataset. What about raising a warning and prompting the user to use `slice_chunks(fileset, slice(N), bydataset=False)` to keep the N max chunks per file?

lgray:
slice_chunks is less idiomatic, but it's less automagical... and users like the latter a lot. hmm.

src/coffea/dataset_tools/manipulations.py
```diff
@@ -28,36 +28,68 @@ def max_chunks(fileset: FilesetSpec, maxchunks: int | None = None) -> FilesetSpe
...
+        # 3) zero-out all steps in the output
+        for fname in out[dname]["files"]:
+            out[dname]["files"][fname]["steps"] = []
+
+        # 4) repopulate in order, up to maxchunks total
+        for fname, step in kept:
+            out[dname]["files"][fname]["steps"].append(step)
+
+        # 5) drop files with no steps
+        out[dname]["files"] = {
```
lgray, line 102: Here we can reuse filter_files for an extra loop over datasets.

PR #1361, opened 2025-07-05
chore: Remove workaround in boolean_masks_to_categorical_integers function

NJManganelli:
This is a draft as it hits a still-extant limitation in awkward which prevents actually removing this TODO for now:
https://github.com/scikit-hep/awkward/pull/3312#issuecomment-3040247268

@ikrommyd tested changing the prototype array's dtype to > int8, which resolves the immediate error. Given the name and what it seems to do, it's unclear to me if there should be a try-except logic that starts with int8 and swaps when it reaches overflow, or if this needs some different approach, or if just brute-force swapping to int64 is sufficient.

If and when such an extension is added to awkward and that becomes the minimal version for coffea, this can be further developed (`irregular_mask_sequel` will need to be validated as having identical contents to the `irregular_mask`, at which point the slice logic can be deprecated and removed)

### DIFF
--- a/src/coffea/analysis_tools.py
+++ b/src/coffea/analysis_tools.py
@@ -113,20 +113,9 @@ def boolean_masks_to_categorical_integers(
         mask_inputs.insert(0, awkward.ones_like(mask_inputs[0], dtype=bool))
     if insert_commonmask_as_zeros is not None:
         mask_inputs.insert(0, insert_commonmask_as_zeros[:, None])
-    irregular_masks = []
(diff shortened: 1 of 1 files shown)


NJManganelli:
This is done, and the function is exercised pretty thoroughly (indirectly) via the nminusone and cutflow testing. Any objections to merging?

nsmith- (approved):

ikrommyd (approved):
LGTM

PR #1363, opened 2025-07-07
fix: make offsets start at zero for`ListOffsetArray` coming from uproot (fix physlite entry start problem)

### LINKED ISSUE #1280: Physlite schema can't deal with `entry_start`
```python

events = NanoEventsFactory.from_root({"/Users/iason/work/coffea_dev/coffea/tests/samples/PHYSLITE_example.root": "CollectionTree"}, mode="eager", schemaclass=PHYSLITESchema, entry_start=31).events()
```
gives
```
TypeError: size of array (90) is less than size of form (134)

    ak.from_buffers(
        {'class': 'RecordArray', 'fields': ['xTrigDecision', 'METAssoc_MET', ...
        69
        <coffea.nanoevents.mapping.uproot.UprootSourceMapping object at 0x113...
        buffer_key = partial-instance
        behavior = {'Systematic': <class 'coffea.nanoevents.methods.base.Syst...
        attrs = {'@events_factory': <coffea.nanoevents.factory.NanoEventsFact...
    )
```

ikrommyd:
Fixes https://github.com/scikit-hep/coffea/issues/1280

What's probably happening is that uproot gives some listoffsetarrays out whose offsets don't start at zero. That has to do with how tbaskets are aligned
Then you try to "zip" different such arrays in physlite at the form level
In nanoaod, you don't because you have a counter branch. You only zip contents together with one offset.
And because you zip together different arrays with different offsets together, you get this problem in `from_buffers` . The array coming from uproot is valid, it's not valid to be zipped with other arrays.
The fix now calls `the_array.layout.to_ListOffsetArray64(True)` to ALL listoffsetarrays  coming from uproot
This takes a listoffsetarray, and makes it a listoffsetarray with offsets that start at zero (while slicing the contents properly).

### DIFF
--- a/src/coffea/nanoevents/mapping/uproot.py
+++ b/src/coffea/nanoevents/mapping/uproot.py
@@ -233,11 +233,17 @@ class UprootSourceMapping(BaseSourceMapping):
             decompression_executor=self.decompression_executor,
             interpretation_executor=self.interpretation_executor,
         )
+        if isinstance(the_array.layout, awkward.contents.ListOffsetArray):
+            the_array = awkward.Array(the_array.layout.to_ListOffsetArray64(True))
 
         if allow_missing:
-            the_array = awkward.contents.IndexedOptionArray(
-                awkward.index.Index64(numpy.arange(stop - start, dtype=numpy.int64)),
-                awkward.contents.NumpyArray(the_array),
+            the_array = awkward.Array(
+                awkward.contents.IndexedOptionArray(
+                    awkward.index.Index64(
+                        numpy.arange(stop - start, dtype=numpy.int64)
+                    ),
+                    awkward.contents.NumpyArray(the_array),
+                )
             )
 
         return the_array
--- a/src/coffea/nanoevents/transforms.py
+++ b/src/coffea/nanoevents/transforms.py
@@ -36,7 +36,7 @@ def offsets(stack):
 
     Signature: array,!offsets
     """
-    stack.append(to_layout(stack.pop()).offsets)
+    stack.append(to_layout(stack.pop())._compact_offsets64(True))
 
 
 def mask(stack):
--- a/src/coffea/nanoevents/transforms.py
+++ b/src/coffea/nanoevents/transforms.py
@@ -36,7 +36,7 @@ def offsets(stack):
 
     Signature: array,!offsets
     """
-    stack.append(to_layout(stack.pop())._compact_offsets64(True))
+    stack.append(to_layout(stack.pop()).offsets)
 
 
 def mask(stack):
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,7 @@ jobs:
     - name: Test with pytest (${{ matrix.dask-client }} dask Client - run in parallel)
       if: matrix.dask-client == 'without'
       run: |
-        python -m pytest --cov-report=xml --cov=coffea --deselect=test_taskvine -m "not dask_client"
+        python -m pytest --cov-report=xml --cov=coffea --deselect=test_taskvine -m "not dask_client" -n 4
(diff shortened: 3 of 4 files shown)


src/coffea/nanoevents/transforms.py
```diff
@@ -36,7 +36,7 @@ def offsets(stack):
 
     Signature: array,!offsets
     """
-    stack.append(to_layout(stack.pop()).offsets)
+    stack.append(to_layout(stack.pop())._compact_offsets64(True))
```
lgray: Doesn't this mean, instead, that we're not preserving something correctly through slices and we should be using starts/stops in PHYSLITE rather than offsets?

ikrommyd: Maybe? I'm not 100% sure where the wrong offsets are coming from.

lgray: We should be a little careful, forcibly setting the starting offset to zero could easily lead to skewed event data.

ikrommyd: It may be better to change the schema to use listarrays instead of listoffsetarrays indeed.

src/coffea/nanoevents/mapping/uproot.py
```diff
@@ -233,11 +233,17 @@ def extract_column(
...
-                awkward.index.Index64(numpy.arange(stop - start, dtype=numpy.int64)),
-                awkward.contents.NumpyArray(the_array),
+            the_array = awkward.Array(
+                awkward.contents.IndexedOptionArray(
+                    awkward.index.Index64(
+                        numpy.arange(stop - start, dtype=numpy.int64)
+                    ),
+                    awkward.contents.NumpyArray(the_array),
+                )
             )
```
ikrommyd, line 247: keeping it consistent here and always returning a high-level array instead of a low-level content. `the_array` is always high-level as returned by uproot.

src/coffea/nanoevents/mapping/uproot.py
```diff
@@ -233,11 +233,17 @@ def extract_column(
             decompression_executor=self.decompression_executor,
             interpretation_executor=self.interpretation_executor,
         )
+        if isinstance(the_array.layout, awkward.contents.ListOffsetArray):
+            the_array = awkward.Array(the_array.layout.to_ListOffsetArray64(True))
```
ikrommyd, line 237: `the_array = awkward.to_packed(the_array)` is also possible here but it just does more stuff that are not required.

ikrommyd:
@lgray I really don't understand why python windows without dask only materializes virtual arrays during reading. It has nothing to do with this PR as I noticed when I removed the changes and kept the test in https://github.com/scikit-hep/coffea/pull/1363/commits/e71652e2f2947aa1c3584c876cdc1e291f0e6aba. I propose we skip it and fix it elsewhere because I will need a windows system to debug this. Issue open at https://github.com/scikit-hep/coffea/issues/1364

ikrommyd:
@nikoladze Could you please have a look and see that it procudes the correct result for physlite? Proper values aligned with the right event numbers.

nsmith- (approved):
Transforming the layout of the awkward array at the IO layer in a way that doesn't change the meaning seems safe to me. In some sense, the mapping DSL takes uproot output, does some transforms, and then manually implements the equivalent of `to_buffers` but probably should pack the data before exploding it into buffers since the `from_buffers` call is quite sensitive to these things.

nikoladze:
Thanks @ikrommyd! Looks good - seems like a good place to fix this. I manually checked some values against plain uproot and ROOT and it looks fine. Maybe we can also add a small regression test for some values. Created one here: https://github.com/nikoladze/coffea/commit/72b1dfad8531adef022215c732925ad9138303a5

lgray (approved):

lgray:
LGTM - merging!

PR #1366, opened 2025-07-12
fix: clarify function names in systematics

lgray:
_add_systematics_wrapper was supposed to be _ensure_systematics_wrapper, minor but confusing typo!


lgray:
@ikrommyd putting this on automerge.

PR #1369, opened 2025-07-16
fix: reading of sub-branches in PHYSLITE schema for eager and virtual mode

nikoladze:
@ikrommyd found that eager and virtual mode currently doesn't read branches with sub-branches correctly. This is because branches with sub-branches are removed from the forms received in eager mode (by `UprootSourceMapping._extract_base_form`), but not in dask mode.

I added a test for this in case of the Jet collection, parametrized for all 3 modes (also made the existing test for the electron collection parametrized) and it passes when i remove the corresponding logic from `UprootSourceMapping._extract_base_form`. So far it doesn't seem to break anything except for triggering another unreadable branch (happens in the currently failing physlite test). We should probably try to make a bit more consistent what happens in `_remove_not_interpretable` and `UprootSourceMapping._extract_base_form`.

@ikrommyd - can you add your fix for that, maybe also having a brief look at `_remove_not_interpretable`?

### DIFF
--- a/src/coffea/dataset_tools/preprocess.py
+++ b/src/coffea/dataset_tools/preprocess.py
@@ -17,7 +17,7 @@ import numpy
 import uproot
 from uproot._util import no_filter
 
-from coffea.util import _remove_not_interpretable, compress_form, decompress_form
+from coffea.util import _is_interpretable, compress_form, decompress_form
 
 
 def get_steps(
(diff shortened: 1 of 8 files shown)


ikrommyd:
Approved IRL!

nsmith- (commented):
🚀

PR #1371, opened 2025-07-18
feat: pass `iteritems_options` through executors

ikrommyd:
Allows the user to define `iteritems_options` (maybe to filter out branches) when running with executors.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -259,8 +259,8 @@ class NanoEventsFactory:
         schemaclass=NanoAODSchema,
         metadata=None,
         uproot_options={},
-        access_log=None,
         iteritems_options={},
+        access_log=None,
         use_ak_forth=True,
         mode="virtual",
         known_base_form=None,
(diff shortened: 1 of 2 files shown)


ikrommyd:
@lgray I think this is a good idea because we can't use `uproot_options` to filter out branches in eager/virtual mode. That's done with `iteritems_otions` as we iterate over the tree to extract the base form.

lgray:
Just to be nitpicky about this, but then shouldn't `uproot_options` be named something more precise? `iteritems_options` is also an option for uproot, so concerns are not well separated and it could be confusing to users.

lgray (dismissed):
The naming of `uproot_options`, given that we have `iteritems_options`, is ambiguous. We should refine this scheme.

ikrommyd:
It's been like this since coffea 0.7: https://github.com/scikit-hep/coffea/blob/36d0b31a05d4d1a4879f743d5067e36226a7ea0f/coffea/nanoevents/factory.py#L62-L64

`uproot_options` are general options you would pass to `uproot.open` or `uproot.dask` and we have this name in dataset tools as well. `iteritems_options` is something specific to `_extract_base_form` when doing the iteration. using `tree.iteritems(**iteritems_options)`.

Do you want to change this now or leave it for the future? It's going to break users code who rely on `allow_read_erros_wth_report` in `uproot_options`.

PR #1387, opened 2025-08-06
feat: support explicit bulk preloading (uproot-only)

pfackeldey:
This PR allows to bulk preload multiple branches with `NanoEventsFactory.from_root`. This is useful to group multiple reads into a single request. In the virtual mode the preloaded branches are still kept virtual (wrapped essentially by a lambda) to delay evaluating the DSL until the field is _actually_ needed, see:

```python
# explicitly preload "Jet_pt"
preload = lambda b: b.name == "Jet_pt"

  {"nanoaod.root": "Events"}, 
  mode="virtual", 
  access_log=(access_log := []), 
  preload=preload,  # new arg!
).events()

# this is expensive as it wasn't preloaded, needs to be loaded on the spot
%timeit -n 1 -r 1 ak.materialize(events.Jet["eta"])
# 468 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

# this is much faster because it's preloaded, but still the DSL has to run 
%timeit -n 1 -r 1 ak.materialize(events.Jet["pt"])
# 2.19 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

# check what has been accessed
print(access_log)
# ['nJet', 'Jet_eta', 'Jet_pt']

# this is much faster because it is preloaded _and_ the DSL has already been run
%timeit -n 1 -r 1 ak.materialize(events.Jet["pt"])
# 610 μs ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

In addition you can trigger bulk loading of everything (single read request for all columns) in the eager mode as follows:
```python
preload = lambda _: True  # match anything

  {"nanoaod.root": "Events"}, 
  mode="eager", 
  access_log=(access_log := []), 
  preload=preload,  # new arg!
).events()
```
Currently, coffea reads with a single request _per_ column.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -267,6 +267,7 @@ class NanoEventsFactory:
         decompression_executor=None,
         interpretation_executor=None,
         delayed=uproot._util.unset,
+        preload=None,
     ):
         """Quickly build NanoEvents from a root file
 
@@ -310,6 +311,9 @@ class NanoEventsFactory:
                 see: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_dask.py#L109
             interpretation_executor (None or Executor with a ``submit`` method):
                 see: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_dask.py#L113
+            preload (None or Callable):
+                A function to call to preload specific branches/columns in bulk. Only works in eager and virtual mode.
+                Passed to ``tree.arrays`` as the ``filter_branch`` argument to filter branches to be preloaded.
 
         Returns
         -------
@@ -415,6 +419,24 @@ class NanoEventsFactory:
             f"{entry_start}-{entry_stop}",
         )
         uuidpfn = {partition_key[0]: tree.file.file_path}
+
+        preloaded_arrays = None
+        if preload is not None:
+            preloaded_arrays = tree.arrays(
+                filter_branch=preload,
+                entry_start=entry_start,
+                entry_stop=entry_stop,
+                ak_add_doc=True,
+                decompression_executor=decompression_executor,
+                interpretation_executor=interpretation_executor,
+                how=dict,
+            )
+            # this ensures that the preloaded arrays are only sliced as they are supposed to be
+            preloaded_arrays = {
+                k: _OnlySliceableAs(v, slice(entry_start, entry_stop))
+                for k, v in preloaded_arrays.items()
+            }
+
         mapping = UprootSourceMapping(
             TrivialUprootOpener(uuidpfn, uproot_options),
             entry_start,
@@ -423,6 +445,7 @@ class NanoEventsFactory:
             access_log=access_log,
             use_ak_forth=use_ak_forth,
             virtual=mode == "virtual",
+            preloaded_arrays=preloaded_arrays,
         )
         mapping.preload_column_source(partition_key[0], partition_key[1], tree)
 
--- a/src/coffea/nanoevents/mapping/uproot.py
+++ b/src/coffea/nanoevents/mapping/uproot.py
@@ -122,6 +122,7 @@ class UprootSourceMapping(BaseSourceMapping):
         virtual=False,
         decompression_executor=None,
         interpretation_executor=None,
+        preloaded_arrays=None,
     ):
         super().__init__(
             fileopener=fileopener,
@@ -138,6 +139,7 @@ class UprootSourceMapping(BaseSourceMapping):
         self.interpretation_executor = (
             interpretation_executor or uproot.source.futures.TrivialExecutor()
         )
+        self.preloaded_arrays = preloaded_arrays
 
     @classmethod
     def _extract_base_form(cls, tree, iteritems_options={}):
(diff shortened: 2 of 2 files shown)


src/coffea/nanoevents/mapping/uproot.py
```diff
@@ -211,16 +213,22 @@ def extract_column(
...
-            entry_start=start,
-            entry_stop=stop,
-            decompression_executor=self.decompression_executor,
-            interpretation_executor=self.interpretation_executor,
-        )
+        if (
+            self.preloaded_arrays is not None
+            and columnhandle.name in self.preloaded_arrays
+        ):
+            the_array = self.preloaded_arrays[columnhandle.name][start:stop]
```
ikrommyd, line 220: I may be missing something but why do we need to slice it here? Don't the preloaded arrays already have entry start and entry stop specified?

src/coffea/nanoevents/factory.py
```diff
@@ -267,6 +267,7 @@ def from_root(
         decompression_executor=None,
         interpretation_executor=None,
         delayed=uproot._util.unset,
+        preload=None,
```
ikrommyd, line 270: I'd probably move this further up the list. Maybe after entry start and entry stop? We want this to be visible no?

src/coffea/nanoevents/factory.py
```diff
@@ -310,6 +311,9 @@ def from_root(
                 see: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_dask.py#L109
             interpretation_executor (None or Executor with a ``submit`` method):
                 see: https://github.com/scikit-hep/uproot5/blob/main/src/uproot/_dask.py#L113
+            preload (None or Callable):
+                A function to call to preload specific branches/columns in bulk. Only works in eager and virtual mode.
+                Passed to ``tree.arrays`` as the ``filter_branch`` argument to filter branches to be preloaded.
```
ikrommyd, line 316: Same comment as above, I'd not put this in the end.

src/coffea/nanoevents/mapping/uproot.py
ikrommyd: Oh nvm, is this to just satisfy `_OnlySliceableAs`? Then I'm stupid.

ikrommyd:
I like this @pfackeldey and have only left minor comments. I also like the fact that it is a callable like `filter_branch` is.

Is it possible to test something like this? I'm not sure because it's still gonna be a `VirtualArray` buffer but it's callable is just getting an array from this preloaded array dict. I don't think we can see it somehow.

One request though, is it possible to fill out the access log during preloading? I think it's best to see what has been loaded (and that's also a unit test). If that involves moving the access_log around, that's fine.

src/coffea/nanoevents/mapping/uproot.py
pfackeldey: I was explicitly following the logic for the normal case that specifies `start` and `stop` as well: https://github.com/scikit-hep/coffea/blob/master/src/coffea/nanoevents/mapping/uproot.py#L219-L220. The `_OnlySliceableAs` makes sure there's no bug, i.e. a different slice would be applied.

src/coffea/nanoevents/factory.py
pfackeldey: I didn't mean to mess with the order in case people pass positional args here - that would be a breaking change. Maybe coffea may want to force all of those options to be passed as kw-only at some point?

ikrommyd: Ohh yeah you're right about that. I guess I've never seen a person not pass kwargs that's why I said this. Indeed forcing kwargs would be best here from the coffea side.

src/coffea/nanoevents/mapping/uproot.py
ikrommyd: Yeah this is fine. It's a trivial slice.

pfackeldey:
> I like this @pfackeldey and have only left minor comments. I also like the fact that it is a callable like `filter_branch` is.
> 
> Is it possible to test something like this? I'm not sure because it's still gonna be a `VirtualArray` buffer but it's callable is just getting an array from this preloaded array dict. I don't think we can see it somehow.
> 
> One request though, is it possible to fill out the access log during preloading? I think it's best to see what has been loaded (and that's also a unit test). If that involves moving the access_log around, that's fine.

I didn't want to add them to the access_log, because it's an _access_-log and not a _loaded_-log. In my understanding, this should track what has been accessed through e.g. `getattr` or other ops. I would find it confusing if the access_log would contain information that was never needed, independently if it was loaded or not. Also, I think it's nice to be able to pass the access_log in a consecutive run to `preload` and thus not load data that wasn't used. Maybe I'm misinterpreting the idea of this log though?

src/coffea/nanoevents/factory.py
ikrommyd: @lgray what do you think? Should we enforce kwargs in a separate PR? Should we just break the order here and not care? I've actually never seen anybody use positional args here because the first argument `treepath` is something that's never used because no one is actually passing in  an `uproot.reading.ReadOnlyDirectory` to `from_root`.
`

ikrommyd:
I think the access log is meant to keep track of which branches were read overall. The executors also report back the bytesread via the following:
```py
                    if isinstance(file, uproot.ReadOnlyDirectory):
                        metrics["bytesread"] = file.file.source.num_requested_bytes
```
I may be miss-interpreting something though.

src/coffea/nanoevents/factory.py
lgray: Ah - sorry for missing this. I would take it to a separate PR. I hope no one is actually using those and positional arguments though.

lgray:
I'm going to enable auto-merge on this one and we can fix the args/kwargs things before the next release. I've lodged this to-do in an Issue.

PR #1398, opened 2025-08-19
feat: Make preprocess rigorous with IOFactory and pydantic dataclasses

NJManganelli:
This PR is a precursor to both parquet reading being re-enabled (pre-processing parquet data, most importantly) and column-joining (where dataset specifications become very complicated and hard-to-construct/validate by hand). This exposes and expands the explicit Dataclasses using Pydantic models. `preprocess` is updated to handle these dataclasses explicitly. Column-joining imposes some requirements on the status of a dataset for automatic column-determination, so a `joinable()` method is introduced to check that status after preprocessing, though some followup work may be generated by propagating the pydantic models through all the followup PRs.

Sidenote: These dataclasses are especially crucial for column-joining, as creating a joinable dataset spec becomes complicated when one must combine two disparate datasets (with different forms, if one wishes to understand their structure and determine necessary columns automatically) AND a non-trivial join specification (which tells column-joining how to handle the inputs to deliver a joined dataset to the user), and that spec needs to be preprocessed (separately for root and parquet sub-datasets to be joined) then recomposed into a joinable (i.e. non-Optional FileSpec) version of the joinable spec

For now, most of the simpler functions interacting with datasets have been updated with dual-path code to simultaneously support legacy pure-dictionary inputs and the pydantic models. This includes e.g. apply_to_(file|data)set, slice_(chunks|files), preprocess, etc. An example notebook including usage is introduced. In the future, once these are thoroughly tested in the wild, the old legacy dictionaries can be removed from code paths, since the Pydantic models already handle conversions trivially)

A followup PR will introduce the `preprocess_parquet` function. Another will be needed to handle threading the classes through the runner (where there's an open question of whether to harmonize the runner and preprocess+apply_to_fileset interfaces first). Another is expected to finish the changes for processing parquet again.

Tests should cover the overwhelming majority of cases, originally generated by Copilot, with heavy editing, expansion, corrections. The test code is not concise, unfortunately.

Pydantic automatically buys us serialization, as they can be saved into json via each model's `.model_dump()` function

This PR is an alternative to https://github.com/scikit-hep/coffea/pull/1395

NOTE: considering the huge number of commits, squash and merge may be desired instead...


NJManganelli:
@lgray @nsmith- @ikrommyd This is ready for a fresh set of eyes, particularly for anything I may have missed/overlooked so far.

Keep in mind, supporting runner, preprocess_parquet needs to come in followup PRs (latter will need to be converted for pydantic, former I have not considered yet - do we want to (semi-)unify runner and preprocess+apply_to_fileset interfaces first?)

pyproject.toml
```diff
@@ -58,6 +58,8 @@ dependencies = [
   "mplhep>=0.1.18",
   "packaging",
   "pandas",
+  "pydantic",
+  "eval_type_backport; python_version < '3.10'", #TODO: remove after python 3.9 EOL (filespec.py type Unions)
```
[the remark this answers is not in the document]
NJManganelli: Copilot says the math isn't mathing

src/coffea/dataset_tools/preprocess.py
```diff
@@ -499,14 +477,27 @@ def preprocess(
...
+                out_available[name]["form"] = None
 
-        if "metadata" not in out_updated[name]:
+        if is_datasetspec:
+            pass
+        elif "metadata" not in out_updated[name]:
             out_updated[name]["metadata"] = None
             out_available[name]["metadata"] = None
 
+    if is_datasetspec:
```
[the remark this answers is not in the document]
NJManganelli: This is a vulnerability. If the fileset is empty the logic will fail on undefined variable, and if the fileset is mixed and the last one is not a DatasetSpec, it won't trigger the return type (should be very unlikely)

NJManganelli:
Ping @lgray

lgray:
@NJManganelli looking at the volume of code that's been produced I assume this is AI generated for the most part? Not that it matters, just understanding how to approach it.

NJManganelli:
The core code (non-`tests`/`notebooks` code, in particular filespec.py) is 100% human-written (w/ occasional copilot suggestions/fixes, but it was the pydantic conversion of non-assisted code I wrote before).

For the tests, it was a mix, lets call it 2/3rds Copilot-initiated and 1/3rd by-hand (edits + consolidating via parametrization for the copilot tests, and updating older tests to explicitly parametrize Pydantic variation).

Notebook is almost entirely Copilot, with some edits for style, presentation, and corrections.

Hopefully that helps in what/how to review them respectively

lgray:
An initial comment looking through the user-facing parts`filespec.ipynb` the very last tutorial just stops without explaining everything it sets out to do. The rest of the tutorials seem well covering.

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,385 @@
...
+        else:
+            # validate the format, if present
+            if not IOFactory.valid_format(self.format):
+                raise ValueError(f"format: format must be one of {IOFactory._formats}")
+
+        return self
+
+    # @computed_field
+    # @property
+    def joinable(self) -> bool:
```
lgray, line 588: `joinable` and `check_form` share a lot of code - would try to condense to shared parts and reuse.

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,385 @@
...
+    def preprocess_data(cls, data: Any) -> Any:
+        data = copy.deepcopy(data)
+        if isinstance(data, FilesetSpec):
+            return data.model_dump()
+        return data
+
+
+def identify_file_format(name_or_directory: str) -> str:
+    root_expression = re.compile(r"\.root")
+    parquet_expression = re.compile(r"\.parq(?:uet)?$")
```
lgray: I've also seen `.pq` running around as well for parquet files, may want to add this.

The right way to do it for both would be to check the magic bytes in the files but that's very slow for this purpose.

lgray: Though, riffing on that, when we go to process the files in preprocess we could also validate the determined file type against the magic bytes. However, this is largely already accomplished just by assuming the extension is correct and opening with the appropriate package. 

Maybe we could give a more helpful error message? Then again I expect this kind of error (file with wrong extension) is very rare.

Moreover the actual filename is more a bookkeeping tool at the end of the day, so perhaps it's better to fall back to trying to open whatever ways we have available and notify the user (possibly by failing) if we exhausted all those options.

lgray:
Aside from that one additional comment and some pondering, I think this looks reasonable.

I appreciate the automation of writing core functionality unit tests.

@ikrommyd anything that catches your attention?

lgray:
As for unifying the actual processing and running interfaces. 

My idea around `apply_to_fileset` in the old-style processor based framework is starting settle towards it producing  generator of all the futures much like apply_to_fileset does in the dask case

Dask vs. non-dask can be driven by a boolean flag passed to `apply_to_fileset`, which should be the same as the `mode` flag for nanoevents and start out unset.

Then we introduce one new function `coffea.compute(computable, runner=None, **dask_kwargs)`, where `computable` is a dask collection or the future container that we produce from `apply_to_fileset` in the non-dask case. It then returns the result of `dask.compute` on the dask collection or it returns the accumulated result of running the given futures.

NJManganelli:
> An initial comment looking through the user-facing parts`filespec.ipynb` the very last tutorial just stops without explaining everything it sets out to do. The rest of the tutorials seem well covering.

Yeah, I deleted Copilot's suggestion for the IOFactory, partially because it was lying, and partially because while that class was the User-facing gateway to using the pure `dataclass` implementation (and Copilot treated it as such), Pydantic automates so much (for example, I had a function which would recurse down and convert the dataclass to pure dictionary, when the right option was set, or would only 'unwrap' the top-most layer and give you a dictionary with possibly-`dataclass` values(). With pydantic, these are respectively accomplished by just calling `AModel.model_dump()` and `dict(AModel)`). I was starting to think the IOFactory might be removed completely, but it's also the base class for the column-join IOJoinFactory, and that has a fair bit more work to do, and I haven't yet propagated this all through to see if that can be eliminated/converted to standalone functions yet. 

In short, I was thinking of not advertising the IOFactory much anymore for users, in case it makes more sense to break into standalone functions.

But I'll write an example for the few use-cases it still has

Let me also add the `.pq` support and consolidate the joinable/check code. 

For the magic-byte checking, lemme have a look into that in tandem with the preprocess_parquet PR. If I get that, I'll add it for both preprocess and preprocess_parquet (and maybe I can upstream some of the functionality from `join_preprocess` that's in column-join to coffea, a third of which is just dispatching root/parquet preprocessing to the appropriate function variation and re-`zip`ing the fileset together at the end. That would let us just unify into a single user-facing function too.

lgray:
Ah - OK if it's not user facing then we shouldn't talk about it unless we have to. :-) It just sorta stuck out to me!

For the magic-byte checking I started thinking too much about the right way to do it. The end of that thinking on my side is:
- I don't think we need to check the magic bytes, uproot and pyarrow.parquet do this for us
- what we should do is protect the user against feeding incomprehensible data to the system, and also tell the user when a file is mislabeled.

i.e. add in some logic if a file doesn't open to check if a root file is actually parquet (and vice versa), and if it's not either valid input format then we mark the file as bad, otherwise open the file as what it really is based on the filename guess and then make a note that it has the wrong postfix.

This is more a creature comfort than anything else, consider it low priority.

NJManganelli:
Sounds appropriate to me, do you think the format checking should be exclusively in e.g. preprocess, or both preprocess and apply_to_*set?

lgray:
That's a good question.

My first instinct is that once it's in `apply_to_*set` we should error rather than politely inform, since the user should providing that level of processing sanitized inputs of known-to-be-good files which it's at scale. When it's just messing around (like process/threads executors) we should allow globs, the jobs are typically fast enough at that scale to where time is not really wasted.

This is just the beginning of an idea though.

NJManganelli:
Okay, .pq formats accounted for, more tests for formats, factorized code for joinable and the datasetspec validation function (renamed), and some docs in the notebook for IOFactory showing off the still-relevant bits. 

Branch was out of date, rebased, and the tests are off to the races...

Ciao

lgray:
Giving it another look through I think it looks pretty good to me. I think it sets us up for a much more robust future when it comes to dataset communication, manipulation, and tracking.

@nsmith- @ikrommyd if y'all could have a look as well, at whatever depth of attention you can spare, I'd appreciate it.

lgray (approved):

ikrommyd (approved):
I've never used pydantic so I don't know good and bad patterns with pydantic. I've looked at this by looking at the tests and the example notebook mainly and I like the features and I think it sets up a good path for good dataset specifications and also the unification of executor-dask preprocessing.

lgray:
@nsmith- Could you give this a look since you've got the most pydantic experience among people I know in HEP.

NJManganelli:
@nsmith- do you have any preliminary thoughts that might dramatically change things? Otherwise I will start to work on the next couple PRs in the second half of this month, hoping only minor changes will be requested

src/coffea/dataset_tools/apply_processor.py
```diff
@@ -62,6 +60,8 @@ def apply_to_dataset(
         report : dask_awkward.Array, optional
             The file access report for running the analysis on the input dataset. Needs to be computed in simultaneously with the analysis to be accurate.
     """
+    if isinstance(dataset, DatasetSpec):
```
nsmith-: This seems a bit backwards from how I would expect: the interface prefers a narrow type (DatasetSpec) but allows a very generic type (any dict) and then the first line of the implementation weakens type, losing any type information for the remainder of the function. How come?

I would expect
```python
if isinstance(dataset, dict):
     dataset = DatasetSpec.model_validate(dataset)
```
or if the dict is not the same structure then to do some conversion

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+    from typing_extensions import Self
+
+from pydantic import BaseModel, Field, RootModel, computed_field, model_validator
+
+StepPair = Annotated[
+    list[Annotated[int, Field(ge=0)]], Field(min_length=2, max_length=2)
+]
+
+
+class UprootFileSpec(BaseModel):
```
nsmith-: A better name might be ROOTFileSpec, since ParquetFileSpec is for Parquet files and this is for ROOT files

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+    object_path: str
+    steps: Annotated[list[StepPair], Field(min_length=1)] | StepPair | None = None
+    format: Literal["root"] = "root"
```
nsmith-, line 149: You could inherit a `FileSpec` for the pieces that are common between the root and parquet files

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+class CoffeaFileDict(
+    RootModel[
+        dict[
+            str,
+            Union[
+                CoffeaUprootFileSpec,
+                CoffeaParquetFileSpec,
+                CoffeaUprootFileSpecOptional,
+                CoffeaParquetFileSpecOptional,
+            ],
```
nsmith-: This might be an opportunity to use a [discriminated union](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) using `format` as the discriminator. Though, if we want to keep separate types for the preprocessed and input file sets then we need a function discriminator. Personally I would separate the preprocessed and input at this level: CoffeaFileDict -> InputFiles, PreprocessedFiles

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+class CoffeaParquetFileSpecOptional(ParquetFileSpec):
+    num_entries: Annotated[int, Field(ge=0)] | None = None
+    uuid: str | None = None
+
+
+class CoffeaParquetFileSpec(CoffeaParquetFileSpecOptional):
+    steps: Annotated[list[StepPair], Field(min_length=1)] | StepPair
+    num_entries: Annotated[int, Field(ge=0)]
+    uuid: str
+    # directory: Literal[True, False] #identify whether it's a directory of parquet files or a single parquet file, may be useful or necessary to distinguish
```
nsmith-: `is_directory: bool`

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+
+class CoffeaUprootFileSpecOptional(UprootFileSpec):
+    num_entries: Annotated[int, Field(ge=0)] | None = None
+    uuid: str | None = None
+
+
+class CoffeaUprootFileSpec(CoffeaUprootFileSpecOptional):
+    steps: Annotated[list[StepPair], Field(min_length=1)] | StepPair
+    num_entries: Annotated[int, Field(ge=0)]
+    uuid: str
```
nsmith-: If I recall, we basically have two cases, something like `FileSpec` and `FileSpecPreprocessed` that dictate whether the extra fields exist, right?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+        return self.root.get(key, default)
+
+    def pop(self, key: str, default=...):
+        return self.root.pop(key, default)
+
+    def update(self, other=None, **kwargs):
+        self.root.update(other, **kwargs)
+
+
+class CoffeaFileDict(
```
nsmith-: What motivates using RootModel? To make it act like a dictionary so that existing code doesn't have to be touched?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,390 @@
...
+    def format(self) -> str:
+        """Identify the format of the files in the dictionary."""
+        union = set()
+        identified_formats_by_name = {
+            k: identify_file_format(k) for k, v in self.root.items()
+        }
+        stored_formats_by_name = {
+            k: v.format for k, v in self.root.items() if hasattr(v, "format")
+        }
+        assert all(
```
nsmith-:
```python
if not ...
    raise
```
but really this check should happen in a model validator function

nsmith- (commented):
I left a few comments now just to better understand the idea. Will look more.

src/coffea/dataset_tools/filespec.py
NJManganelli: I'd say both to ensure current patterns continue working, and because it makes sense to me that it remains iterable like a dict. The other thing that came to mind was completely eliminating the "filename" as a level, such that the FileSpec itself contained the path information. But I'm unsure that's in any way an improvement (rather than an orthogonal change), and more wary of re-architecting things unnecessarily.

src/coffea/dataset_tools/filespec.py
NJManganelli: What do you mean by separating the preprocessed and input levels? I suppose I've been thinking of this as "maybe-maybenot" ready-to-use and "definitively ready-to-use"; the latter may be accomplished via preprocessing, but may be manually set (if e.g. one doesn't need/care about the saved_form). As this all coalesces, maybe it makes sense to be even stricter, so that's not an objection, but where do the boundaries get defined?

src/coffea/dataset_tools/apply_processor.py
NJManganelli: I'll admit this was a bit of exhaustion-guided coding, and is just the minimal shim for users to try inserting the pydantic filesets. What I had been thinking was to start "opting" users in by, in one of the followup PRs, trying to coerce dictionaries to the pydantic classes in preprocessing/downstream, and then either converting back at the end (upon success) to not surprise users, or (upon a pydantic-path failure) falling back to pure-dictionary processing and raising a warning to users with a link to an issue where they'd be encouraged to upload their failing fileset, to start building a set of failure cases and address them. Then, sometime down the lines, fully deprecate the pure-dictionary inputs, with sufficient testing-in-the-wild behind us.

src/coffea/dataset_tools/filespec.py
NJManganelli: Should we consider support for directories with e.g. xrootd/root files in general, too? Similar to the format validation/magic-bytes discussion @lgray brought up, that's only possible to check rigorously by opening the paths/files, and I'm unsure what/how much is supported via fsspec and what alternative paths would be needed. But if this is the right moment to crystallize some handling of directories, lets do that?

src/coffea/dataset_tools/filespec.py
NJManganelli: Preprocessing (successfully) is sufficient, though not always necessary? I suppose "for sure" nobody is manually finding/calculating uuids though (under normal circumstances), but the "Optional" / "Concrete" binary has been floating in my head, naming-wise. Nevertheless, yes, minimal info vs fully-suffiicent-for-scaled-up-analysis

src/coffea/dataset_tools/filespec.py
NJManganelli: I think I left out format validation here because it should be enforced at the Dataset level just above this in the pydantic heirarchy, so it's a matter of whether we want to be very strict up-front and check at both levels? From the testing so far, I suppose that's more likely to give a clear validation/instantiation error to users.

src/coffea/dataset_tools/apply_processor.py
nsmith-: So we want to:
- for backwards compatibility, return dictionaries to users
- validate user input
- ensure our pydantic model exhaustively describes the currently supported input types

I would think all these can be accomplished by immediately running `DatasetSpec.model_validate` on the input, making sure that the body of the functions passes the type checker, and then casting back to dict on output.

NJManganelli: This would also work, it'll just have extra information (e.g.`format`) added in. I doubt many users are checking for exact equality of dictionaries (tests will just need modification)

NJManganelli: Also, is there anything special about model_validate versus calling `DatasetSpec(input_dict)`? Extra checks?

src/coffea/dataset_tools/filespec.py
NJManganelli: I've separted this into separate InputFiles (with automatic promotion) and PreprocessedFiles. There's not yet a discriminated union implementation, but I think that can be deferred to later, since at least for the moment that would mostly be a validation optimization, if I understand the purpose correctly.

src/coffea/dataset_tools/filespec.py
NJManganelli: Name is swapped, but this needs more thinking on how exactly `is_directory` is implemented, tested for (e.g. in `preprocessing` where directories/files are actually opened), and utilized (I think there are instances, particularly with parquet, where being able to treat things as a directory is advantageous. On the other hand, for column-joining, there'll be optimizations when the file-level view is utilized)

NJManganelli:
Would appreciate another look now, I addressed most comments. Things are not perfect, but in the interest of perfect not becoming the enemy of good, I hope it's reasonably close enough. Would like to get this merged and propagate changes to preprocessing and column-joining before circling back to those kinds of changes (the `is_directory` feature might make more sense after those downstream changes, too). And we have the unification of the runner / apply_to_X to consider as well. I might refactor preprocess in that followup PR to similarly try to force to the pydantic class and only fallback if it fails, which is the more aggressive form of getting things tested in the wild.

lgray:
@nsmith- could you please review after Nick's updates? Thanks!

src/coffea/dataset_tools/apply_processor.py
```diff
@@ -62,12 +60,14 @@ def apply_to_dataset(
...
+    if isinstance(dataset, dict):
+        dataset = DatasetSpec.model_validate(dataset)
+    maybe_base_form = dataset.form
     if maybe_base_form is not None:
         maybe_base_form = awkward.forms.from_json(decompress_form(maybe_base_form))
-    files = dataset["files"]
+    files = dataset.files
     events = NanoEventsFactory.from_root(
-        files,
+        files.model_dump(),
```
nsmith-, line 66: `NanoEventsFactory.from_root` could accept type `InputFiles`. This could be for a future PR since this changeset focuses on dataset_tools only. If so, let's make an issue

src/coffea/dataset_tools/apply_processor.py
```diff
@@ -62,12 +60,14 @@ def apply_to_dataset(
         report : dask_awkward.Array, optional
             The file access report for running the analysis on the input dataset. Needs to be computed in simultaneously with the analysis to be accurate.
     """
-    maybe_base_form = dataset.get("form", None)
+    if isinstance(dataset, dict):
+        dataset = DatasetSpec.model_validate(dataset)
+    maybe_base_form = dataset.form
     if maybe_base_form is not None:
         maybe_base_form = awkward.forms.from_json(decompress_form(maybe_base_form))
```
nsmith-: `DatasetSpec.form` should be a property that decompresses and returns the form if it is available, otherwise `None`. Then this code can be simplified to passing `known_base_form=dataset.form`

src/coffea/dataset_tools/apply_processor.py
```diff
@@ -119,10 +119,12 @@ def apply_to_fileset(
...
+    if isinstance(fileset, dict):
+        fileset = FilesetSpec.model_validate(fileset)
     out = {}
     report = {}
     for name, dataset in fileset.items():
-        metadata = copy.deepcopy(dataset.get("metadata", {}))
+        metadata = copy.deepcopy(dataset.metadata)
         if metadata is None:
             metadata = {}
         metadata.setdefault("dataset", name)
```
nsmith-, line 126: I think this is an example where the choice of `dict[str, UnnamedThing]` vs. `list[NamedThing]` makes it a little unergonomic: we could just pass a named dataset object to `apply_to_dataset` rather than copy and alter the metadata here. It's also not great for schemas to have high key cardinality. But we're locked into this by the legacy structure for now.

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+    format: str | None = None
+
+
+class ROOTFileSpec(GenericFileSpec):
+    object_path: str
+    format: Literal["root"] = "root"
+
+
+class CoffeaROOTFileSpecOptional(ROOTFileSpec):
+    num_entries: Annotated[int, Field(ge=0)] | None = None
```
nsmith-, line 25: Could `num_entries` be a computed property, deduced from `steps[-1][1]`?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+from pydantic import BaseModel, Field, RootModel, computed_field, model_validator
+
+StepPair = Annotated[
+    list[Annotated[int, Field(ge=0)]], Field(min_length=2, max_length=2)
+]
+
+
+class GenericFileSpec(BaseModel):
+    object_path: str | None = None
+    steps: Annotated[list[StepPair], Field(min_length=1)] | StepPair | None = None
```
nsmith-: What's the use case for a bare StepPair?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+                f"identified formats and stored formats do not match: identified formats: {identified_formats_by_name}, stored formats: {stored_formats_by_name}"
+            )
+        union.update(identified_formats_by_name.values())
+        if len(union) == 1:
+            return union.pop()
+        return "|".join(union)
+
+    @model_validator(mode="before")
+    def preproc_data(cls, data: Any) -> Any:
+        data = copy.deepcopy(data)
```
nsmith-, line 244: Is this to convert data that is not in the canonical form? Would it be possible to detect that and then decide to copy? Would it make more sense as a classmethod `from_legacy_dict()`?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+            str,
+            Union[
+                CoffeaROOTFileSpec,
+                CoffeaParquetFileSpec,
+                CoffeaROOTFileSpecOptional,
+                CoffeaParquetFileSpecOptional,
+            ],
+        ]
+    ],
+    DictMethodsMixin,
```
nsmith-: Use [collections.abc.Mapping](https://docs.python.org/3/library/collections.abc.html) (or MutableMapping) and implement just `__getitem__, __iter__, __len__` in this class. Also make sure that `__getitem__(self, key: str) -> GenericFileSpec`, or create a type alias for the union of the derived types it might return

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+
+    def __iter__(self) -> Iterable[str]:
+        return iter(self.root)
+
+    @model_validator(mode="after")
+    def promote_and_check_files(self) -> Self:
+        preprocessed = {k: False for k in self.root}
+        for k, v in self.root.items():
+            try:
+                if type(v) in [CoffeaROOTFileSpecOptional]:
```
nsmith-: `isinstance(v, CoffeaROOTFileSpecOptional)`

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+                    self.root[k] = CoffeaROOTFileSpec(v)
+                if type(v) in [CoffeaParquetFileSpecOptional]:
+                    self.root[k] = CoffeaParquetFileSpec(v)
+                preprocessed[k] = True
+            except Exception:
+                pass
+        if all(preprocessed.values()):
+            try:
+                self.root = PreprocessedFiles(self.root).root
```
nsmith-, line 380: This check seems redundant with the above. Anyway, is there any harm in letting `InputFiles` always hold optional FileSpec types?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+            except Exception:
+                pass
+        return self
+
+
+class PreprocessedFiles(
```
nsmith-, line 387: Is the design of this type still constrained by legacy format? If not, here we might consider changing it to `list[NamedObject]` for simlifying some of the backend code. Alongside, that would mean that the non-optional type would be separate from the optional one all the way down.

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+                preprocessed[k] = True
+            except Exception:
+                pass
+        if all(preprocessed.values()):
+            return self
+
+
+class DatasetSpec(BaseModel):
+    files: InputFiles
+    metadata: dict[Hashable, Any] | None = None
```
nsmith-: Should we place limits on metadata types to being json-serializable? i.e. `metadata: dict[str, str | int | float]` or so?
Also, rather than have optional dict, we can just make it always a dict with a default value `{}`.

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+    format: str | None = None
+    form: str | None = None
```
nsmith-: Change to `form_blob` or `compressed_form` or something and make `form` a property

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+
+    @model_validator(mode="before")
+    def preprocess_data(cls, data: Any) -> Any:
+        data = copy.deepcopy(data)
```
nsmith-: Can we check if it is already a `DatasetSpec` and pass it through here rather than the copy?

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+            data["files"] = files
+        elif isinstance(data, DatasetSpec):
+            data = data.model_dump()
+        elif not isinstance(data, DatasetSpec):
+            raise ValueError(
+                "DatasetSpec expects a dictionary with a 'files' key or a DatasetSpec instance"
+            )
+        return data
+
+    def check_form(self) -> bool | None:
```
nsmith-: Seems like this could be a validator that maybe also caches the decompressed and deserialized form for later access.

edit: ok later I see this is called by a validator. It should be private then, I think

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+
+    # @computed_field
+    # @property
+    def joinable(self) -> bool:
+        """Identify DatasetSpec criteria to be pre-joined for typetracing (necessary) and column-joining (sufficient)"""
+        if self.check_form() and self.set_check_format():
+            return True
+
+
+class FilesetSpec(RootModel[dict[str, DatasetSpec]], DictMethodsMixin):
```
nsmith-: As earlier, use MutableMapping and annotate the `__getitem__` to return `DatasetSpec` type

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+    elif "." not in name_or_directory.split("/")[-1]:
+        # could be a parquet directory, would require a file opening to determine
+        return "parquet"  # maybe "parquet?" to trigger further inspection?
+    else:
+        raise RuntimeError(
+            f"identify_file_format couldn't identify if the string path is for a root file or parquet file/directory for {name_or_directory}"
+        )
+
+
+class IOFactory:
```
nsmith-: In general, I think some IOFactory utilities would better be type classmethods, at least when the return type is fixed.
For any remainders, would suggest `DatasetConverter` or `ModelFactory` (since this doesn't really make an object that performs I/O)

src/coffea/dataset_tools/manipulations.py
nsmith-, line 1: I think basically all these free functions would be much more user-friendly if they were implemented on the `FileSpec` and related types themselves with method-chaining patterns, e.g. 
```python
spec = (
    spec.limit_steps(4, per_file=False)
    .limit_files(3)
    .filter_files(pattern='*DY*.root')
    .repartition(step_size=1000)
)
```

The methods can be implemented once in their most generic type, e.g. `GenericFileSpec.limit_steps`, with the parent containers also having the same function name that maps over their items.

We may want to leave these in for backwards-compatibility but they can then just call the method-chain versions.

`select_datasets_by_criteria` is another candidate for chaining, but I would suggest an interface more like `spec.filter_datasets(pattern="ttbar*", metadata={"key": "value"})`

Also, we're using `steps` and `chunks` in various places. Since uproot has consistently gone with `steps` we should probably move towards that for the names of the bound methods.

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,441 @@
...
+                        assert (
+                            v["format"] == "parquet"
+                        ), f"Expected 'format' to be 'parquet', got {v['format']} for {k}"
+        return data
+
+
+class InputFiles(
+    RootModel[
+        dict[
+            str,
```
nsmith-: Should we require the dict keys to be URLs here?

src/coffea/dataset_tools/preprocess.py
```diff
@@ -499,14 +481,27 @@ def preprocess(
...
         if union_form_jsonstr is not None:
             compressed_union_form = compress_form(union_form_jsonstr)
-            out_updated[name]["form"] = compressed_union_form
-            out_available[name]["form"] = compressed_union_form
+            if is_datasetspec:
+                out_updated[name].form = compressed_union_form
+                out_available[name].form = compressed_union_form
+            else:
+                out_updated[name]["form"] = compressed_union_form
+                out_available[name]["form"] = compressed_union_form
```
nsmith-: Why not always upconvert to DatasetSpec, mainpulate under that assumption, and then at the end downconvert if the user passed in a legacy object?

src/coffea/dataset_tools/preprocess.py
```diff
@@ -252,7 +227,7 @@ def _normalize_file_info(file_info):
```
nsmith-, line 228: Optional: add
```python
from typing import overload

@overload
def preprocess(
    fileset: FilesetSpec,
    step_size: None | int = None,
    align_clusters: bool = False,
    recalculate_steps: bool = False,
    files_per_batch: int = 1,
    skip_bad_files: bool = False,
    file_exceptions: Exception | Warning | tuple[Exception | Warning] = (OSError,),
    save_form: bool = False,
    scheduler: None | Callable | str = None,
    uproot_options: dict = {},
    step_size_safety_factor: float = 0.5,
    allow_empty_datasets: bool = False,
) -> tuple[FilesetSpec, FilesetSpec]: ...

@overload
def preprocess(
    fileset: dict,
    step_size: None | int = None,
    align_clusters: bool = False,
    recalculate_steps: bool = False,
    files_per_batch: int = 1,
    skip_bad_files: bool = False,
    file_exceptions: Exception | Warning | tuple[Exception | Warning] = (OSError,),
    save_form: bool = False,
    scheduler: None | Callable | str = None,
    uproot_options: dict = {},
    step_size_safety_factor: float = 0.5,
    allow_empty_datasets: bool = False,
) -> tuple[dict, dict]: ...
```

src/coffea/dataset_tools/preprocess.py
```diff
@@ -264,13 +239,13 @@ def preprocess(
     uproot_options: dict = {},
     step_size_safety_factor: float = 0.5,
     allow_empty_datasets: bool = False,
-) -> tuple[FilesetSpec, FilesetSpecOptional]:
+) -> tuple[FilesetSpec | dict]:
```
nsmith-: This should be `tuple[FilesetSpec, FilesetSpec] | tuple[dict, dict]`

nsmith- (changes requested):
First, thanks for implementing the last round of review! I'm trying to keep in mind the balance of not breaking existing types but we have a chance to really improve the ability to type-check the processing tools and I think its worth it even at this early stage of refactoring.

src/coffea/dataset_tools/filespec.py
nsmith-: If we want to support wildcard paths as well, then that's another variant. Basically, the key of the InputFiles dictionary is not so trivial: could be a local file, URL, local directory, directory URL, glob pattern, ...

src/coffea/dataset_tools/filespec.py
nsmith-: My personal preference (see also the later review comment on the return type of `preprocess()`) is we want to go for a type hierarchy more like

```python
class InputFileSpec(BaseModel):
    steps: Annotated[list[StepPair], Field(min_length=1)] | None
    format: Literal["root", "parquet"]

class ROOTInputSpec(InputFileSpec):
    format = "root"
    object_path: str
    uuid: str | None = None

class ParquetInputSpec(InputFileSpec):
    format = "parquet"

class InputFilesSpec(
    RootModel[
        dict[
            str,
            Union[
                ROOTInputSpec,
                ParquetInputSpec,
            ],
        ]
    ]
)

class FileSpec(BaseModel):
    path: str
    steps: Annotated[list[StepPair], Field(min_length=1)]
    format: Literal["root", "parquet"]

class CoffeaROOTFileSpec(FileSpec):
    format = "root"
    object_path: str
    uuid: str

class CoffeaParquetFileSpec(FileSpec):
    format = "parquet"
    uuid: str

class PreprocessedFilesSpec(
    RootModel[
        list[
            Union[
                CoffeaROOTFileSpec,
                CoffeaParquetFileSpec,
            ],
        ]
    ],
)

# and then similarly for
class InputDatasetSpec: ...

class DatasetSpec: ...
```

src/coffea/dataset_tools/apply_processor.py
NJManganelli: Former done, latter for propagation later. This will be a bit of a breaking change, as people expect form to be the compressed string representation, but I agree this makes sense. There'll be some breaking changes no matter what, so better to introduce now.

src/coffea/dataset_tools/apply_processor.py
NJManganelli: Aye, something to rethink later on

src/coffea/dataset_tools/filespec.py
NJManganelli: I guess that depends on what num_entries really needs to be used for: a quick summary of steps (which may be filtered from the full set in the original file) or the actual number of entries in the TTree

src/coffea/dataset_tools/preprocess.py
NJManganelli: I'm in agreement, will do this in the preprocess PR that has to follow this.

src/coffea/dataset_tools/filespec.py
NJManganelli: Would have to look back to even see if there was a good use case. I'm fine cutting it, and if people scream there's a good use-case, can add it back in.

src/coffea/dataset_tools/filespec.py
NJManganelli: I think there may have been two things, one was the multiple legacy formats, the other was avoiding modifying data in-place and screwing with inputs that might be reused (both in tests and the wild)

src/coffea/dataset_tools/filespec.py
NJManganelli: Not quite, this is the self-promotion of the type/container. I think it was necessary when I wrote it to actually get it working as intended. I'm loathe to even touch it now, at the risk of breaking things again and chasing them back down. 

If the InputFiles don't get force promoted somewhere, it'll eventually be "promotable" but not promoted, and then the subsequent checks to see if things are fully promoted and/or joinable have to try to promote. Do not want to alter now. This can be relaxed later if I've remembered wrong from that point.

src/coffea/dataset_tools/filespec.py
NJManganelli: Seems like a good idea, but I don't think we want to force it to be a flat dictionary (dict-in-dict-in-dict should be valid, even if over-engineered by users). Do you know a shorthand for doing so?

src/coffea/dataset_tools/filespec.py
NJManganelli: Done

src/coffea/dataset_tools/filespec.py
NJManganelli: I think this should actually stay as a deepcopy, because one of the expectations with the dictionary format is that each spec is independent. If you end up with a reference and modify it in-place in one dataset, you can propagate it unintentionally to others if e.g. they're (starting out) the same as from a template. Somewhere along the ways these deep copies were put in place because I got tripped up by unintentionally modifying like that.

src/coffea/dataset_tools/filespec.py
NJManganelli: Made sense to me as a user-facing function also, but now with form being the computable function which will effectively trip now, if there's something wrong with the compressed form, it can be made private

NJManganelli:
No merging for the moment, I'll do so if needed once I'm finished working on this round of updates. Also we seem to have a lot of errors being triggered in this fcc sample. Tertiary problem, but all the warnings about the ATLAS sample fields really fills up the test logs and gets in the way sometimes, can we clean that up somehow?

ikrommyd:
> No merging for the moment, I'll do so if needed once I'm finished working on this round of updates. Also we seem to have a lot of errors being triggered in this fcc sample. Tertiary problem, but all the warnings about the ATLAS sample fields really fills up the test logs and gets in the way sometimes, can we clean that up somehow?

If I remember correctly, that's an attribute of the schema you can turn off. Right?

src/coffea/dataset_tools/filespec.py
NJManganelli: Done

src/coffea/dataset_tools/filespec.py
NJManganelli: Done

src/coffea/dataset_tools/filespec.py
NJManganelli: This didn't work with original class inheritance I think, but does with the concrete inheriting from the optional type.

src/coffea/dataset_tools/filespec.py
NJManganelli: For PreprocessedFiles? We'd have to embed the key as the file/directory path instead, right?

src/coffea/dataset_tools/filespec.py
NJManganelli: What would this look like when supporting local files? Would there be a catch to insert the right prefixes?

src/coffea/dataset_tools/filespec.py
nsmith-: How is it currently used? I guess there's three options for the definition:
- number events in the tree
- `steps[-1][1]`
- `sum(hi-lo for lo, hi in steps)`

src/coffea/dataset_tools/filespec.py
nsmith-: Correct, what was the key of the dict becomes another field in the list item object

src/coffea/dataset_tools/filespec.py
nsmith-: Ok maybe this is something to think about when we have a signature more like
```python
def preprocess(x: InputFiles) -> PreprocessedFiles: ...
```
Is the idea that `join()` would only work on preprocessed files, a mix of the two, or something else?

src/coffea/dataset_tools/filespec.py
nsmith-: local could be `pathlib.Path` instances, that we canonicalize during preprocess (`path.resolve()`)

src/coffea/dataset_tools/filespec.py
nsmith-: It's probably not worth it at that point, unless there's a type alias somewhere in `json` that does what we need. Still, can we make it always present?

src/coffea/dataset_tools/filespec.py
nsmith-: Can you sketch what you mean? I'm just not quite following

src/coffea/dataset_tools/filespec.py
NJManganelli: Join will only work on preprocessed filespecs, and specifically the subset of those that have the form stored (so as to determine columns without opening files), so there's a method in there to check specifically for that additional criteria.

src/coffea/dataset_tools/filespec.py
NJManganelli: Lets try to discuss next week, if I can make it to the Coffea Users meeting that might be ideal

src/coffea/dataset_tools/apply_processor.py
NJManganelli: https://github.com/scikit-hep/coffea/issues/1449

src/coffea/dataset_tools/filespec.py
NJManganelli: Case 1 it seems: https://github.com/scikit-hep/coffea/blob/d5b7c1897da28c4473b9629af2bc9a2e5c75950a/src/coffea/nanoevents/factory.py#L416

What we could do is have a computable field, perhaps `num_selected` which could calculate from the steps. (`selected_entries` would more align with an explicit array of e.g. indices or run/event numbers in CMS case)

src/coffea/dataset_tools/filespec.py
NJManganelli: I got thrown off by the initial question, in the code there's already a model_dump for DatasetSpec, so the deepcopy (which is needed in case a dict is used as template for multiple specs) is moved inside the preprocessing of the dict path.

src/coffea/dataset_tools/filespec.py
NJManganelli: num_entries_in_steps is now a computed field. Verbose but clear

Better name suggestions welcome. `num_selected_entries`?

src/coffea/dataset_tools/filespec.py
NJManganelli: Enforcing keys to be URLs is maybe best left for preprocessing step, permitting str | pathlib.Path in InputFiles. Is there a dict-key compatible URL type we should use for PreprocessedFiles then?

src/coffea/dataset_tools/filespec.py
NJManganelli: @lgray, thoughts on making PreprocessedFiles list-like instead of dict-like, and moving the key/filepath into the type? Minimally we'd want, I guess, the `lfn` and `pfn` for remote files? And we resolve the pathlib.Path objects appropriately into a local file specification

src/coffea/dataset_tools/preprocess.py
NJManganelli: As discussed, we ignore this for now / add it into preprocessing PR

src/coffea/dataset_tools/filespec.py
nsmith-: Maybe https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.Url could be used?
It is hashable:
```
>>> from pydantic_core import Url
>>> Url
<class 'pydantic_core._pydantic_core.Url'>
>>> Url("http://blah")
Url('http://blah/')
>>> Url("http://blah").__hash__
<method-wrapper '__hash__' of pydantic_core._pydantic_core.Url object at 0x10064b900>
>>> Url("http://blah").__hash__()
4254906002024934011
```

src/coffea/dataset_tools/filespec.py
nsmith-: `num_selected_entries` is good IMO

src/coffea/dataset_tools/filespec.py
NJManganelli: And it would be great to have the dataset identifier -> this is the implicit "zeroth minus one" key for events, before "zeroth" filename and run, luminosityBlock, event, if one is thinking about joining

src/coffea/dataset_tools/manipulations.py
NJManganelli: prototypes of `limit_steps`, `limit_files`, `filter_files` and `filter_datasets` in place. The latter two have two kwargs, `filter_name` (ala uproot; regex matching on keys) and `filter_callable` (which applies to the "values" --> filespecs or datasetspecs). The latter I hope is generic enough without inventing half a dozen different ways of accomplishing things.

Still to-do: make the standalone functions in manipulations just call the method-chain versions

src/coffea/dataset_tools/filespec.py
NJManganelli: Lets leave the PreprocessedFiles as dict-like, but enforce the pfn(/lfn) to be present in the concrete FileSpec items. Should these be strings or (local) URL types?

NJManganelli:
We can subclass the DatasetSpec later on to e.g. CMSDatasetSpec. If it has a valid `did`, we can have a bunch of computable parameters that parse it to return: `year`, `run period`, `is_data`/`is_mc`, `NanoAOD version`, `processing version`, whether it's an `extension` (some of which can be used to more easily find correctionlib JSONS...), whether it can be combined with another dataset... etc. That's for the future though.

Another future item can be expanding the DatasetSpec to accept e.g. `DIDFinder` and `ParquetDirectory` as inputs, which might (not) get Preprocessed into fully expanded files

A third future item: validation and determination of lfn/pfn, etc. We've already discussed the utility of having the lfn such that if e.g. a specific replica is used and fails (repeatedly) we can fallback to a different redirector.

@nsmith- What add/subtract methods do you think would be needed? I am not familiar with all the variations and their main purpose, so if you let me know anything an AI summary might fail to highlight, that'd be good. 

I think these would naturally apply for both DatasetSpec ('concatenate' distinct files) and FileSpec (concatenate / set difference of `steps`). The DataGroupSpec can just be dict-like concatenation (+) and overlap removal (-)

src/coffea/dataset_tools/filespec.py
nsmith-: How about if the dict key is LFN and the FileSpec holds the PFN?

nsmith-: Actually for now maybe we should just go with PFN since adding location discovery to the preprocess is a bigger UI change

src/coffea/dataset_tools/filespec.py
NJManganelli: We'll want more support, so I think in the future, we make explicit InputDID, inputDir, etc. classes. The question of whether we canonicalize them all to an explicit list of files (and let compute or other interfaces potentially do smarter things with them, e.g. loading parquet as a dataset instead of a list of individual files?) is then left to later on.

src/coffea/dataset_tools/filespec.py
NJManganelli: I gave this a shot. In principle, it could work (pydantic has a FIleUrl, HttpUrl, and their base class AnyUrl). Unfortunately, a lot of things started to break taking this path. I am not sure it's worth trying to change the key type. Maybe the best thing to do is to ensure that the `pfn` and `lfn` are one of these acceptable types. Also, the AnyUrl cannot properly deduce the path when chained protocols are used (e.g. `blockcache::zip::s3://webserver:8090//store/path/to/file.type`), so it's maybe a little less useful than would be desirable. 

Keeping a generic key string lets one also have `duplicate` entries, for example processing the same file under different systematics trees (a real use case we have on ATLAS)...

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,870 @@
...
+                f"Can only subtract DataGroupSpec from DataGroupSpec, got {type(other)}"
+            )
+        new_dict = dict(self)
+        for k, v in other.items():
+            if k in new_dict:
+                new_dict[k] = new_dict[k] - v
+        return type(self)(new_dict)
+
+    @model_validator(mode="before")
+    def preprocess_data(cls, data: Any) -> Any:
```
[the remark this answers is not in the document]
NJManganelli: Grossly wrong, this must be cls or the code all breaks

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,870 @@
...
+            if self.did != other.did:
+                raise ValueError(
+                    f"Cannot add DatasetSpec with different dids: {self.did} and {other.did}"
+                )
+        new_spec = self.model_dump()
+        new_spec["files"] = self.files - other.files
+        return type(self)(**new_spec)
+
+    @model_validator(mode="before")
+    def preprocess_data(cls, data: Any) -> Any:
```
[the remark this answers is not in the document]
NJManganelli: Incorrect

src/coffea/dataset_tools/filespec.py
```diff
@@ -0,0 +1,870 @@
...
+        self.root[key] = value
+
+    def __delitem__(self, key: str):
+        del self.root[key]
+
+    def __len__(self) -> int:
+        return len(self.root)
+
+    @model_validator(mode="after")
+    def promote_and_check_files(self) -> Self:
```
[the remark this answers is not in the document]
NJManganelli: Explicit "demotion" to InputFiles inserted instead, to make the logic clear. Raising an error causes some failures, so I think that's not the right path (might change course after preprocessing and compute modules plus their interactions understood...)

NJManganelli:
@lgray @nsmith- I am done, barring new change requests.

docs/source/examples.rst
nsmith-, line 1: Was this file moved? Looks like the whole file is new now rather than just the filespec item.

docs/source/notebooks/filespec.ipynb
nsmith-, line 1: `v.model_dump(exclude="form")` should rather exclude `compressed_form`

nsmith-, line 1: Inaccessible files seems to still have pydantic models, as I understand we want preprocess to return either all models or all dicts?
```
{
    'ZJets': {},
    'Data': {
        'nano_dimuon_not_there.root': CoffeaROOTFileSpecOptional(
            object_path='Events',
            steps=None,
            num_entries=None,
            format='root',
            lfn=None,
            pfn=None,
            uuid=None,
            num_selected_entries=None
        )
    }
}
```

nsmith-, line 1: Maybe re-render? I see
```
DEBUG checking format
DEBUG A
DEBUG B
```

nsmith- (changes requested):

docs/source/examples.rst
NJManganelli: I think this is something that was merged in from master and then I added filespec.ipynb to the list? I don't know if something got broken in all the back-and-forth of those merge-from-master (I really prefer rebasing... can't handle these nonlinear histories)

docs/source/notebooks/filespec.ipynb
NJManganelli: I think we want to keep the compressed_form, the form is generated dynamically from the compressed_form

nsmith-: In the notebook where we're just demonstrating, there's a comment about suppressing it

docs/source/examples.rst
NJManganelli: Deleting this, I don't see it in master. Docs integration should probably wait until the compute (maybe also preprocessing) are integrated anyway, the models are deadends at the moment

docs/source/notebooks/filespec.ipynb
NJManganelli: Done

docs/source/examples.rst
nsmith-: It's now in a markdown file after the myst change https://github.com/scikit-hep/coffea/blob/master/docs/source/examples.md

docs/source/notebooks/filespec.ipynb
NJManganelli: ~What do you mean?~ Nvm, I see where this was attached now. github's interface is nigh-useless sometimes

NJManganelli: You're right. The form does not even need exclusion, it is by default because it's not a computed property, so this all checks out

nsmith-: yeah I couldn't put it at the line in the preview

NJManganelli: Should be pydantic in --> pydantic out, and vice-versa. This example (now I remember) deliberately converted back to suppress the "compressed_form" output (originally "form," hence the outdated reference here in the notebook).

NJManganelli:
@nsmith- All resolved I think

nsmith- (approved):
🚀

ikrommyd:
Let me just check the docs preview and it's good from my side if that's fine

ikrommyd:
Should be good now!

ikrommyd (approved):

PR #1400, opened 2025-08-21
feat: add dataset to the `FileMeta` hash to allow different metadata for the same file if used in multiple datasets

### LINKED ISSUE #1384: Different metadata attached to the same file with processor interface
**Describe the bug**
Fileset metadata with the old/new processor interface seems to depend on the input files instead of the top-level `fileset` keys. This means it is not possible to process the same files with different metadata. With `dask-awkward` this works fine.

**To Reproduce**
```python
import uproot
import awkward as ak
from coffea import dataset_tools
from coffea import processor
from coffea.nanoevents import BaseSchema
import dask

# create some input
with uproot.recreate("f1.root") as f:
   f["tree"] = {"arr": ak.Array([1])}
with uproot.recreate("f2.root") as f:
   f["tree"] = {"arr": ak.Array([1])}

fileset = {
    "nominal": {"files": {"f1.root": "tree"}, "metadata": {"variation": "nominal"}},
    "NP1": {"files": {"f1.root": "tree"}, "metadata": {"variation": "NP1"}}
}

# with dask-awkward
samples, _ = dataset_tools.preprocess(fileset)
tasks = dataset_tools.apply_to_fileset(
    lambda evts: print(f"dask-awkward: {evts.metadata['variation']}"),
    samples,
    schemaclass=BaseSchema
)
_ = dask.compute(tasks)


# with old / new processor interface
class Analysis(processor.ProcessorABC):
    def process(self, events):
        print(f"processor interface: {events.metadata['variation']}")
        return {}

        pass


run = processor.Runner(executor=processor.IterativeExecutor(), schema=BaseSchema)
samples = run.preprocess(fileset)
out = run(samples, processor_instance=Analysis())
```

**Expected behavior**
Seeing in the printout that the metadata differs for each `fileset` entry.

**Output**
```
dask-awkward: nominal
dask-awkward: NP1

processor interface: nominal
processor interface: nominal
```
Changing the `fileset` to use `"f2.root"` as input for the second entry allows the metadata to differ with both implementations.

`coffea` 2025.7.3

**Additional context**
n/a

ikrommyd:
Fixes https://github.com/scikit-hep/coffea/issues/1384

We use a cache to cache the metadata. The keys are the `hash` of the `FileMeta` objects. So when you preprocess a file from a dataset, its metadata are cached. The `hash` however, is only a function of the file name and the tree name, so when you preprocess the same file but from a different dataset, the code says "Oh I have the metadata for that in the cache, let's grab it from there". I'm solving it by making the hash a function of the dataset name too.

In general, it's bad to use the same file twice because you are sub-optimizing reading. They will be treated as two different files and be read separately. However, if people are allowed to do it, we shouldn't deliberately mess up their metadata I think.

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -62,7 +62,7 @@ class FileMeta:
         self.metadata = metadata
 
     def __str__(self):
-        return f"FileMeta({self.filename}:{self.treename})"
+        return f"FileMeta({self.dataset}:{self.filename}:{self.treename})"
 
     def __hash__(self):
         return _hash((self.dataset, self.filename, self.treename))
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -65,12 +65,15 @@ class FileMeta:
         return f"FileMeta({self.filename}:{self.treename})"
 
     def __hash__(self):
-        # As used to lookup metadata, no need for dataset
-        return _hash((self.filename, self.treename))
+        return _hash((self.dataset, self.filename, self.treename))
 
     def __eq__(self, other):
         # In case of hash collisions
-        return self.filename == other.filename and self.treename == other.treename
+        return (
+            self.dataset == other.dataset
+            and self.filename == other.filename
+            and self.treename == other.treename
+        )
 
     def maybe_populate(self, cache):
         if cache and self in cache:


ikrommyd:
@lgray I don't think this harms anybody. It's generally a poor choice to duplicate files across datasets but it's an allowed poor choice.

alexander-held:
Thanks for the fix! I agree that it is generally best to handle each file once, but I believe it is very useful to not restrict the user too much. I tend to use all kinds of patterns especially during debugging and flexibility there can be very helpful. I also think there are setups where running over the same files is still fairly natural, in particular when you want to split the file content across multiple histograms and run your full analysis logic after an initial cheap cut that splits up the file. Instead of doing a loop within the processor (which might make each chunk be a lot slower to process for this specific sample), running the main analysis logic once seems reasonable to me.

I think the update also makes the behavior more intuitive, whether or not people chose to take advantage of the functionality it unlocks.

PR #1403, opened 2025-08-23
feat: enable eager/virtual parquet reading

ikrommyd:
It looks easy to turn on enable/virtual parquet reading.
I had to load and re-save the test parquet files without the awkward/pyarrow extension array disabled to be able to load them because it's causing many problems and extension types are erroring on `.to_pandas_dtype()`

This PR is only meant to add it to the factory and another PR will enable parquet to be processed with executors.


lgray:
@ikrommyd is there anything else you want to do here before we review? shall I mark as ready?

ikrommyd:
Uhmm I’m just not sure if there is anything we can do regarding parquet files with the awkward pyarrow extension types. Such files will currently error because these types don’t have a pandas type. Also, it seems like we can’t read ANY parquet  file. Files with nested records will not work. That’s the same for coffea 0.7 of course but I’m not sure if we can do anything about that. What do you think?

NJManganelli:
I swear I tested both extensionarray True and False in my branch. I thought there was some breaking change in pyarrow+awkward somewhere, and I made both version of parquet files fresh (extensionarray True and False) and they both were loadable. Right NOW they're loadable. Did extensionarray conversion (as called via `hepconvert`) get silently removed/forced to False? If not....

Using this branch:
https://github.com/NJManganelli/coffea/tree/datafactory_parquet

Screenshot 2025-08-28 at 23 45 00

ikrommyd:
@NJManganelli are you not getting this with the file on master branch? This is what I'm getting with this branch and `nano_dy.parquet` from master.
```
In [1]: from coffea.nanoevents import NanoEventsFactory, NanoAODSchema

In [2]: x = NanoEventsFactory.from_parquet("tests/samples/nano_dy.parquet", schemaclass=NanoAODSchema).events()
NotImplementedError                       Traceback (most recent call last)
Cell In[2], line 1
----> 1 x = NanoEventsFactory.from_parquet("tests/samples/nano_dy.parquet", schemaclass=NanoAODSchema).events()

File ~/Dropbox/work/coffea_dev/coffea/src/coffea/nanoevents/factory.py:590, in NanoEventsFactory.from_parquet(cls, file, treepath, entry_start, entry_stop, runtime_cache, persistent_cache, schemaclass, metadata, parquet_options, skyhook_options, access_log, mode)
    584     shim = TrivialParquetOpener.UprootLikeShim(
    585         table_file, dataset, openfile=fs_file
    586     )
    588 mapping.preload_column_source(partition_key[0], partition_key[1], shim)
--> 590 base_form = mapping._extract_base_form(table_file.schema_arrow)
    592 return cls._from_mapping(
    593     mapping,
    594     partition_key,
   (...)    600     mode,
    601 )

File ~/Dropbox/work/coffea_dev/coffea/src/coffea/nanoevents/mapping/parquet.py:164, in ParquetSourceMapping._extract_base_form(cls, arrow_schema)
    162 else:
    163     schema = field.type
--> 164     form = arrow_schema_to_awkward_form(schema)
    165     form = json.loads(form.to_json())
    167 if (
    168     form["class"].startswith("ListOffset")
    169     and form["content"]["class"] == "NumpyArray"  # noqa
    170 ):

File ~/Dropbox/work/coffea_dev/coffea/src/coffea/nanoevents/mapping/parquet.py:84, in arrow_schema_to_awkward_form(schema)
     76     return awkward.forms.ListOffsetForm(
     77         offsets="i64",
     78         content=awkward.forms.NumpyForm(
   (...)     81         ),
     82     )
     83 elif isinstance(schema, pa.lib.DataType):
---> 84     dtype = schema.to_pandas_dtype()()
     85     return awkward.forms.NumpyForm(
     86         inner_shape=[],
     87         primitive=dtype.dtype.name,
     88     )
     89 else:

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/pyarrow/types.pxi:410, in pyarrow.lib.DataType.to_pandas_dtype()

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/pyarrow/types.pxi:199, in pyarrow.lib._to_pandas_dtype()

NotImplementedError: ak:uint32
```

ikrommyd:
I get the same thing with your branch and the parquet file from master.

NJManganelli:
Yes, there was a breaking change along the way and I don't remember half of what I had tracked down, but try creating *new* files with and without extensionarray options. If that machinery still works and isn't silently flipping to False always, then we can still handle the extensionarray.

ikrommyd:
Ah then it should be good then.

NJManganelli:
BTW, in my branch, and given there's "no cost" to do this, I just renamed (`git mv`) the old parquet files as "nano_dy_semver.parquet" and likewise for the data file. That would permit keeping it exposed for testing, e.g. with this extensionarray failure on the old version of the file. I don't know if in the future we have a workaround to support already-made files like that (not everyone will be able to recreate or roundtrip via hepconvert, I expect), but our index size will be the same whether it's renamed or deleted, so maybe we should keep it exposed and add an XFAIL test for reading them?

ikrommyd:
I just converted with the following code and I'm getting the `NotImplementedError: ak:uint32` error? Can you send me your converted file with extensionarray? Or how you converted it?
```py
import awkward as ak
import pyarrow.parquet as pq

x = ak.to_arrow_table(ak.from_parquet("tests/samples/nano_dimuon.parquet"))
pq.write_table(x, "tests/samples/nano_dimuon_extensionarray.parquet")
```

ikrommyd:
I really don't know how you made it work 🤣 
```py
In [24]: x = uproot.open("tests/samples/nano_dy.root")["Events"].arrays()

In [27]: schema = ak.to_arrow_table(x, extensionarray=False).schema

In [28]: schema.types[0].to_pandas_dtype()
Out[28]: numpy.uint32

In [29]: schema.types[0]
Out[29]: DataType(uint32)

In [30]: schema = ak.to_arrow_table(x, extensionarray=True).schema

In [31]: schema.types[0]
Out[31]: awkward<DataType(uint32)>

In [32]: schema.types[0].to_pandas_dtype()
NotImplementedError                       Traceback (most recent call last)
Cell In[32], line 1
----> 1 schema.types[0].to_pandas_dtype()

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/pyarrow/types.pxi:410, in pyarrow.lib.DataType.to_pandas_dtype()

File ~/micromamba/envs/awkward-coffea/lib/python3.13/site-packages/pyarrow/types.pxi:199, in pyarrow.lib._to_pandas_dtype()

NotImplementedError: ak:uint32
```

NJManganelli:
Ah, I never committed that to the branch, you have to bump the pyarrow version. I've tested both 19.0.0 and 21.0.0
I tried directly committing to the PR but I suppose I should have pushed to your fork/branch instead. I'll leave it to you, the codespace won't spawn. BTW, with those version, I can directly run your two lines for going from the old semver file to arrow table to parquet file without erroring out at that stage.

NJManganelli:
And I converted using hepconvert.

```
import hepconvert
hepconvert.root_to_parquet("tests/samples/nano_dy.root", "tests/samples/nano_dy_extensionarray.parquet", tree="Events", extensionarray=True)
```

ikrommyd:
That unfortunately doesn't create extension types. And it's readtable.
```
In [1]: import hepconvert
   ...: hepconvert.root_to_parquet("tests/samples/nano_dy.root", "tests/samples/nano_dy_extensionarray.parquet", tree="Events", extensionarray=True)

In [2]: import pyarrow.parquet as pq

In [3]: schema = pq.read_schema("tests/samples/nano_dy_extensionarray.parquet")

In [4]: schema.types[0]
Out[4]: DataType(uint32)
```
To arrow table on the other hand does and that is not readable if I write it out.
```
In [7]: x = uproot.open("tests/samples/nano_dy.root")["Events"].arrays()

In [10]: schema = ak.to_arrow_table(x, extensionarray=True).schema

In [11]: schema.types[0]
Out[11]: awkward<DataType(uint32)>
```

NJManganelli:
Then extensionarray is dead for us, at least for now. Martin said somewhere it's preferable we don't use them anyway. 

If we could gracefully inform the user when that error is encountered, that they can rewrite the files without extensionarray (and how), that would be nice, but otherwise...

ikrommyd:
I don't even know why `to_parquet` doesn't even write it with extension array but I'm very tired of the consistent problems that I don't care anymore. I think that awkward should probably default to `extensionarray = False`.

We can probably add an error there at the `to_pandas_dtype()` call.

NJManganelli:
https://github.com/scikit-hep/awkward/pull/3127

This stores the info needed to reconstruct extensionarrays in metadata, because extensionarrays weren't/aren't compatible with selective reading. So, that explains this I think... indeed the `extensionarray=True` doesn't serialize that way. So...

ikrommyd:
I may have managed to patch it in https://github.com/scikit-hep/coffea/pull/1403/commits/64b0c6f7fde7919ea4e854a4329cd351f7233427. I'm basically trying to recursively look inside the extension type to get the "proper" pyarrow type. The tests pass now with both type of parquet files.

ikrommyd:
This is ready from my side.

PR #1409, opened 2025-09-03
feat: do not enforce dask awkward arrays in jetmet tools

ikrommyd:
These `dak.from_awkward` calls for arrays containing virtual buffers are very problematic. They cause headaches and are problematic with pickling especially when a distributed client is alive.
They also force the user to do a `.compute()` on the output of the jetmet tools which has been very confusing to the users so far. Creating dask-awkward arrays when using executors is very counter-intuitive. It should be avoided

### DIFF
--- a/src/coffea/jetmet_tools/FactorizedJetCorrector.py
+++ b/src/coffea/jetmet_tools/FactorizedJetCorrector.py
@@ -6,7 +6,6 @@ import dask_awkward
 import numpy
 
 from coffea.lookup_tools.jme_standard_function import jme_standard_function
-from coffea.util import maybe_map_partitions
 
 
 def _checkConsistency(against, tocheck):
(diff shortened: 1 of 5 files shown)


src/coffea/jetmet_tools/FactorizedJetCorrector.py
```diff
@@ -179,7 +180,7 @@ def getCorrection(self, **kwargs):
                 behavior=zl_out.behavior,
             )
 
-            return dask_awkward.map_partitions(
+            return maybe_map_partitions(
```
lgray: This shouldn't be necessary given the if block that it lives in? It checks if dask_awkward otherwise it eagerly evaluated.

ikrommyd: That's true. Thanks for catching it.

ikrommyd:
Regarding preliminary testing so far, I have added breakpoints here:
https://github.com/scikit-hep/coffea/blob/ab56206112a6e89eb748cc7948951395dfcc9589/tests/test_jetmet_tools.py#L783
and here:
https://github.com/scikit-hep/coffea/blob/ab56206112a6e89eb748cc7948951395dfcc9589/tests/test_jetmet_tools.py#L937
and doing
```py
jet_factory.build(jets.compute())
met_factory.build(met.compute(), corrected_jets.compute())
```
works fine without errors.

ikrommyd:
It is now tested by running the full corrected jet/met factory in virtual mode too. From my side, this is ready

lgray (approved):

PR #1415, opened 2025-09-08
fix(tests): keep dask jetmet tool tests alive

lgray:
@ikrommyd FYI

### DIFF
--- a/tests/test_jetmet_tools.py
+++ b/tests/test_jetmet_tools.py
@@ -715,7 +715,8 @@ def test_jet_resolution_sf_2d(optimization_enabled):
         )
 
 
-def test_corrected_jets_factory():
+@pytest.mark.parametrize("mode", ["virtual", "dask"])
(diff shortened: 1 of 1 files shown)


ikrommyd:
Uhmmm there is a test right below that's called `test_corrected_jets_factory_dak`. I didn't remove anything. The extra `.compute()` calls of the dask test made sense to keep them separate (also the fact that it parametrizes over dak optimizaton and also uses a client).
`

ikrommyd:
https://github.com/scikit-hep/coffea/blob/d9488d2c4e7cb6dd87af9f92dd04171b76c31fb4/tests/test_jetmet_tools.py#L942-L944

lgray:
long test file! but why is it failing here.

lgray:
Ahh - I get it, it does stuff with concrete arrays in this one. OK - just a misreading. I became concerned.

Issue #1393, opened 2025-08-13
Robust processing: partial aggregation, recovery, resubmission


alexander-held:
**Is your feature request related to a problem? Please describe.**
When large-scale processing runs into intermittent issues (e.g. I/O problems), this can crash the entire graph and require a full resubmission. If random failures happen at a sufficiently large rate, users need to manually split their input into smaller graphs and keep retrying them. It would be convenient to have centrally available functionality that can help with this.

There are two kinds of issues to distinguish between:
- I/O related problems, or anything outside of user-controlled code,
- exceptions raised in user code.
The latter case can be caught and handled by user code, e.g. by returning a default value alongside metadata capturing the information that a given chunk requires resubmission. The former case requires changes in `_work_function` and is therefore somewhat out of user control.

**Describe the solution you'd like**
Functionality along the lines of `allow_read_errors_with_report` from `uproot.dask`: the possibility to run over all successful `WorkItem` and capture those that failed. This could be combined with something like `dataset_tools.get_failed_steps_for_dataset` for users to resubmit.

One limitation of such an approach is that normalizations in `postprocess` are trickier to handle as a sample may not be complete yet when this runs, so they would probably have to be done standalone once all jobs have succeeded instead.

Here is an example of what I am thinking about conceptually:
```python
import random

import coffea
from coffea import processor
from coffea.nanoevents import BaseSchema, NanoEventsFactory
import dask.bag
import hist


def default_histogram():
    """need(?) a structurally correct default result"""
    return  hist.Hist.new.Regular(10, 0, 1000).Weight()


def my_analysis(workitem):
    """equivalent of process() in a processor, but includes I/O as well (to catch errors)"""
    h = default_histogram()

    try:
        events = NanoEventsFactory.from_root(
            {workitem.filename: workitem.treename},
            entry_start=workitem.entrystart,
            entry_stop=workitem.entrystop
        ).events()
        h.fill(events.Jet.pt[:, 0])

        # simulate random failures
        if random.random() < 0.5:
            raise ValueError

        return {"hist": h, "failed_items": set()}  # success

    except:
        return {"hist": default_histogram(), "failed_items": {workitem}}  # track failure


def reduce(x, y):
    """how to combine partial results"""
    return {"hist": x["hist"] + y["hist"], "failed_items": x["failed_items"] | y["failed_items"]}


[...]

PR #1420, opened 2025-09-24
feat: add checkpointing functionality to coffea.processor.Runner

pfackeldey:
This is a first try to add optional checkpointing to coffea workflows that enable resumable workflows. For this, a new `CheckpointerABC` exists that can be subclassed to implement custom `load` and `save` methods (saving and loading the checkpoint). Those can be highly custom, i.e. only checkpointing TTbar chunks or only if we have a full moon...
A `LocalCheckpointer` implementation is provided (can be used if people have access to an NFS for example).

The usage is basically as follows (see also the doc string of `CheckpointerABC`):
```python
from datetime import datetime
from coffea import processor

# create a checkpointer that stores checkpoints in a directory with the current date/time
# (you may want to use a more specific directory in practice)
today = datetime.now().strftime("%Y%m%d")
checkpointer = processor.LocalCheckpointer(checkpoint_dir=f"checkpoints/{today}", verbose=True)

# pass the checkpointer to a Runner
run = processor.Runner(..., checkpointer=checkpointer)
output = run(...)
```
After the run, the checkpoints will be stored in the directory `checkpoints/{today}`. On a subsequent run, if the same chunks are processed (and the same checkpointer, or rather `checkpoint_dir` is used), the results will be loaded from disk instead of being recomputed.

The basic idea is that it checkpoints each chunk output. It is not clear to me how to checkpoint a tree-reduction, which is why this is not implemented here. Thus, re-running with everything checkpointed still means to run the tree-reduction step - but the bulk processing is not.

I verified that it works locally. All of this of course comes at the price of additional IO, but I fear there's no way to achieve resumable workflows otherwise. That's why this is opt-in.

Let me know what you think :)

### DIFF
--- a/tests/test_checkpointing.py
+++ b/tests/test_checkpointing.py
@@ -64,13 +64,10 @@ def test_checkpointing():
     # number of WorkItems
     n_expected_checkpoints = len(chunks)
     ntries = 0
-    fs, token, paths = fsspec.get_fs_token_paths(checkpoint_dir)
+    fs, path = fsspec.url_to_fs(checkpoint_dir)
 
     # keep trying until we have as many checkpoints as WorkItems
-    while (
-        len(list(filter(fs.isfile, fs.glob(f"{paths[0]}/**"))))
-        != n_expected_checkpoints
-    ):
+    while len(list(filter(fs.isfile, fs.glob(f"{path}/**")))) != n_expected_checkpoints:
         fs.invalidate_cache()
         ntries += 1
         try:
@@ -81,13 +78,10 @@ def test_checkpointing():
 
     # make sure we have as many checkpoints as WorkItems
     fs.invalidate_cache()
-    assert (
-        len(list(filter(fs.isfile, fs.glob(f"{paths[0]}/**"))))
-        == n_expected_checkpoints
-    )
+    assert len(list(filter(fs.isfile, fs.glob(f"{path}/**")))) == n_expected_checkpoints
 
     # make sure we got the right answer
     assert out == {"cutflow": {"Data_pt": np.int64(84), "ZJets_pt": np.int64(18)}}
 
     # cleanup
-    fs.rm(paths[0], recursive=True)
+    fs.rm(path, recursive=True)
--- a/tests/test_checkpointing.py
+++ b/tests/test_checkpointing.py
@@ -46,7 +46,7 @@ def test_checkpointing():
 
     executor = processor.IterativeExecutor()
 
-    checkpointer = processor.LocalCheckpointer(
+    checkpointer = processor.SimpleCheckpointer(
         path := (Path(__file__).parent / "test_checkpointing")
     )
     run = processor.Runner(
--- a/tests/test_checkpointing.py
+++ b/tests/test_checkpointing.py
@@ -56,23 +56,28 @@ def test_checkpointing():
         format="root",
         checkpointer=checkpointer,
     )
+    # use the chunk generator to not re-run the preprocessing step
+    chunks = list(run.preprocess(filelist, "Events"))
+
+    def chunk_gen():
+        yield from chunks
 
     # number of WorkItems
-    expected_checkpoints = len(list(run.preprocess(filelist, "Events")))
+    n_expected_checkpoints = len(chunks)
     is_file = operator.methodcaller("is_file")
     ntries = 0
 
     # keep trying until we have as many checkpoints as WorkItems
-    while len(list(filter(is_file, path.rglob("*")))) != expected_checkpoints:
+    while len(list(filter(is_file, path.rglob("*")))) != n_expected_checkpoints:
         ntries += 1
         try:
-            out = run(filelist, UnstableNanoEventsProcessor(), "Events")
+            out = run(chunk_gen(), UnstableNanoEventsProcessor(), "Events")
         except Exception:
             print(f"Run failed, trying again, try number {ntries}...")
             continue
 
     # make sure we have as many checkpoints as WorkItems
-    assert len(list(filter(is_file, path.rglob("*")))) == expected_checkpoints
+    assert len(list(filter(is_file, path.rglob("*")))) == n_expected_checkpoints
 
     # make sure we got the right answer
     assert out == {"cutflow": {"Data_pt": np.int64(84), "ZJets_pt": np.int64(18)}}
--- a/tests/test_checkpointing.py
+++ b/tests/test_checkpointing.py
@@ -1,10 +1,9 @@
-import operator
 import os.path as osp
 import random
-import shutil
 from pathlib import Path
 
 import awkward as ak
+import fsspec
 import numpy as np
 
 from coffea import processor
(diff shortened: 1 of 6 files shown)


lgray:
As a test, could you make a processor that has some moderate probability of failure and then the test succeeds when you get back a list of results that's the length of the inputs? Or something to that effect.

pfackeldey:
> As a test, could you make a processor that has some moderate probability of failure and then the test succeeds when you get back a list of results that's the length of the inputs? Or something to that effect.

done. Hope this is an acceptable IO/file handling in tests.

ikrommyd:
The test is fine IMO but I'm wondering about something else. There are cases where the workers don't have access to write the checkpoints to some disk.
@lgray should this update `coffea.util.load/save` too to use `fsspec` to write/load checkpoints from remote storages. I don't think `lz4.frame.open` handles them unless I'm mistaken.

pfackeldey:
> The test is fine IMO but I'm wondering about something else. There are cases where the workers don't have access to write the checkpoints to some disk. @lgray should this update `coffea.util.load/save` too to use `fsspec` to write/load checkpoints from remote storages. I don't think `lz4.frame.open` handles them unless I'm mistaken.

That's why I called it the "LocalCheckpointer". People can subclass the `CheckpointerABC` and implement the saving and loading as they wish (potentially using `fsspec`). I wasn't sure if a general `RemoteCheckpointer` can be implemented, if so, I'm lacking the knowledge of `fsspec` to do so. If you know how to, please add it.

ikrommyd:
For the most part I think it's just `with fsspec.open(...)`. I don't know the equivalent of `lz4.frame.open` though in fsspecese.

ikrommyd:
Also one more comment here for reference is that this will also capture the cases where the processor does not return anything meaningful and just writes an output parquet to disk. The reason is that processors are not allowed to `return None`. They need something they can accumulate even if that something is just a single integer which will be saved by the checkpoint if I understand correctly. Meaning you still know that the processor finished.

pfackeldey:
> Also one more comment here for reference is that this will also capture the cases where the processor does not return anything meaningful and just writes an output parquet to disk. The reason is that processors are not allowed to `return None`. They need something they can accumulate even if that something is just a single integer which will be saved by the checkpoint if I understand correctly. Meaning you still know that the processor finished.

In these cases it doesn't make much sense to checkpoint, because you kind of do that already by writing to parquet in each processor-chunk processing.

ikrommyd:
Yes but the checkpointer offers an easy way to keep track of which files you were able to run over.

lgray:
Guess one has to check what's installed?

https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.available_compressions

From examples looks pretty straightforward?
```python3
import fsspec
import lz4.frame # You might need to explicitly import lz4.frame for frame-based compression

# Create some data to compress
data = b"This is some data that will be compressed using LZ4."

# Write compressed data to a file using fsspec
with fsspec.open("my_file.lz4", "wb", compression="lz4") as f:
    f.write(data)

# Read compressed data from the file using fsspec
with fsspec.open("my_file.lz4", "rb", compression="lz4") as f:
    read_data = f.read()

print(read_data)
```

All one needs is to have lz4 installed in the executing environment.

I'll play around with this later today.

ikrommyd:
Done. Updated utils to load/dump with fsspec:
```
In [1]: from coffea.util import save, load

In [2]: save(1, "root://cmseos.fnal.gov//store/user/ikrommyd/test/dummy.coffea")

In [3]: load("root://cmseos.fnal.gov//store/user/ikrommyd/test/dummy.coffea")
Out[3]: 1
```
@pfackeldey could you check that the checkpointing is fine in the example you have in your machine?

lgray:
Cool thanks for doing that! Nice that it worked out of the box as advertised.

pfackeldey:
> Done. Updated utils to load/dump with fsspec:
> 
> ```
> In [1]: from coffea.util import save, load
> 
> In [2]: save(1, "root://cmseos.fnal.gov//store/user/ikrommyd/test/dummy.coffea")
> 
> In [3]: load("root://cmseos.fnal.gov//store/user/ikrommyd/test/dummy.coffea")
> Out[3]: 1
> ```
> 
> @pfackeldey could you check that the checkpointing is fine in the example you have in your machine?

Given that the test passes, everything is good :+1:. Do you want me to rename the `LocalCheckpointer` to `FSSpecCheckpointer` or `SimpleCheckpointer` or something else?

ikrommyd:
I vote for `SimpleCheckpointer` since it's not "local" anymore.

ikrommyd:
And I'm overall happy with the status of this as a first implementation. We can get this in and get some feedback and move forward from there. I don't want to start thinking of ways to make this more featurefull but also more complicated at this point.

pfackeldey:
> And I'm overall happy with the status of this as a first implementation. We can get this in and get some feedback and move forward from there. I don't want to start thinking of ways to make this more featurefull but also more complicated at this point.

The only thing that is missing now is the `.mkdir` and `.exists` logic should probably be done with fsspec now in the `SimpleCheckpointer` instead of `pathlib`?

ikrommyd:
As far as I'm aware, fsspec makes the dirs for you if you have permissions. If you wanna check for some existence before trying to read to raise a better error yeah it should be with fsspec now. Creation though is not needed with fsspec I think.
```
In [1]: from coffea.util import save, load

In [2]: save(1, "root://cmseos.fnal.gov//store/user/ikrommyd/haha/hihi/dummy.coffea")

In [3]: save(2, "haha/hihi/tmp.coffea")

In [4]: exit
(awkward-coffea) ➜  coffea git:(pfackeldey/checkpointing) ✗ ls haha/hihi
tmp.coffea
```
`haha/hihi` didn't exist before of course 🤣.

alexander-held:
This is great to see, thanks for the work on that feature!

> It is not clear to me how to checkpoint a tree-reduction, which is why this is not implemented here.

Having chatted a bit on Slack now, I would view this PR as one very useful feature in a set of a handful of potential API changes / additions. I think there is no need to overcomplicate this feature, especially not in a first iteration. I think it could be complementary to some kind of `allow_read_errors_with_report` mode along the lines of #1393, but this could be modular functionality that users can employ as needed.

lgray:
Are we ready to merge this one?

ikrommyd:
> Are we ready to merge this one?

Well waiting for your review and I also wanted to see if I can ask Martin whether I added something "bad" when it comes to fsspec usage here. I had planned merging by EOD tomorrow and then releasing.

tests/test_checkpointing.py
```diff
@@ -0,0 +1,93 @@
...
+        "Data": {
+            "treename": "Events",
+            "files": [osp.abspath("tests/samples/nano_dimuon.root")],
+        },
+    }
+
+    executor = processor.IterativeExecutor()
+
+    checkpoint_dir = str(Path(__file__).parent / "test_checkpointing")
+    # checkpoint_dir = "root://cmseos.fnal.gov//store/user/ikrommyd/test"
```
ikrommyd, line 49: @pfackeldey If I use this path (which I will leave here) the test passes so I think I'm happy. The only thing that fails in the test is the last cleanup step `fs.rm(paths[0], recursive=True)` with
```
OSError: File not removed properly: [ERROR] Server responded with an error: [3016] Unable to remove /store/user/ikrommyd/test/ZJets/a9490124-3648-11ea-89e9-f5b55c90beef/Events; Is a directory
```
This is unrelated to this PR and maybe an fsspec xrootd bug or an xrootd feature. It seems like it will delete the files but cannot delete the directories recursively. Anyways, something that doesn't affect coffea or this PR.

ikrommyd (approved):
I'm happy with the current state and the test passes locally and remotely over xrootd (apart from the test output cleanup which has nothing to do with coffea or this PR).

PR #1421, opened 2025-09-25
fix: enforce some kwargs for high-level API

### LINKED ISSUE #1405: Enforce kwargs in `NanoEventsFactory.from_X`
Pursuant to what @ikrommyd noticed in https://github.com/scikit-hep/coffea/pull/1387 we should enforce a bunch of parameters as kwargs rather than positional arguments.

pfackeldey:
Closes https://github.com/scikit-hep/coffea/issues/1405.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -250,6 +250,7 @@ class NanoEventsFactory:
     def from_root(
         cls,
         file,
+        *,
         treepath=uproot._util.unset,
         entry_start=None,
         entry_stop=None,
@@ -469,6 +470,7 @@ class NanoEventsFactory:
     def from_parquet(
         cls,
         file,
+        *,
         entry_start=None,
         entry_stop=None,
         runtime_cache=None,
@@ -631,6 +633,7 @@ class NanoEventsFactory:
     def from_preloaded(
         cls,
         array_source,
+        *,
         entry_start=None,
         entry_stop=None,
         runtime_cache=None,
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -287,6 +287,9 @@ class NanoEventsFactory:
                 Stop at this entry offset in the tree (default end of tree)
             steps_per_file: int, optional
                 Partition files into this many steps (previously "chunks")
+            preload (None or Callable):
+                A function to call to preload specific branches/columns in bulk. Only works in eager and virtual mode.
+                Passed to ``tree.arrays`` as the ``filter_branch`` argument to filter branches to be preloaded.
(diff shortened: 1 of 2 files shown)


ikrommyd:
Maybe in https://github.com/scikit-hep/coffea/blob/e997042ce47854384d3017721174fae94b76fb6a/src/coffea/processor/executor.py#L1481-L1489 and https://github.com/scikit-hep/coffea/blob/e997042ce47854384d3017721174fae94b76fb6a/src/coffea/processor/executor.py#L1556-L1564 too?

ikrommyd:
Ughh...that may be breaking to some people though.....

ikrommyd:
The changes in the executors may be breaking to some people but I really don't like args there at all.

ikrommyd:
So, I'm convinced that we should use kwargs even without telling anyone in the `from_*` methods because the first args were useless enough that everyone is using kwargs already anyways.

I'm not sure about the executors though....Quite a few people use args there and I don't know if it's worth breaking their code. I think it's a very common pattern and changing the API like that with an announcement is probably not enough for this.

ikrommyd:
@lgray what do you think?

lgray:
We shouldn't break their code out of nowhere. 

I propose the following: leave current function signature with args but allow kwargs versions of them in case, if args are detected emit a warning that args are going away.

ikrommyd:
> We shouldn't break their code out of nowhere.
> 
> I propose the following: leave current function signature with args but allow kwargs versions of them in case, if args are detected emit a warning that args are going away.

Agreed. For nanoevents `from_X`, let's just do kwargs now I think. The signature is
```py
    def from_root(
        cls,
        file,
        treepath=uproot._util.unset,
        entry_start=None,
        entry_stop=None,
        steps_per_file=uproot._util.unset,
        ...
```
The first few args are pretty much very rarely used so everyone is doing kwargs there anyways.

For the executor methods, too many people are doing args from what I've seen that we shouldn't break it in one day.

lgray:
Sounds fine to me, please go ahead.

ikrommyd:
> Sounds fine to me, please go ahead.

I'm not sure of the best way to do this though since we want the first two to be args and then everything else kwargs afterwards and raise a deprecation warning if the user is passing in args. Something like this?
```py
def foo(self, fileset, processor_instance, *args, treename=None, uproot_options={}, iteritems_options={}):
    if args:
        warnings.warn("Use keyword arguments instead of positional", DeprecationWarning, stacklevel=2)
        if len(args) > 0: treename = args[0]
        if len(args) > 1: uproot_options = args[1] 
        if len(args) > 2: iteritems_options = args[2]
```

lgray:
Yeah, that's about right to me.

PR #1422, opened 2025-09-25
fix: use consistent rich console

pfackeldey:
Using a common `rich.console.Console` for logging/printing in e.g. a Processor ensures that the progress bar and the logging of the Runner (with e.g. a IterativeExecutor) is properly displayed.

Users should import `from coffea.util import coffea_console` and use `coffea_console.print(...)` when they want to use a rich Console to log/print something.

PS: If someone wanted to debug interactively, and had issues with embedded IPython shells (or pdb) and `rich` progress bars, have a look at this: https://github.com/Textualize/rich/issues/1053#issuecomment-1970920245 (@felixzinn).

### DIFF
--- a/binder/dataset_discovery.ipynb
+++ b/binder/dataset_discovery.ipynb
@@ -25,14 +25,14 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "09103c77-b8e6-4d61-920b-b1ff8fba8791",
    "metadata": {},
    "outputs": [],
    "source": [
     "from coffea.dataset_tools import rucio_utils\n",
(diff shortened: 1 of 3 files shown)


ikrommyd (approved):

andrzejnovak:
@pfackeldey just checking, what was "properly displayed" supposed to behave like? I was rather hoping this would avoid the "reprints" when trying to show any other logs, but apart from rich pretty printing the text it doesn't seem to help formatting

image

pfackeldey:
Hm, that's weird @andrzejnovak . Using the same rich console should avoid this reprinting... I did confirm this back then locally when I submitted the PR, I'm not sure why this happens?

PR #1425, opened 2025-09-29
fix: do not reverse fileset after preprocessing

ikrommyd:
I don't understand the need for this. Maybe someone expects the first datasets of their fileset to be submitted for execution first.

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1544,10 +1544,6 @@ class Runner:
 
             self._preprocess_fileset_root(fileset)
             fileset = self._filter_badfiles(fileset)
-
-            # reverse fileset list to match the order of files as presented in version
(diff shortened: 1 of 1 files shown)


lgray:
Waaaay back in the day it was returning a fileset that was backwards compared to the coffea 0.7 fileset. Guess it's not needed any more to pass the tests.

lgray:
right, which was causing maxchunks to yield different results since order matters for slicing like that.

lgray:
anyway if it's not needed it's not needed.

ikrommyd:
Yes what I noticed was that everything was reversed. If I had two datasets A and B in my fileset, I would get the output of B first and then A and also max chunks would start counting from the last file of each dataset which is a bit counter intuitive. Maxchunks 3 should mean the first 3 chunks of  each dataset starting from the first file. It would start from the last file before this PR.

PR #1426, opened 2025-09-30
feat: dask rich progressbar

pfackeldey:
This adds a rich progress bar to DaskExecutors (instead of the text-based one that exists from dask distributed).

### DIFF
--- a/src/coffea/processor/_dask.py
+++ b/src/coffea/processor/_dask.py
@@ -55,6 +55,7 @@ class RichProgressBar(ProgressBar):
 
         if not remaining:
             self.pbar.update(self.task, total=all, completed=all)
+            self.pbar.stop()
 
     def _draw_bar(self, remaining, all, **kwargs):
         del kwargs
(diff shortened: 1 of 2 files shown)


ikrommyd:
@pfackeldey can you please test that it works fine in shell and in jupyter too? You can even make a coffea processor that has like a 10% chance of raising an error and increase the dask retries in the config to 20 or something to see if that causes trouble with the progress bar.

pfackeldey:
This is how it looks in the terminal:

https://github.com/user-attachments/assets/90ba350e-93c0-4d2e-b508-edbf42c74d8a

I'm following up with the notebook one soon...

ikrommyd:
Can we have the description consistent? It's set in various places around https://github.com/scikit-hep/coffea/blob/e5caeb588c1fc832deb9939a9de6817e3611bed4/src/coffea/processor/executor.py. The defaults are Preprocessing and Processing.

pfackeldey:
This is in a notebook (using VSCode):


https://github.com/user-attachments/assets/72d6d105-1989-4721-84e0-937e7933292e

ikrommyd:
I think this looks good. And what about failures? Are the videos with random failures?

ikrommyd (approved):
Looks good and appears to be working.

PR #1428, opened 2025-09-30
feat: add saving/loading for remote filesystems + fix pickling infinite recursion problems

ikrommyd:
This adds support for saving/loading using fsspec and also fixes the pickling problems introduced in https://github.com/scikit-hep/coffea/pull/1338.
Also extends the pickling tests.

### DIFF
--- a/src/coffea/util.py
+++ b/src/coffea/util.py
@@ -33,43 +33,23 @@ import warnings
 from functools import partial
 
 import cloudpickle
-import lz4.frame
+import fsspec
 
 
 def load(filename):
     """Load a coffea file from disk"""
-    with lz4.frame.open(filename) as fin:
+    with fsspec.open(filename, "rb", compression="lz4") as fin:
         output = cloudpickle.load(fin)
     return output
 
 
-def save(output, filename, fast=True):
+def save(output, filename):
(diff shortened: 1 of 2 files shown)


ikrommyd:
We did some tests here by using memray to profile dumping a 4GB array.
So, my understanding is that you always need to make at least one copy. That "best case" scenario occurs when you use `dump` where my memory would go up to 8GB.
If you dump to a string first using `dumps` and then to file, you make two copies and my memory would spike at 12 GB.
So we don't wanna use `dumps` to use `pickletools.optimize` to avoid the larger memory spike of copying twice.
At the same time we profiled with compression. Compression always needs more memory than just one copy. My memory would go up to 12 GB like copying twice when using lz4 compression for instance. However, that enables much smaller output file sizes so we wanna keep that in. We decided though to expose a `compression` argument to the user so that they can use any fsspec supported compression algorithm they want or no compression at all.

Finally, regarding memory, the `fast` argument that we are removing is as optimal as `dump` (one copy). Yes `fast` may result in a more optimal pickle file but it's not worth it to use a deprecated argument for it nor to use the `pickletools.optimize` to actually optimize the pickle because that would need `dumps` and therefore two copies.

PR #1431, opened 2025-10-09
feat: make `postprocess` optional in `ProcessorABC` and additionally accept single-argument callables in executors in the place of processors

### LINKED ISSUE #1383: Making `postprocess()` optional?
Now that the processor interface seems to be replacing the dask-awkard approach, I wonder if it could be streamlined a bit: `postprocess` is required at the moment, but
```python
    pass
```
seems to be a sensible choice already (which is in the ABC). Could this become optional or is there another reason to require it?

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1383

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1400,7 +1400,7 @@ class Runner:
         use_dataframes: bool,
         savemetrics: bool,
         item: WorkItem,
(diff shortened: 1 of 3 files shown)


lgray (approved):
changes are all in line with what we discussed, no backwards compatibility issues.

PR #1432, opened 2025-10-09
feat: add object systematic that varies multiple fields

ikrommyd:
Adds a new systematic kind called `UpDownMultiSystematic` which varies multiple fields at the same time.
The varying function in that case should accept a record array that has the unvaried fields and return a record array that has the varied fields. Each field should access a two-column array of the up and down variations of itself. Like so:
```py
def muon_pt_phi_systematic(ptphi):
    pt_var = (1.0 + np.array([0.05, -0.05], dtype=np.float32)) * ptphi.pt[:, None]
    phi_var = (1.0 + np.array([0.1, -0.1], dtype=np.float32)) * ptphi.phi[:, None]
    return ak.zip({"pt": pt_var, "phi": phi_var}, depth_limit=1)
```


lgray:
Could you add an example of using this to the notebook in `binder/`?

ikrommyd:
> Could you add an example of using this to the notebook in `binder/`?

Done

ikrommyd:
@lgray One final thing I'd like to discuss is the API for the varying function definition. When we vary a single field, the varying function is expected to take in an array of flattened numbers and return an array of two-columns with the up-down variations like so
```py
    def muon_pt_scale(pt):
        return (1.0 + np.array([0.05, -0.05], dtype=np.float32)) * pt[:, None]
```
We use this since coffea 0.7 so it's good to keep this.
Regarding this PR though, what should the varying function API look like in your opinion? I currently have it like so
```py
    def muon_pt_phi_systematic(ptphi):
        pt_var = (1.0 + np.array([0.05, -0.05], dtype=np.float32)) * ptphi.pt[:, None]
        phi_var = (1.0 + np.array([0.1, -0.1], dtype=np.float32)) * ptphi.phi[:, None]
        return ak.zip({"pt": pt_var, "phi": phi_var}, depth_limit=1)
```
The input is a record array that has the unvaried fields like `ptphi.pt` and `ptphi.phi`, and it returns an `ak.zip` of the varied fields as two-column arrays each. Do you think it should be like this record array -> record array (it feels a bit against the single field variation case to me where we never have record arrays)?

We could also have the API like this without any zipping required from the user:
```py
    def muon_pt_phi_systematic(pt, phi):
        pt_var = (1.0 + np.array([0.05, -0.05], dtype=np.float32)) * pt[:, None]
        phi_var = (1.0 + np.array([0.1, -0.1], dtype=np.float32)) * phi[:, None]
        return pt_var, phi_var # or return {"pt": pt_var, "phi": phi_var}
```
How do you think the user should define the varying function in the multi-field case?

lgray:
So this is a very good question. 

A bit of history is that this interface was meant to be a "first try" at a systematics interface and have it guided by analysis use and feedback, but since there wasn't much feedback the interface has stuck.

I would vote to slowly evolve the interface to something that's either more convenient and/or easier to think about.

For this particular case - I would say zipped vs. non-zipped is more a matter of boiler plate, each of them seems to have reasonable clarity for the user in different ways.

ikrommyd:
> So this is a very good question.
> 
> A bit of history is that this interface was meant to be a "first try" at a systematics interface and have it guided by analysis use and feedback, but since there wasn't much feedback the interface has stuck.
> 
> I would vote to slowly evolve the interface to something that's either more convenient and/or easier to think about.

I think for the single field case it's fine to stay array -> array + a lot of people use it (HiggsDNA, PocketCoffea for instance). For the multi-field case, I think that the record array example is more compact but it involves the user knowing they have to zip back. In that sense, just getting in the unvaried fields as args and returning a tuple (or dict) of varried arrays make sense. I don't know what I prefer.....

lgray:
Then we should allow both forms and detect which one.

The only possibly major drawback I see to the multiple-returns one is that it's a bit clunky because you need to assume a mapping between the function args and the output, which can lead to more human error. The record array version at least keeps that possibility of confusion down to a single line of code.

ikrommyd:
> Then we should allow both forms and detect which one.
> 
> The only possibly major drawback I see to the multiple-returns one is that it's a bit clunky because you need to assume a mapping between the function args and the output, which can lead to more human error. The record array version at least keeps that possibility of confusion down to a single line of code.

Yeah accepting args and returning a tuple assumes an order which we need to assume is the same as the `what` argument. With record arrays, everything is accessed using keys so maybe it's best to keep the `zip` thing actually.

ikrommyd:
yeah let's not go down that rapid whole because if the user doesn't follow this pattern (the function args being in the same order as `what`), then they can just get wrong numbers without even knowing. Cause different fields will get varied according to incorrect rules. Let's keep the API as is. Allowing args and assuming order is just extra headache.

PR #1437, opened 2025-10-14
fix: do not simplify layout during `from_buffers` call + explicit kwargs passing to `from_buffers`

ikrommyd:
I do not understand why we had `allow_noncanonical_form=True` here. This means to simplify the layout during construction -> it will use `UnionArray.simplified` instead of `UnionArray` to construct a union array for example. However...this often needs to run awkward cpp simplification kernels, meaning we get materialization of virtual arrays when calling `from_root`. I don't think that is desired. Is there a specific reason we had `allow_noncanonical_form=True` here? cc @lgray @nsmith-


lgray:
I think this one dates back to @nsmith-'s original implementation.

ikrommyd:
> I think this one dates back to @nsmith-'s original implementation.

Not that far back because this option did not exist in awkward1: https://awkward-array.org/doc/1.10/_auto/ak.from_buffers.html
I guess it was a nice thing to do to make the layout more compact when reading eagerly but I'm afraid it's not good when reading virtually (now that that's a feature). We don't wanna run operations when opening a file lazily.

lgray:
OK - then perhaps it was something to just make it work from back in the early days of awkward2/uproot5.

lgray:
In any case, passes tests and reduces reads. Good by me.

lgray (approved):

nsmith-:
It was added by @pfackeldey in https://github.com/scikit-hep/coffea/pull/1277/files#diff-5d15ab5d980fb950986e9de53eefdda2391a413f58314be4b0aa8c9b9040d07eR787

pfackeldey:
> It was added by @pfackeldey in https://github.com/scikit-hep/coffea/pull/1277/files#diff-5d15ab5d980fb950986e9de53eefdda2391a413f58314be4b0aa8c9b9040d07eR787

oops, that must've been a mistake on my side. I don't remember changing that on purpose, sorry!

ikrommyd:
> > It was added by @pfackeldey in https://github.com/scikit-hep/coffea/pull/1277/files#diff-5d15ab5d980fb950986e9de53eefdda2391a413f58314be4b0aa8c9b9040d07eR787
> 
> oops, that must've been a mistake on my side. I don't remember changing that on purpose, sorry!

It didn't cause any problems because we don't have simplifiable layouts straight from nanoaod or something so don't worry. It could have caused unpredicted materialization in other cases  though:)

PR #1444, opened 2025-10-26
chore: add python 3.14 support

ikrommyd:
Tests and adds support for python 3.14.


lgray:
~numba is missing and will be for another couple months.~ Ah 0.63b1 is py314. Nice.

ikrommyd:
> ~numba is missing and will be for another couple months.~ Ah 0.63b1 is py314. Nice.

Yeah there is the pre-release. Not gonna merge of course before we can switch to relelased version.

ikrommyd:
Docs ci is failing with
```
Collecting lz4@ https://github.com/python-lz4/python-lz4.git (from coffea==2025.10.3.dev9+g00607fae4.d20251031)
  Downloading https://github.com/python-lz4/python-lz4.git
     - 281.9 kB 7.0 MB/s 0:00:00
  ERROR: Cannot unpack file /tmp/pip-unpack-5qd6aj3j/python-lz4.git (downloaded from /tmp/pip-install-it5nv86s/lz4_62f8d03976fd48eba6d6df83ee983a7a, content-type: text/html; charset=utf-8); cannot detect archive format
ERROR: Cannot determine archive format of /tmp/pip-install-it5nv86s/lz4_62f8d03976fd48eba6d6df83ee983a7a
```
which shall be fixed with a release on their part with 314 wheels.

ikrommyd:
okay @lgray it looks like this is good. We just need to wait for these packages to make actual releases. Putting in on draft mode until then.

ikrommyd:
`lz4` made a release with 3.14 wheels so it looks like we are only waiting for numba 0.63.0 here which should be out in a couple weeks: https://github.com/numba/numba/issues/9957#issuecomment-3456157382

lgray (approved):

Issue #1094, opened 2024-05-14
Test failing with coffea 2024.5.0


GaetanLepage:
**Describe the bug**

Context: Updating `coffea` in the nixpkgs repo (https://github.com/NixOS/nixpkgs/pull/307316).
We are updating `uproot` at the same time (to 5.3.6).

The following two tests fail in our sandbox:
```
FAILED tests/test_lumi_tools.py::test_lumimask[https://raw.githubusercontent.com/CoffeaTeam/coffea/master/tests/samples/Cert_294927-306462_13TeV_EOY2017ReReco_Collisions17_JSON.txt] - FileNotFoundError: https://r>
FAILED tests/test_lumi_tools.py::test_lumilist_dask - TypeError: ('Could not serialize object of type HighLevelGraph', '<ToPickle...
```

The first failure makes sense, as our sandbox does not have internet access. (From my understanding, this is the only test in the coffea test suite that requires a network access).
However, I don't know if the error in the second one is expected.

**To Reproduce**
1. Run pytest
2. See error

**Expected behavior**
The tests don't fail.

**Output**
<details>

```
>               result = pickle.dumps(x, **dump_kwargs)
E               AttributeError: Can't pickle local object 'unpack_collections.<locals>.repack'
....
                raise TypeError(msg) from exc
>           raise TypeError(msg, str_x) from exc
E           TypeError: ('Could not serialize object of type HighLevelGraph', '<ToPickle: HighLevelGraph with 20 layers.\n<dask.highlevelgraph.HighLevelGraph object at 0x7ffc468a18d0>\n 0. Dict-ee30ab14-73e1-45
```

</details>

**Desktop (please complete the following information):**
 - OS: NixOS
 - Version 24.05

GaetanLepage:
Full log: https://pastebin.com/YPdvMUJ6

ikrommyd:
The second is caused by the first one failing. It happens on my machine as well If I change `https://raw.githubusercontent.com/CoffeaTeam/coffea/master/tests/samples/Cert_294927-306462_13TeV_EOY2017ReReco_Collisions17_JSON.txt"` to something that does not exist.
If you run the second test individually it should succeed.
I'm seeing `aiohttp` errors in the log like `aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host raw.githubusercontent.com:443 ssl:default [Temporary failure in name resolution]` as it's trying to read the remote file over http. What I don't see at the moment is why.

[...]

PR #1454, opened 2025-11-02
test: add network mark to lumi mask test

### LINKED ISSUE #1456: test: mark all tests requiring network connectivity
Use a custom marker `@pytest.mark.network` to indicate such tests, which can then be filtered with `pytest -m "not network"`

Related to #1094

ikrommyd:


nsmith-:
Is that the only one that downloads data?

ikrommyd:
> Is that the only one that downloads data?

This is the only thing that would fail in the original issue https://github.com/scikit-hep/coffea/issues/1094 in the ci that had no internet access at least.
Plus this is the only valid "http" instance I found in the tests via grep.

nsmith- (approved):

PR #1466, opened 2025-11-04
feat: implement transforms that do zeros like of branches at the schema level

ikrommyd:
Implements transforms to be able to ~add branch aliases at the schema level~ and also add "zeros like" fields whenever something is missing (like Photon mass and charge in nanoaod).

Edit: Branch aliases at the schema level are actually just a extra key in the dictionary like `branch_forms["Photon_PT"] = branch_forms["Photon_pt"]`. No reason to invent a whole transformation for that.


ikrommyd:
@nsmith- as discussed, I was inventing something stupid with the alias form because it's not required so I removed it. If you're happy with the rest of the transforms now, we can merge this one.

src/coffea/nanoevents/transforms.py
```diff
@@ -487,6 +487,58 @@ def eventindex(stack):
...
+def zeros_from_content_form(source_form):
+    form = copy.deepcopy(source_form)
+    form["form_key"] = concat(source_form["form_key"], "!zeros_from_content")
+    if not (form["class"] == "NumpyArray" or form["class"].startswith("ListOffset")):
+        raise RuntimeError
+    if form["class"].startswith("ListOffset"):
+        form["content"]["form_key"] = concat(
+            source_form["form_key"], "!zeros_from_content", "!content"
+        )
+    return form
```
nsmith-: We shouldn't have to change the offsets form key when it is jagged, no?
```python
if form["class"] == "NumpyArray":
    form["form_key"] = concat(source_form["form_key"], "!zeros_from_content")
    return form
elif form["class"].startswith("ListOffset"):
    form["content"]["form_key"] = concat(
        source_form["form_key"], "!zeros_from_content", "!content"
    )
    return form
raise ValueError("Unsupported source form")

File: src/coffea/nanoevents/transforms.py [hunk 2]
```diff
@@ -487,6 +487,58 @@ def eventindex(stack):
...
+            "class": "NumpyArray",
+            "primitive": "float32",
+            "itemsize": 8,
+            "format": "i",
+            "form_key": concat(
+                offsets_form["form_key"], "!zeros_from_offsets", "!content"
+            ),
+        },
+        "form_key": concat(offsets_form["form_key"], "!zeros_from_offsets"),
+        "parameters": offsets_form.get("parameters", None),
```
nsmith-: Is this going to copy the `__doc__` parameter from offsets? This might not be desirable.

src/coffea/nanoevents/transforms.py
```diff
@@ -487,6 +487,58 @@ def eventindex(stack):
...
+    if not offsets_form["class"].startswith("NumpyArray"):
+        raise RuntimeError
+
+    form = {
+        "class": "ListOffsetArray",
+        "offsets": "i64",
+        "content": {
+            "class": "NumpyArray",
+            "primitive": "float32",
+            "itemsize": 8,
```
nsmith-: float32 should be size 4 I think.
Surprised awkward does not yell about this

nsmith- (changes requested):

src/coffea/nanoevents/transforms.py
ikrommyd: Yeah that's a typo. Forgot when I changed to float32. I don't think `ak.from_buffers` cares about this. I think it;s getting the itemsize from the dtype either way but let's have the correct thing here.

src/coffea/nanoevents/transforms.py
ikrommyd: Looks like no?
```py
In [15]: events.Photon["pt"].layout.content.parameters
Out[15]: {'__doc__': 'pt', 'typename': 'float[]'}

In [16]: events.Photon["mass"].layout.content.parameters
Out[16]: {}
```

ikrommyd: But we may wanna remove it right?

ikrommyd: I just deleted this line.

src/coffea/nanoevents/transforms.py
ikrommyd: I changed it to this now which is more or less what you wrote. I'm also popping the `__doc__` parameter actually now because otherwise these zeros fields will have the original doc of what was used to create them. Let me know if you see anything else here.
```py
def zeros_from_content_form(source_form):
    form = copy.deepcopy(source_form)
    if not (form["class"] == "NumpyArray" or form["class"].startswith("ListOffset")):
        raise RuntimeError
    if form["class"] == "NumpyArray":
        form["form_key"] = concat(source_form["form_key"], "!zeros_from_content")
        form["parameters"].pop("__doc__", None)
    elif form["class"].startswith("ListOffset"):
        form["content"]["form_key"] = concat(
            source_form["form_key"], "!zeros_from_content", "!content"
        )
        form["parameters"].pop("__doc__", None)
        form["content"]["parameters"].pop("__doc__", None)
    return form
```

nsmith- (approved):

Issue #1474, opened 2025-11-11
Ensure the right fields are present at the schema level for all schemas


ikrommyd:
scikit-hep/vector expects fields to be named in a very specific way in order to make objects behave like 4-vectors. It also looks at fields only and not behaviors. Defining the following behavior for example like we do for Delphes does not help it. Vector only looks inside fields and not behaviors.
```py
    @property
    def pt(self):
        return self["PT"]
```
We should therefore ensure that all the fields are aliased properly at the schema level. In this example, the delphes schema should alias PT -> pt at the schema level.

PR #1480, opened 2025-11-17
feat: add CorrT1METJet to NanoAOD schema and also add zero TrigObj mass

### LINKED ISSUE #1119: Adding the CorrT1METJet class to nanoaod.py
When propagating JEC to Type1 corrections for raw MET, it is necessary to consider the contributions of both jet and soft jets, but currently the script does not support the use of soft jets branch CorrT1METJet.

Fix the /coffea/nanoevents/methods/nanoaod.py, add these code for CorrT1METJet:
```
@awkward.mixin_class(behavior)
class CorrT1METJet(vector.PtEtaPhiMLorentzVector, base.NanoCollection, base.Systematic):
    """NanoAOD narrow radius jet object"""
    @property
    def pt(self):
        return self["rawPt"]

    @property
    def rawFactor(self):
        self["rawFactor"] = awkward.zeros_like(self["rawPt"])
        return self["rawFactor"]

    @property
    def mass(self):
        self["mass"] = awkward.zeros_like(self["rawPt"])
        return self["mass"]

    @property
    def chEmEF(self):
        self["chEmEF"] = awkward.zeros_like(self["rawPt"])
        return self["chEmEF"]

    @property
    def neEmEF(self):
        self["neEmEF"] = awkward.zeros_like(self["rawPt"])
        return self["neEmEF"]


_set_repr_name("CorrT1METJet")
```

Fix the /coffea/nanoevents/schemas/nanoaod.py, add this key in 'mixins' dictionary:

`"CorrT1METJet":"CorrT1METJet",`

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1119


ikrommyd:
@nsmith- Is it safe to add zero mass to the trigobj 4-vector to allow 4-vector math? Also `CorrT1METJet` now has a `rawMass` field in nano v15 (not in previous versions) which is definitely not zero: https://cms-xpog.docs.cern.ch/autoDoc/NanoAODv15/2024/doc_TTH-Hto2G_Par-M-125_TuneCP5_13p6TeV_amcatnloFXFX-pythia8_RunIII2024Summer24NanoAODv15-150X_mcRun3_2024_realistic_v2-v2.html#CorrT1METJet
image
Should we use that or zero out the mass? 
It looks like this
```
In [16]: events.CorrT1METJet.rawMass.show()
[[2.09, -1.29e-06, 2.11],
 [0.965],
 [1.84],
 [2.74, 1.99, 1.28e-06, 1.78],
 [1.37e-06],
 [2.05, 0.932, 1.35],
 [2.47, 0.841, 0.898, -1.34e-05],
 [],
 [0.606],
 [2.82],
 ...,
 [3.24, 1.26e-06, 2.71, -6.98e-06],
 [2.56, 2.59, 1.99],
 [3.04, -1.01e-06],
 [1.89, 1.71, 1.85, 1.65, 2.15],
 [5.72e-06],
 [0.825, 1.79],
 [1.92, 2.26, 2.59],
 [2.17, 1.6],
 [-9.54e-07, 1.59, 1.72]]
 ```

File: src/coffea/nanoevents/schemas/nanoaod.py [hunk 1]
```diff
@@ -314,6 +315,22 @@ def _build_collections(self, field_names, input_contents):
...
+                branch_forms["CorrT1METJet_mass"] = transforms.zeros_from_offsets_form(
+                    branch_forms["oCorrT1METJet"]
+                )
+
+        # Add zero mass to the trigger objects
+        if "oTrigObj" in branch_forms:
+            if "TrigObj_mass" not in branch_forms:
+                branch_forms["TrigObj_mass"] = transforms.zeros_from_offsets_form(
+                    branch_forms["oTrigObj"]
+                )
```
ikrommyd: All these explicit stuff will be refactored in the form of class variable dictionaries similar to how we define all the other mixins. It will be done as part of closing https://github.com/scikit-hep/coffea/issues/1474. I'm just now doing it in the same pattern as everything else.

ikrommyd:
I don't understand this flaky failure that just showed up......

ikrommyd:
Yeah this currently works but I had to add a big chunk of slightly modified code that I stole from awkward array in order to keep tracing functioning properly....This functionality should probably be in awkward and not here.

ikrommyd:
@nsmith- I found a way to keep the tracing intact by using some awkward internals but without copying over any code. This is  ready I think from my side so can you check my question above only?

nsmith-:
> Is it safe to add zero mass to the trigobj 4-vector to allow 4-vector math? Also CorrT1METJet now has a rawMass field in nano v15 (not in previous versions) which is definitely not zero ...  Should we use that or zero out the mass?

On the first point, I think its probably OK. For CorrT1METJet, I have no idea. I've asked JME conveners let me ping again here @mcremone @cesarecazzaniga

src/coffea/nanoevents/schemas/nanoaod.py
```diff
@@ -314,6 +315,22 @@ def _build_collections(self, field_names, input_contents):
...
+            if "CorrT1METJet_pt" not in branch_forms:
+                branch_forms["CorrT1METJet_pt"] = branch_forms["CorrT1METJet_rawPt"]
+            if "CorrT1METJet_mass" not in branch_forms:
+                branch_forms["CorrT1METJet_mass"] = transforms.zeros_from_offsets_form(
+                    branch_forms["oCorrT1METJet"]
+                )
+
+        # Add zero mass to the trigger objects
+        if "oTrigObj" in branch_forms:
+            if "TrigObj_mass" not in branch_forms:
```
nsmith-: I'm not sure the right solution here, but one eventuality we might have is that `TrigObj_mass` is added to NanoAOD and it is not with the same definition we have here. Then coffea behavior silently changes. There's probably several ways in which a new NanoAOD version silently changes the NanoEvents behavior, so its not unique to this case.

src/coffea/nanoevents/trace.py
```diff
@@ -36,30 +36,39 @@ def _make_length_zero_one_tracer(
...
 
+    def getkey(layout, form, attribute):
+        return buffer_key(
+            form_key=form.form_key,
+            attribute=attribute,
+            form=form,
+        )
+
+    container = {}
+    tmp_array._to_buffers(
```
nsmith-, line 56: We should try to avoid calling private methods in awkward. But if it is "somewhat stable" this is fine.

nsmith- (approved):
I don't have any actionable comments, but this also touches typetracer stuff that I am not super familiar with.

src/coffea/nanoevents/trace.py
ikrommyd: It is the only way I think actually (other than copying over code). It's a layout method so I think it's fine and there aren't really many layout methods that are public (no underscore start). Tracing is inherently using awkward internals in any case.

src/coffea/nanoevents/schemas/nanoaod.py
ikrommyd: Would you rather enforce it and remove the `if "TrigObj_mass" not in branch_forms` check? This keeps it consistent if a "mass" pops up in some nanoaod version until we manually have to change it.

nsmith-: Always rewriting with our definition and emitting a warning if we do see the conflict is probably a good solution

src/coffea/nanoevents/trace.py
nsmith-: Now that you're an awkward maintainer, I suppose you approve of this use :)

cesarecazzaniga:
> > Is it safe to add zero mass to the trigobj 4-vector to allow 4-vector math? Also CorrT1METJet now has a rawMass field in nano v15 (not in previous versions) which is definitely not zero ...  Should we use that or zero out the mass?
> 
> On the first point, I think its probably OK. For CorrT1METJet, I have no idea. I've asked JME conveners let me ping again here @mcremone @cesarecazzaniga

I think it is better to keep non-zero mass for CorrT1METJet, these should be additional low-pt jets for Type-1 MET re-correction, and I am not sure it is safe to set the mass to zero. What do you think @mcremone ?

ikrommyd:
I don’t really know what this “raw mass” means and whether it is the right temporal coordinate of the 4-vector. It’s also nano v15+ only so for older versions we definitely need to zero it out. Let me know.

cesarecazzaniga:
> I don’t really know what this “raw mass” means and whether it is the right temporal coordinate of the 4-vector. It’s also nano v15+ only so for older versions we definitely need to zero it out. Let me know.

Should be just the mass of the low pT jets that are used to recompute the MET-T1 correction. I think in principle should have a mass which is not zero in the (pT, eta, phi, mass) representation. Looking a bit the code here (https://github.com/nurfikri89/NanoSkimmer/blob/1b4db934993267761710ab2401caf43d7a19d710/modules/AddJEC_ForNanoValidation.C#L398) seems that the rawMass is not used to compute the jec, but it is used later on (https://github.com/nurfikri89/NanoSkimmer/blob/1b4db934993267761710ab2401caf43d7a19d710/modules/AddJEC_ForNanoValidation.C#L416).

But it is true that for older versions they set it to zero. So, for sure for older versions I think zeroing it out is fine.

ikrommyd:
> > I don’t really know what this “raw mass” means and whether it is the right temporal coordinate of the 4-vector. It’s also nano v15+ only so for older versions we definitely need to zero it out. Let me know.
> 
> Should be just the mass of the low pT jets that are used to recompute the MET-T1 correction. I think in principle should have a mass which is not zero in the (pT, eta, phi, mass) representation. Looking a bit the code here (https://github.com/nurfikri89/NanoSkimmer/blob/1b4db934993267761710ab2401caf43d7a19d710/modules/AddJEC_ForNanoValidation.C#L398) seems that the rawMass is not used to compute the jec, but it is used later on (https://github.com/nurfikri89/NanoSkimmer/blob/1b4db934993267761710ab2401caf43d7a19d710/modules/AddJEC_ForNanoValidation.C#L416).
> 
> But it is true that for older versions they set it to zero. So, for sure for older versions I think zeroing it out is fine.

Yes but that means that if you have a v13 sample, that you then reprocess with v15, then your four-vector math change slightly depending on the nano version...I'm not super comfortable with this. I guess `CorrT1METJet` is a rare collection to use but still....

cesarecazzaniga:
Hi

> > > I don’t really know what this “raw mass” means and whether it is the right temporal coordinate of the 4-vector. It’s also nano v15+ only so for older versions we definitely need to zero it out. Let me know.
> > 
> > 
> > Should be just the mass of the low pT jets that are used to recompute the MET-T1 correction. I think in principle should have a mass which is not zero in the (pT, eta, phi, mass) representation. Looking a bit the code here (https://github.com/nurfikri89/NanoSkimmer/blob/1b4db934993267761710ab2401caf43d7a19d710/modules/AddJEC_ForNanoValidation.C#L398) seems that the rawMass is not used to compute the jec, but it is used later on (https://github.com/nurfikri89/NanoSkimmer/blob/1b4db934993267761710ab2401caf43d7a19d710/modules/AddJEC_ForNanoValidation.C#L416).
> > But it is true that for older versions they set it to zero. So, for sure for older versions I think zeroing it out is fine.
> 
> Yes but that means that if you have a v13 sample, that you then reprocess with v15, then your four-vector math change slightly depending on the nano version...I'm not super comfortable with this. I guess `CorrT1METJet` is a rare collection to use but still....

Hi, I pinged the JME L2, i think he can reply directly here and suggest the best way to proceed.

nurfikri89:
Hi, purely from Type-1 MET correction perspective, all we care about is the `px` and `py` of each jet, which we need just `pt` and `phi`. `eta` is needed because we want to retrieve jet energy corrections (binned in jet `eta`). `rawMass` was included so that one can use the four-vector of the jets in this collection.

ikrommyd:
> Hi, purely from Type-1 MET correction perspective, all we care about is the `px` and `py` of each jet, which we need just `pt` and `phi`. `eta` is needed because we want to retrieve jet energy corrections (binned in jet `eta`). `rawMass` was included so that one can use the four-vector of the jets in this collection.

Okay so raw mass seems to be the correct temporal coordinate of the 4-vector. And if that does not exist, since it’s a v15+ feature, should we zero it out or not allow 4-vector math that involve mass cause those would be wrong?

ikrommyd:
Okay, I think the right thing to do here is only alias rawMass -> mass and enable the temporal coordinate only if rawMass is present (post v15) and in pre v15, we just let vector error with
```
ValueError: array does not have temporal coordinates (t/E/e/energy or tau/M/m/mass): EmEF, area, eta, muonSubtrFactor, phi, rawPt, pt
```
and the user can choose to manually add a zero mass if they want instead of us doing it for them.

Issue #1473, opened 2025-11-11
Add behavior validation hook to ensure that all the right fields are present


ikrommyd:
Using https://github.com/scikit-hep/awkward/pull/3710, we should add behavior validation hooks to all particle objects to ensure that all the necessary fields are present at the schema level that scikit-hep/vector expects.
This is not currently true for all schemas so updating the schemas should be done first.

PR #1481, opened 2025-11-17
feat: ensure the right fields are present at the schema level to satisfy 4-vector behaviors

### LINKED ISSUE #1474: Ensure the right fields are present at the schema level for all schemas

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1474


tests/test_nanoevents_vector.py
```diff
@@ -680,6 +678,8 @@ def test_photon_zero_mass_charge(optimization_enabled):
                 - np.cos(daskdiphotons.tag.phi - daskdiphotons.probe.phi)
             )
         )
-        assert_eq(eagerdiphotons.mass, eagermll)
-        assert_eq(daskdiphotons.mass, daskmll)
-        assert_eq(eagerdiphotons.mass, daskdiphotons.mass)
+        assert ak.almost_equal(eagerdiphotons["mass"], eagermll, check_parameters=False)
+        assert ak.almost_equal(
+            daskdiphotons["mass"].compute(), daskmll.compute(), check_parameters=False
+        )
```
ikrommyd, line 684: I don't understand how moving things into dicts in the schema instead of explicit code causes awkward to drop parameters here

ikrommyd:
This is almost ready but there's a few things we can't directly covert at the schema level by aliases/zeros without implementing new transforms. For example
```python
@awkward.mixin_class(behavior)
class TrackParticle(vector.LorentzVector, base.NanoCollection):
    """Collection of track particles, following `xAOD::TrackParticle_v1
    <https://gitlab.cern.ch/atlas/athena/-/blob/21.2/Event/xAOD/xAODTracking/Root/TrackParticle_v1.cxx#L82>`_.
    """

    @property
    def p(self):
        return 1.0 / numpy.abs(self.qOverP)

    @property
    def x(self):
        return self.p * numpy.sin(self.theta) * numpy.cos(self.phi)

    @property
    def y(self):
        return self.p * numpy.sin(self.theta) * numpy.sin(self.phi)

    @property
    def z(self):
        return self.p * numpy.cos(self.theta)

    @property
    def t(self):
        return numpy.sqrt(139.570**2 + self.x**2 + self.y**2 + self.z**2)
 ```
 We'd need to add new transforms to do all this math at the schema level for example. Any takes here @nsmith-? The other option would just be to not expect such objects to behave like 4-vectors but I don't think that's good.

**Comment by nsmith-** [comment]
How about using the `rho,phi,theta,tau` [coordinate system](https://vector.readthedocs.io/en/latest/index.html#coordinate-systems) ? Then we just implement `tau` as full with the pion mass using a full_like transform similar to the zeros_like in #1466, and make a transform for `rho = abs(1/qOverP)`?

**Comment by ikrommyd** [comment]
> How about using the `rho,phi,theta,tau` [coordinate system](https://vector.readthedocs.io/en/latest/index.html#coordinate-systems) ? Then we just implement `tau` as full with the pion mass using a full_like transform similar to the zeros_like in #1466, and make a transform for `rho = abs(1/qOverP)`?

I suppose that's good yes. We have phi and theta available as branches. We upgrade zeros_like to full_like (with an argument for the value) and then `zeros_like` becomes `full_like(form, 0)` and we just implement one extra schema-specific transform for `rho`

And do you have anything smarter than two transforms for this. This is the only other missing behavior I see from schemas (we may discover more in the future though).
```py
@awkward.mixin_class(behavior)
class MissingET(vector.SphericalThreeVector, base.NanoCollection):
    @property
    def rho(self):
        return self["MET"] * numpy.cosh(self.eta)

    @property
    def theta(self):
        return 2 * numpy.arctan(numpy.exp(-self.eta))
```
For this one we have met, eta and phi present as branches.

ikrommyd:
@nsmith- the problem with `full_like` is that these transforms are "static" They take in the `stack` as input only and do something. You can't pass a "fill value" argument to them as the only argument has to be the `stack.` Therefore we have to hardcode a full like form with the pion mass for this case. I'm fine with that as I don't want to change the whole logic to be able to pass in arguments just for this single use case.

nsmith-:
Just pop the fill value from the stack and cast it from string to float.

ikrommyd:
I was under the impression that the stack is only filled with arrays that come from certain form keys. Am I mistaken? Is there any chance you can show me a tiny snippet of what you mean?

nsmith-:
For nodes that aren't commands (starting with `!`) then their value just gets pushed
https://github.com/scikit-hep/coffea/blob/4e596edcedcdf19893e696a06b1f8dad5818ee72/src/coffea/nanoevents/mapping/base.py#L148-L149
So you can have a string like `phi,!load,139.570,!full_like` and the `!full_like` implementation would be
```python
fill_value, like_array = stack.pop(), stack.pop()
...
```
You can see a similar thing done for `!item` https://github.com/scikit-hep/coffea/blob/4e596edcedcdf19893e696a06b1f8dad5818ee72/src/coffea/nanoevents/transforms.py#L478-L481

ikrommyd:
Thanks! I was not aware of this. I implemented "full like" like this where we pass the fill value as a string in the key and then it gets popped from the stack.

ikrommyd:
From this part in the vector docs
image
It looks we for MET
```py
@awkward.mixin_class(behavior)
class MissingET(vector.SphericalThreeVector, base.NanoCollection):
    @property
    def rho(self):
        return self["MET"] * numpy.cosh(self.eta)

    @property
    def theta(self):
        return 2 * numpy.arctan(numpy.exp(-self.eta))
```
We can get away with one transformation. We do not define `theta` at the schema level since `eta` is already present and we just define one transformation for `rho`. We use the rho-phi-eta system.

ikrommyd:
This should be ready for review. We may find more such cases once we add the behavior validation hooks (https://github.com/scikit-hep/coffea/issues/1473). I'm just fixing the existing ones here and don't want to make it super big.

ikrommyd:
This waits for https://github.com/scikit-hep/coffea/pull/1480 first but is otherwise ready from my side.

src/coffea/nanoevents/methods/delphes.py
```diff
@@ -87,25 +86,21 @@ class ScalarHT(base.NanoCollection): ...
 
 
 @awkward.mixin_class(behavior)
-class MissingET(vector.SphericalThreeVector, base.NanoCollection):
-    @property
-    def rho(self):
-        return self["MET"] * numpy.cosh(self.eta)
+class MissingET(vector.SphericalThreeVector, base.NanoCollection): ...
```
nsmith-, line 94: Delphes MET has an `eta`?? What does it mean? @kratsg any idea?

src/coffea/nanoevents/schemas/nanoaod.py
```diff
@@ -165,6 +166,27 @@ class NanoAODSchema(BaseSchema):
...
+            transforms.full_like_from_offsets_form,
+            ("oCorrT1METJet", 0.0),
+        ),
+    }
+    """Arrays that should be filled with constant values if not present to satisfy 4-vector requirements"""
+    rename_items = {
+        "Electron_energy": "Electron_regrEnergy",
+        "Photon_energy": "Photon_regrEnergy",
+    }
+    """Arrays that should be renamed to ensure proper 4-vector behavior"""
```
nsmith-, line 181: Any reason to rename rather than alias here? The downside being that people might expect to find `events.Electron.regrEnergy`

src/coffea/nanoevents/schemas/physlite.py
nsmith-, line 1: @alexander-held and @gordonwatts, who would be a good contact to cc on physlite changes in coffea?

src/coffea/nanoevents/schemas/nanoaod.py
ikrommyd: Yes, a very strong one. If you keep energy you have pt eta phi mass AND energy as fields and 4-vector chooses one literally depending on the order of which it finds first. And energy is not the "reconstructed energy". It is the energy measured from the regression.

ikrommyd: I spoke with @valsdav about this. It's bad that this was even called `energy` in the first place and it only affects the EGM Nano flavor of NanoAOD. We should rename it.

nsmith-: Ahh ok I read it backwards as renaming regrEnergy -> energy. Now it makes sense!

src/coffea/nanoevents/schemas/physlite.py
ikrommyd: Yeah I think I did not do anything sketchy here as I only moved things around and made the photons 4-vectors too but it would be nice to have someone that knows physlite.

src/coffea/nanoevents/schemas/nanoaod.py
ikrommyd: Ah I see. No the key-value mapping is from-to. And obviously the `.energy` method will still be available that gets you the reconstructed energy.

src/coffea/nanoevents/methods/delphes.py
ikrommyd: Yeah this one has "MET", "Eta" and "Phi" present. I think @kpedro88 also knows something here.

kratsg: It did (I never understood what Delphes was doing).

kpedro88: I asked about this many, many years ago, and was told it is a convenience for theorists. 🤷‍♂️

src/coffea/nanoevents/schemas/physlite.py
kratsg: I'd say at least me, but I'm not sure who else. Maybe Nikolai?

src/coffea/nanoevents/methods/delphes.py
kratsg: I stand corrected, I never understand what ~Delphes~ theorists are doing.

src/coffea/nanoevents/schemas/physlite.py
nsmith-: Thanks @kratsg! I understand Nikolai is no longer in HEP, otherwise he would be the default.

src/coffea/nanoevents/methods/delphes.py
nsmith-: Is it always `0` in the data files? If so, we can assume that and never read it in our functions.

ikrommyd:
```
In [3]: events.MissingET.Eta * 1
Out[3]: <Array [-7.48, 8.03, -7.87, ..., 4.89, 6.25] type='25 * float32[parameters=...'>
```

nsmith-: 🤯 sounds like it is a convenient way for theorists that don't look too carefully at their code to get really good results for final states with neutrinos...
(also young students)

kratsg: perhaps let's add in a comment that refers to this PR in case future people have questions about this line of code?

nsmith- (approved):

src/coffea/nanoevents/schemas/nanoaod.py
```diff
@@ -173,19 +172,16 @@ class NanoAODSchema(BaseSchema):
...
-            transforms.full_like_from_offsets_form,
-            ("oCorrT1METJet", 0.0),
-        ),
     }
     """Arrays that should be filled with constant values if not present to satisfy 4-vector requirements"""
     rename_items = {
-        "Electron_energy": "Electron_regrEnergy",
-        "Photon_energy": "Photon_regrEnergy",
+        "Electron_regrEnergy": "Electron_energy",
+        "Photon_regrEnergy": "Photon_energy",
```
nsmith-, line 179: Ok so now the targets are consistently the keys, thanks!

src/coffea/nanoevents/schemas/nanoaod.py
```diff
@@ -321,39 +317,34 @@ def _build_collections(self, field_names, input_contents):
 
         # Create full-like arrays
         for name, (fcn, args) in self.full_like_items.items():
-            if name not in branch_forms and args[0] in branch_forms:
+            if args[0] in branch_forms:
+                if name in branch_forms:
+                    warnings.warn(
+                        f"Branch {name} already exists but will be replaced with zeros",
```
nsmith-: It might be replaced with something other than zero

nsmith- (approved):

src/coffea/nanoevents/schemas/nanoaod.py
ikrommyd: Yeah I noticed that it was the case in all other dicts we create haha

src/coffea/nanoevents/schemas/nanoaod.py
ikrommyd: Ah yeah even though it is only zeros now, we can get the value from the `args`.

PR #1487, opened 2025-11-20
fix: improve fcc tests branch filtering pattern

ikrommyd:
The FCC file has these branches/subbranches:
```
 'PARAMETERS',
 'PARAMETERS/_intMap',
 'PARAMETERS/_intMap/_intMap.first',
 'PARAMETERS/_intMap/_intMap.second',
 'PARAMETERS/_floatMap',
 'PARAMETERS/_floatMap/_floatMap.first',
 'PARAMETERS/_floatMap/_floatMap.second',
 'PARAMETERS/_stringMap',
 'PARAMETERS/_stringMap/_stringMap.first',
 'PARAMETERS/_stringMap/_stringMap.second'
 ```
When we do `tree.iteritems()` to extract the base form, subbranch names only contain the subbranch part and not the full path. The function sees PARAMETERS first. Then it sees _intMap, then it sees _intMap.first and so on. Therefore, the filtering function `lambda x: "PARAMETERS" not in x` is not enough. I'm updating it to a regex to exclude `_*Map*` patterns too.

### CODE REVIEW

File: tests/test_nanoevents_edm4hep.py [hunk 1]
```diff
@@ -19,14 +19,14 @@ def _events(**kwargs):
 @pytest.fixture(scope="module")
 def eager_events():
     return _events(
-        mode="eager", iteritems_options={"filter_name": lambda x: "PARAMETERS" not in x}
+        mode="eager", iteritems_options={"filter_name": "/^(?!.*(PARAMETERS|_.*Map))/"}
```
ianna, line 22: I think that this is a very specific case. Perhaps, finding a more generic solution would be better?

ianna (commented):
@ikrommyd - thanks for looking into it! It seems to me that this is a temporary solution. Perhaps we should do some more digging? Thanks!

tests/test_nanoevents_edm4hep.py
ikrommyd: Well we should have been skipping these either way. The only reason we did not have an issue is because uproot was saying that there is no known intepretation so they were being skipped because of that. In general, we should be skipping this `PARAMETERS` branch and all the subbranches when we do `tree.iteritems()` for these FCC files. If there is a better way, let me know. I couldn't tell from the docs of `tree.iteritems()` if I can say "skip this and all sub branches" somehow.

ikrommyd:
> @ikrommyd - thanks for looking into it! It seems to me that this is a temporary solution. Perhaps we should do some more digging? Thanks!

Yeah definitely needs digging on the uproot side but I'm not responsible for that. This is a patch that was needed anyways to exclude all these subbranches that we wanted to exclude anyways.

Issue #1386, opened 2025-08-05
Extracting more information during preprocessing


pfackeldey:
The preprocessing step could write out _persistently_ more information, such as:
- The awkward-forms of files (maybe just 1 per dataset)
- The lengths of all relevant forms (e.g. count branches)

This could be useful to constructing virtual arrays with `ak.from_buffers` without the need to even open the file. In addition, the preprocessing step may be done just once instead of every time if this information is serialized to json.

However, this does add some more data reading to the preprocessing step (reading count branches), which I'm not sure is wanted.

(in the perfect world we'd be able to query this information from dataset databases instead, but that's just wishful thinking on my side 😅)

ikrommyd:
The dask-awkward based preprocessing does all that. It creates a union form. It's a `save_form=True` argument that you need to turn on. The executor preprocessing doesn't though. This is valid issue however regarding merging the two into one preprocessing to rule them all.

NJManganelli:
@pfackeldey What might the necessary lengths info look like? I imagine it's simplest to have it live alongside the form (as saved by `preprocess(..., save_form=True)` with some key (`lengths` for example. Would it need lengths per-branch, and could it cross-reference the `form`'s list of contents, or would it just need to duplicate information from the form?)

pfackeldey:
> [@pfackeldey](https://github.com/pfackeldey) What might the necessary lengths info look like? I imagine it's simplest to have it live alongside the form (as saved by `preprocess(..., save_form=True)` with some key (`lengths` for example. Would it need lengths per-branch, and could it cross-reference the `form`'s list of contents, or would it just need to duplicate information from the form?)

It would be nice to have a known length per `form_key` for `ak.from_buffers`. That would eliminate the need to handle `unknown_length` in awkward-array. This is in principle solved by lazifying the lengths as well. However, that also means that if some ops only need to know the lengths, e.g. `ak.num`, they usually have to load the count branches from a file. If we have this information already in the layout we can skip loading those. 
Reading the lengths upfront once and storing this information can potentially reduce IO on every analysis processing afterwards. The assumption would be that reading the lengths is not a too expensive task for a preprocessing - but that's something you know better.

nsmith-:
Do you need to read the counts branches or just know the length of the buffers that hold the counts and the content?
For multiply-jagged data, would you descend and read all index arrays?

PR #1498, opened 2025-12-09
feat: implement persistent preprocessor cache

Ian2327:
**Problem:**
The current Coffea preprocessor repeated recomputes the same preprocessed ROOT data on every run, and considering the infrequency that these files tend to change, it causes unnecessary CPU use and longer overall processing time. Preprocessing can also be a large portion of the total processing time and since the processing stage can only commence after preprocessing has finished, if using a cluster with N workers, there is a high likelihood that many workers will be idling while waiting for the remaining workers to finish the preprocessing step. 

**Proposed Solution:**
Introduce a persistent, file-backed cache to the preprocessor stage.
Key features in this PR:

- A pickle-backed cache (`.coffea_metadata_cache.pkl`) stores per-file metadata after first access
- Data are automatically reused on subsequent runs if no changes to file metadata are detected
- Cache is loaded on Runner class initialization and written atomically after preprocessing
- Fallback to existing in-memory cache is retained if the persistent cache cannot be loaded or has not been created

**Impact:**
The persistent cache significantly reduces preprocessing time as shown in the graph below which is run on the UGE cluster. The left side shows a run with the original code, and the right side shows the same run with a filled persistent cache which clocks the preprocessing time to just under 6 milliseconds.
image

**Feedback:**
Before merging, any feedback on any aspect of the cache such as cache location/naming, eviction policy, configuration options, etc. would be greatly appreciated.


nsmith-:
This may address some of the scope of issue #1386 

You may also be interested in the coffea.compute refactor #1470

nsmith-:
Hi @Ian2327 can you resolve the conflicts and get the CI to pass? Thanks!

lgray:
@Ian2327 Could you please add some unit/integration testing to ensure consistent functioning? Thanks!

nsmith-:
Tests failing due to the new hist release ( https://github.com/scikit-hep/hist/releases/tag/v2.9.1 )

ikrommyd:
Hi @Ian2327, I think we should not enforce this and create pickle files for people without them asking. Can you make this an optional feature? It's probably best if people opt into this and manually provide a path for the pickle file.
Also this currently has extremely extensive testing. Just a single simple test is enough for a feature like this.

ikrommyd:
You are overcomplicating it. There is no need to have a boolean flag and a string path and a path object. You just have an attribute of the Runner class that is either a None or a string. If None, no metadata is cached. If a string, it can be an absolute or a relative path and it uses that location to cache the metadata.

Issue #1141, opened 2024-07-29
Unclear exception with processing dataset with no events


alexander-held:
**Describe the bug**
When a dataset (out of `dataset_tools.preprocess`) has zero chunks to run over, `dataset_tools.apply_to_fileset` will raise `FileNotFoundError` on `{}`. I think this usually happens when some other more serious thing has gone wrong at first but I occasionally see people run into this and it's not immediately obvious what happened.

**To Reproduce**
the following makes use of a fix for #1140 to run:
```python
import uproot
import awkward as ak
from coffea import dataset_tools
from coffea.nanoevents import BaseSchema
import dask

with uproot.recreate("f1.root") as f:
   f["tree"] = {"arr": ak.Array([])}
with uproot.recreate("f2.root") as f:
   f["tree"] = {"arr": ak.Array([1])}

fileset = {"dummy": {"files": {"f1.root": "tree"}}}
# fileset = {"dummy": {"files": {"f1.root": "tree", "f2.root": "tree"}}}  # this works
samples, _ = dataset_tools.preprocess(fileset)
tasks = dataset_tools.apply_to_fileset(lambda evts: None, samples, schemaclass=BaseSchema)
_ = dask.compute(tasks)
```

**Expected behavior**
A warning along the lines of "no useable files found for dataset xyz" and no exception raised by default.

**Output**
```pytb
Traceback (most recent call last):
  File "[...]]/test.py", line 17, in <module>
    tasks = dataset_tools.apply_to_fileset(lambda evts: None, samples, schemaclass=BaseSchema)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/coffea/dataset_tools/apply_processor.py", line 125, in apply_to_fileset
    dataset_out = apply_to_dataset(
                  ^^^^^^^^^^^^^^^^^
  File "[...]/coffea/dataset_tools/apply_processor.py", line 73, in apply_to_dataset
    ).events()
      ^^^^^^^^
  File "[...]/coffea/nanoevents/factory.py", line 684, in events
    events = self._mapping(form_mapping=self._schema)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/uproot/_dask.py", line 183, in dask
[...]

PR #1499, opened 2025-12-10
fix: add better error message when zero chunks survive preprocessing

ikrommyd:
The error message now is this
```
  File "/usr/local/lib/python3.11/site-packages/coffea/processor/executor.py", line 1762, in run
    wrapped_out, e = executor(chunks, closure, None)
    ^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable NoneType object
```
Because it's being raised during proper execution as we try to run the processor over a length zero list of chunks.
The output of preprocessing is never checked if it's empty. That's what this PR adds.

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1745,6 +1745,11 @@ class Runner:
             )
 
         chunks = list(chunks)
+        if len(chunks) == 0:
+            raise ValueError(
+                "No chunks survived preprocessing.\n"
+                "If you used skipbadfiles=True or similar, it is possible all your files are bad."
+            )
 
         exe_args = {
             "unit": "chunk",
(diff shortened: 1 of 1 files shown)


nsmith-:
This fixes #1141 for executors, though it was originally reported for dask processing.
Can you fix it on the dask side as well, just to wrap it up?

ikrommyd:
This is a tiny bit different. That issue is about files having zero events. This PR is about all files failing pre-processing which has a better error in the dask-based case already:
```
Exception: There was no populated list of files returned from querying your input dataset.
Please check your xrootd endpoints, and avoid redirectors.
Input dataset: dummy
As parsed for querying: [{'file': 'dummy.root', 'object_path': 'tree', 'steps': None, 'num_entries': None, 'uuid': None}]
```

ikrommyd:
@nsmith- I think this is good to go from my side as I have opened a separate PR for what you mentioned as it is a separate problem https://github.com/scikit-hep/coffea/pull/1501

nsmith- (approved):

PR #1501, opened 2025-12-10
feat: let zero-event files go through pre-processing

### LINKED ISSUE #1140: Preprocessing files with zero events
**Describe the bug**
When running preprocessing over files containing zero events, `num_entries` will be `0` and a division in the step size determination will fail with `ZeroDivisionError` here: https://github.com/CoffeaTeam/coffea/blob/e06c4b84d0a641ab569ae7c16fecc39fe74c9743/src/coffea/dataset_tools/preprocess.py#L127

I assume the solution would be to bypass all of https://github.com/CoffeaTeam/coffea/blob/e06c4b84d0a641ab569ae7c16fecc39fe74c9743/src/coffea/dataset_tools/preprocess.py#L79-L156 in this case to instead run https://github.com/CoffeaTeam/coffea/blob/e06c4b84d0a641ab569ae7c16fecc39fe74c9743/src/coffea/dataset_tools/preprocess.py#L158-L173
and return.

**To Reproduce**
```python
import uproot
import awkward as ak
from coffea import dataset_tools

arr = ak.Array([])
with uproot.recreate("f.root") as f:
    f["tree"] = {"arr": arr}

fileset = {"dummy": {"files": {"f.root": "tree"}}}
dataset_tools.preprocess(fileset)
```
The actual case where I saw this happen contained a tree without any arrays inside. I'm not sure if `uproot` supports writing that for a reproducer (I don't think so?) but in either case I think only the `.num_entries` property matters so the tree content should be irrelevant here.

**Expected behavior**
handle empty files without crash

**Output**
```
  File "[...]/lib/python3.11/site-packages/coffea/dataset_tools/preprocess.py", line 127, in get_steps
    n_steps_target = max(round(num_entries / target_step_size), 1)
                               ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
ZeroDivisionError: division by zero
```

n/a

**Additional context**
cc @sebastien-rettie

### LINKED ISSUE #1141: Unclear exception with processing dataset with no events

ikrommyd:
Fixes https://github.com/scikit-hep/coffea/issues/1141 and https://github.com/scikit-hep/coffea/issues/1140

I think we can just append a length zero chunk with no problems during execution (we should be able to read with `entry_start=entry_stop=0`)


ikrommyd:
@nsmith- what's your opinion here. The executors currently silently skip length zero files from the chunk list entirely due to this:
https://github.com/scikit-hep/coffea/blob/7548af72dab42ae95ee77140a9398476ea8fd8b0/src/coffea/processor/executor.py#L150
What should we do here. Should we just output 0 steps? Should we always skip the file? Should we only skip the file if `skip_bad_files=True`? All 3 are possible and I don't wanna change too many things with the whole `coffea.compute` thing.

nsmith-:
I suppose it comes down to what "bad file" means. A file with no events is certainly suspicious, but it may be expected if upstream a skim was applied. Since the original issue asked to skip, I would just go for whatever allows it to be ignored easiest.

ikrommyd:
@nsmith- now we should get length 0 valid chunks out of pre-processing in both executors and the dask-based preprocessing too. This is my testing script locally which I can turn into a pytest
```py
import awkward as ak
import uproot
import dask
from coffea import processor
from coffea.nanoevents import NanoAODSchema, BaseSchema
from coffea.dataset_tools import preprocess, apply_to_fileset


class MyProcessor(processor.ProcessorABC):
    def process(self, events):
        return len(events)


with uproot.recreate("dummy1.root") as f:
    f["Events"] = {"arr": ak.Array([])}
with uproot.recreate("dummy2.root") as f:
    f["Events"] = {"arr": ak.Array([])}
with uproot.recreate("dummy3.root") as f:
    f["Events"] = {"arr": ak.Array([1, 2, 3])}

fileset = {
    "dataset1": {
        "files": {
            "dummy1.root": "Events",
            "dummy2.root": "Events",
        },
        "metadata": {
            "is_mc": True,
        },
    },
    "dataset2": {
        "files": {
            "dummy2.root": "Events",
            "dummy3.root": "Events",
        },
        "metadata": {
            "is_mc": True,
        },
    },
    "dataset3": {
        "files": {
            "dummy3.root": "Events",
        },
        "metadata": {
            "is_mc": True,
        },
    },
}


iterative_run = processor.Runner(
    executor=processor.IterativeExecutor(compression=None),
    schema=BaseSchema,
    chunksize=100_000,
    savemetrics=True,
    skipbadfiles=False,
    align_clusters=False,
)
samples = iterative_run.preprocess(fileset)
samples = list(samples)
print(len(samples))
print(samples)

out, metrics = iterative_run(
    samples,
    processor_instance=MyProcessor(),
)
print(out)

print("-" * 100)
samples, _ = preprocess(fileset, align_clusters=False)
print(samples)
tasks = apply_to_fileset(MyProcessor(), samples, schemaclass=BaseSchema)
out = dask.compute(tasks)[0]
print(out)
```
which prints out this
image

nsmith-:
What if the user processor reads a particular field inside the events? With no entries in a file it might not have a compatible schema in a processor.

ikrommyd:
> What if the user processor reads a particular field inside the events? With no entries in a file it might not have a compatible schema in a processor.

I would assume that in this the case that the rest of the structure is the same no? A length zero event chunk is the same as any other chunk except it has length zero. All the branches etc would be there, they would just be empty. Otherwise we're talking about reading files with different structures together right?

ikrommyd:
I created an empty copy of nano_dy using
```
rooteventselector --recreate -f -1 -l 0 nano_dy.root:Events nano_dy_empty.root
```
that we can use for tesing 
```
In [1]: from coffea.nanoevents import NanoEventsFactory, NanoAODSchema

In [2]: fname = "/Users/iason/work/pyhep_dev/coffea/tests/samples/nano_dy_empty.root"; events = NanoEventsFactory.from_root({fname: "Events"}, access_log=(access_log := [
      ⋮ ]), mode="eager").events()
      
In [3]: events
Out[3]: <NanoEventsArray [] type='0 * event'>

In [4]: events.Electron.pt
Out[4]: <Array [] type='0 * var * float32[parameters={"__doc__": "p_{T}", "typename...'>
```

ikrommyd:
@nsmith- I've added testing for dataset tools and executors

nsmith- (approved):
Looks good to me. @alexander-held can you take a peek at the test to see if this covers the problem?

Issue #659, opened 2022-04-11
Multi-tree support


kreczko:
Following up on the discussion in the CMS Analysis Tools Task Force Monday Open Meeting

**Is your feature request related to a problem? Please describe.**
I am trying to process multiple TTree inputs at once using coffea

**Describe the solution you'd like**
I've been tracing `treeName` through the coffea code, and it seems there is no simple way to allow for multiple trees.
I would like to
- `coffea.processor.run_uproot_job` to accepts a list of treenames
- `coffea.processor.ProcessorABC.process` data parameter to provide access to these data via `df["treename"]["varname]` or `df["treename.var"]` (not fussed about the actual delimiter)

**Additional context**
The CMS L1T analysis (for L1T development) needs to access multiple trees.

Example files (only small samples available publicly):
```json
{
    "2k.root": "https://cernbox.cern.ch/index.php/s/bRvsnkhSl3eCFNl/download",
    "2kmc.root": "https://cernbox.cern.ch/index.php/s/eyAEZH6LQyiySQs/download",
    "2kmu.root": "https://cernbox.cern.ch/index.php/s/FcWgoWCOg6vXOAn/download"
}
```

Example treenames:
```yaml
treenames:
      - "l1CaloTowerEmuTree/L1CaloTowerTree"
      - "l1CaloTowerTree/L1CaloTowerTree"
```
example variables:
```yaml
- "l1CaloTowerTree/L1CaloTowerTree.L1CaloTower.et" 
- "l1CaloTowerEmuTree/L1CaloTowerTree.L1CaloTower.et"
```
Note: An uproot file handle can access these variables via `<directory>/<tree name>/<object>/<attribute>`:
```python
import uproot
f = uproot.open("2kmu.root")
emu_calo_et = f["l1CaloTowerEmuTree/L1CaloTowerTree/L1CaloTower/et"].array()
calo_et = f["l1CaloTowerTree/L1CaloTowerTree/L1CaloTower/et"].array()
```

alexander-held:
There is also a fairly common workflow in ATLAS which involves dedicated trees for systematic variations. Multi-tree support would come in handy there too.

lgray:
This is now possible to a limited extent with coffea 2023. 
We don't have full table joining but you can fill a histogram from two separate dask-awkward sources and/or specify files in the same dataset with different trees.

It is not incredibly efficient yet but it does work now.

lgray:
[...]

PR #1502, opened 2025-12-10
feat: add file handle as an attribute of the events factory

MoAly98:
When opening files with uproot, it may be useful for benchmarking purposes to be able to access the file object to extract information such as the number of requested bytes inside a processor. Currently this is not possible as the file handle is not exposed inside the processor. This PR includes the file handle as metadata inside the processor. 

I have tested the PR with the following script:

```
from coffea import processor
from dask.distributed import Client
import hist
import awkward as ak


class SimpleProcessor(processor.ProcessorABC):
  def process(self, events):

    # Inspect metadata
    self.metadata = events.metadata
    filesource = self.metadata.get("filesource", None)
    print(f"Number of requested bytes 0: {filesource.num_requested_bytes}")

    jets = events.Jet[events.Jet.pt > 30]

    print(f"Number of requested bytes 1: {filesource.num_requested_bytes}")

    muons = events.Muon[events.Muon.pt > 10]
    print(f"Number of requested bytes 2: {filesource.num_requested_bytes}")

    h = hist.Hist.new.Reg(50, 0., 500., name="pt").Double()
    h.fill(pt=ak.flatten(jets.pt))

    return {"hist": h, "nevents": len(events)}

      return accumulator


if __name__ == "__main__":

    fileset = {
        "signal": {
            "files": {
                "root://eospublic.cern.ch//eos/opendata/cms/mc/RunIISummer20UL16NanoAODv9/ZprimeToTT_M2000_W20_TuneCP2_PSweights_13TeV-madgraph-pythiaMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v1/270000/22BAB5D2-9E3F-E440-AB30-AE6DBFDF6C83.root": "Events"
            }
        }
    }


    itexec = processor.IterativeExecutor()

    run = processor.Runner(executor=itexec, schema=NanoAODSchema, savemetrics=True, chunksize=5_000)

    output, metrics = run(fileset, processor_instance=SimpleProcessor())

    print(f"Processed {output['nevents']} events")
    print(f"Histogram entries: {output['hist'].sum()}")
```

cc: @pfackeldey


lgray:
@MoAly98 I'm in principle fine with this addition. I agree with your proposed uses. Please make your tests pass and we'll continue with review.

nsmith-:
Is there some reason why https://github.com/scikit-hep/coffea/blob/a4e1d6c940796e22abedbccb269abaa48cedc97e/src/coffea/processor/executor.py#L1538-L1539 is not doing the job?
If you want to add additional metrics we very much welcome it. Otherwise this seems like it will just create an opportunity for user recipes that propagate without maintainability. If you need to add this for debugging purposes, you can always do so on a personal fork.

nsmith-:
And here's a way to access it from the nanoevents object now:
```python
tree = list(events.attrs["@events_factory"]._mapping._cache.values())[0]
print(tree.file.source.num_requested_bytes)
```
not pretty though :)

pfackeldey:
> Is there some reason why
> 
> https://github.com/scikit-hep/coffea/blob/a4e1d6c940796e22abedbccb269abaa48cedc97e/src/coffea/processor/executor.py#L1538-L1539
> 
> is not doing the job?
> If you want to add additional metrics we very much welcome it. Otherwise this seems like it will just create an opportunity for user recipes that propagate without maintainability. If you need to add this for debugging purposes, you can always do so on a personal fork.

I think another good motivation (apart from additional metadata) is that there may be additional data in the file that is not present in `events`, e.g., a `TH1` with some data that one wants to access during the processing. This is not the case for NanoAOD, but may be the case for other input data.

MoAly98:
@nsmith- The current implementation allows the user to access the `bytesread` aggegated over all chunks. One use case I already have for this new implementation is to track per chunk information in a clean way. This helps with performance evaluation as we have been trying to do for the integration challenge.

@pfackeldey the current implementation i only keep the file source, but your point is interesting and I didn't think about it. It would warrant the inclusion of the full file handle. If we do so, I think I would also add a forced `del metadata["filehandle"]` before we return the output to user, making sure we don't have a dangling reference.

MoAly98:
@lgray this is now passing. It should be easy to expose the file handle instead of the source, to benefit from a setup like the one @pfackeldey describes. Let me know if you think that would be a better way forward!

nsmith-:
That would also address some of #659 but I really am concerned we are creating a workaround to the larger problem of capturing more complex processing workflows than iterating over events in a single tree. The disadvantage of not having it as part of the framework are reduced ability to implement automatic recovery and more chances for user error, since they have to manage the relation between the chunks of a tree and the rest of the file themselves. The hope with #1470 was to design something that could capture this.

MoAly98:
@nsmith-  Maybe this is some ignorance on my side, but I am not sure I understand why exposing extra information within the processor to a (power)-user would be adding complexity, given that it does not really change code logic. A nominal user would not be looking for this, and will not be hurt by the existence of the information in the metadata. 

I agree in the need for a more complex overhaul of the executor code to allow more flexibility to user, and I have some ideas there that I will at some point add to #1470. For now, I see this PR as a low-cost additional flexibility for people who want a more thorough control.

nsmith-:
OK, then at the very least I would say to expose the more useful file handle per

> It should be easy to expose the file handle instead of the source, to benefit from a setup like the one @pfackeldey describes

pyproject.toml
```diff
@@ -142,3 +142,25 @@ line-length = 160
 
 [tool.ruff.lint]
 ignore = ["F403", "F405", "E402"]
+
```
lgray: Can we move all the pixi dust to a different PR? I'd prefer that we keep the PR focused to the advertised change.

MoAly98: oops you're right, i added these by mistake

MoAly98: @lgray done!

MoAly98:
@lgray I think this should be ready :)

ikrommyd:
Why do I have a feeling that it is better to have this as an attribute of the factory like in https://github.com/scikit-hep/coffea/pull/1504.
I think `events.attrs["@events_factory"].filehandle` or similar is a better way to access this that doesn't add extra info to the metadata that the regular user uses.
Is there a reason why y'all think it's better if it lives in the metadata?

MoAly98:
@ikrommyd I am only keeping this in the same place as rest of the file information, such as the name, uuid, etc... I don't have a strong feeling here!

ikrommyd:
@MoAly98 I think it's best to add it to the factory. There's a difference between the file handle and name, uuid etc and that is that everything else is a string. There is a general idea to have things as json-serializable in the fileset and therefore in the metadata too.  Could you please make this change (similar to the access log basically)?

ikrommyd (approved):

src/coffea/processor/executor.py
```diff
@@ -1550,6 +1550,10 @@ def _work_function(
                     )
                 # save the output
                 checkpointer.save(out, metadata, processor_instance)
+
+            # avoid potential dangling references to file
+            del factory._mapping._file_handle
+
```
nsmith-: Can you explain a bit why this is needed? The factory should go out of scope after this function exits, cleaning up any of its owned data.

src/coffea/nanoevents/factory.py
```diff
@@ -438,6 +440,7 @@ def from_root(
             entry_stop,
             cache={},
             access_log=access_log,
+            file_handle=file_handle,
```
nsmith-, line 443: In principle, the `TrivialUprootOpener` is in charge of actually opening the file to fetch data, so the file handle used may not be the same as this one. In practice, the `mapping.preload_column_source()` call immediately after this makes sure it is the same. The concept is that `NanoEventsFactory` objects themselves are pickleable, though we don't actually need that requirement anymore, I think.

nsmith- (approved):
This looks good, though I think the early deletion is extraneous. Let me know if I'm missing something.

src/coffea/processor/executor.py
MoAly98: this was because i didn't want any dangling references to the file handle when i used the `events.metadata`. This can be removed now.

src/coffea/nanoevents/factory.py
MoAly98: @nsmith- do you propose here to take the handle returned from the `TrivialUprootOpener` directly inside the `UprootSourceMapping` method instead of passing it through?

I am not sure how that links to pickleable factories though.  i am also not sure  how `preload_column_source()` is ensuring consistency either.

src/coffea/processor/executor.py
ikrommyd: Done

src/coffea/nanoevents/factory.py
nsmith-: I should have specified the above is just a comment, no action is needed.

lgray (approved):

PR #1504, opened 2025-12-12
feat: add the access log as an attribute of the events factory

MoAly98:
This very small change exposes the `access_log` of columns to the events factory, which can in turn be accessed by an interested user inside their coffea processor. This provides a granular handle to the quantity that shows up in the coffea report if `savemetrics=True`, so that user can investigate what they are reading from each file, and they can compute quantities such as the (compressed/uncimpressed) bytes read per file. 

I tested the PR with the followign small example:

```
from coffea import processor
import hist
import awkward as ak


class SimpleProcessor(processor.ProcessorABC):
    def process(self, events):
        # Get access_log from factory
        factory = events.attrs["@events_factory"]
        access_log = factory.access_log

        print(f"Accessed branches before: {len(access_log)}")

        jets = events.Jet[events.Jet.pt > 30]
        print(f"Accessed branches after Jet.pt: {len(access_log)}")

        muons = events.Muon[events.Muon.pt > 10]
        print(f"Accessed branches after Muon.pt: {len(access_log)}")

        print("Access log:")
        for entry in access_log:
            print(f"  - {entry.branch}")

        h = hist.Hist.new.Reg(50, 0., 500., name="pt").Double()
        h.fill(pt=ak.flatten(jets.pt))

        return {"hist": h, "nevents": len(events)}

        return accumulator


if __name__ == "__main__":
    fileset = {
        "ZJets": {
            "files": {
                "root://eospublic.cern.ch//eos/opendata/cms/mc/RunIISummer20UL16NanoAODv9/ZprimeToTT_M2000_W20_TuneCP2_PSweights_13TeV-madgraph-pythiaMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v1/270000/22BAB5D2-9E3F-E440-AB30-AE6DBFDF6C83.root": "Events"
            }
        }
    }

    itexec = processor.IterativeExecutor()
    run = processor.Runner(executor=itexec, schema=NanoAODSchema, savemetrics=True, chunksize=5_000)

    output, metrics = run(fileset, processor_instance=SimpleProcessor())

    print(f"Processed {output['nevents']} events")
    print(f"Histogram entries: {output['hist'].sum()}")

```

### DIFF
--- a/tests/test_nanoevents.py
+++ b/tests/test_nanoevents.py
@@ -236,14 +236,13 @@ def test_access_log(tests_directory, mode):
     events = factory.events()
 
     assert factory.access_log is access_log
+    if mode == "eager":
+        assert len(factory.access_log) > 1500
 
-    if mode == "virtual":
+    elif mode == "virtual":
         # In virtual mode, access_log starts empty until columns are accessed
-        assert len(access_log) == 0
+        assert len(factory.access_log) == 0
         # Access a column to trigger lazy loading
-        _ = events.Muon.pt
-
-    # In both modes, access_log should be populated after accessing data
-    assert len(access_log) > 0
-    branches = {entry.branch for entry in access_log}
-    assert "Muon_pt" in branches or "nMuon" in branches
+        _ = ak.materialize(events.Muon.pt)
+        branches = {entry.branch for entry in factory.access_log}
+        assert branches == {"nMuon", "Muon_pt"}
--- a/tests/test_nanoevents.py
+++ b/tests/test_nanoevents.py
@@ -210,3 +210,40 @@ def test_missing_eventIds_warning_dask(tests_directory):
             mode="dask",
         ).events()
         events.Muon.pt.compute()
+
+
+@pytest.mark.parametrize("mode", ["eager", "virtual"])
+def test_access_log(tests_directory, mode):
+    """Test that access_log is available on the factory."""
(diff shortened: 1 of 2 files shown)


tests/test_nanoevents.py
```diff
@@ -210,3 +210,40 @@ def test_missing_eventIds_warning_dask(tests_directory):
...
+    assert factory.access_log is access_log
+
+    if mode == "virtual":
+        # In virtual mode, access_log starts empty until columns are accessed
+        assert len(access_log) == 0
+        # Access a column to trigger lazy loading
+        _ = events.Muon.pt
+
+    # In both modes, access_log should be populated after accessing data
+    assert len(access_log) > 0
```
ikrommyd: in eager, access_log should be completely full upon opening the file so it would move this above `if mode == "virtual":` and maybe assert something like that it is > 1000 even (all the branches will be loaded). Then in the `if mode == "virtual"` case, you can just trigger materialization of the muon pt for example by doing `events.Muon.pt * 1`, and then assert that exactly `nMuon` and `Muon_pt` are the only branches that are present in the access_log.

nsmith- (approved):
Looks fine to me

tests/test_nanoevents.py
ikrommyd: I've done this refactoring

ikrommyd (approved):

PR #1505, opened 2025-12-12
chore: more restrictive linting in ruff

lgray:
... and corresponding linting


lgray:
Ah some circular import shenanigans. I'll fix it up later.

PR #1507, opened 2025-12-16
feat: disable virtual array caching on themselves if a nanoevents buffer cache is used

ikrommyd:
This PR disables the caching of the materialized virtual arrays on themselves in the case where a global buffer cache is used at the factory. In that case, the loaded buffers will only be cached there and not on the virtual array objects as well.


pfackeldey (approved):
Looks good to me :+1:

I think this means now that we want to remove the existing column cache: https://github.com/scikit-hep/coffea/blob/master/src/coffea/processor/dask/__init__.py#L10, and instead implement executor agnostic buffer caches, right? (i.e. they still can use the LRU, compression, etc logic, but shouldn't be dask worker plugins)

PR #1511, opened 2026-01-16
feat: Schema updates to run on PHYSLITE processed through ServiceX + link improvements

ponyisi:
Tweaks to support handling of `ElementLinks` when processed through ServiceX (will still work with standard PHYSLITE). Give muons minimum four-vector functionality and improve set of linked objects.

### DIFF
--- a/src/coffea/nanoevents/mapping/preloaded.py
+++ b/src/coffea/nanoevents/mapping/preloaded.py
@@ -5,9 +5,9 @@ import awkward
 import numpy
 
 from coffea.nanoevents.mapping.base import BaseSourceMapping, UUIDOpener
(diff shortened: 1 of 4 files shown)


lgray:
Can you add a test for the functionality you're adding here to ensure it keeps functioning into the future and doesn't drift away from us?

lgray:
I think it would be nice in the future to have an example of the ServiceX output that this is intended to be run on, but I can imagine the format of that may be in flux at the moment. Given that, and that required functionality is tested by other means right now I think that can wait.

Otherwise everything appears to be in scope, and I expect there will be further updates as things evolve on the service X side.

lgray (approved):

Issue #1483, opened 2025-11-18
Dynamic chunking in coffea.compute


nsmith-:
@btovar thanks for bringing up the dynamic chunking, I now remember you had implemented that at some point in the TaskVine Executor. So then we re-conceptualize `Computable` not as in iterator but perhaps more a `Generator` with a send channel for the Backend to request changes to the iterator in flight. Maybe something like:
```python
from typing import Generator, Protocol, TypeAlias

from coffea.compute.protocol import (
    Computable,
    EmptyResult,
    InputT,
    ResultT,
    WorkElement,
)


class SizedWorkElement(WorkElement[InputT, ResultT], Protocol):
    def __len__(self) -> int:
        "Return the size of this work element in some unit (e.g., number of events)"
        ...


NewSizeRequest: TypeAlias = int


class ResizableComputable(Computable[InputT, ResultT], Protocol):
    def generate(
        self,
    ) -> Generator[SizedWorkElement[InputT, ResultT], NewSizeRequest, None]:
        "Generate work elements, possibly adapting their size based on external factors"
        ...


def compute_now(items: ResizableComputable[InputT, ResultT]) -> ResultT | EmptyResult:
    out = EmptyResult()
    work_gen = items.generate()
    # Let it tell us the initial size
    work_element = next(work_gen, None)
    if work_element is None:
        return out
    while True:
        result = work_element()
        out += result
        # Here we could adapt the size of future work elements based on performance metrics
        # For simplicity, we just request the same size
        try:
            work_element = work_gen.send(len(work_element))
        except StopIteration:
            break
    return out
```

(a real implementation would have to wrap this into a Task)

This would require a bit of re-imagining how the `FailedTaskElement` is implemented, it could not anymore just keep track of the index in the iteratable, but the whole materialized WorkElement for later re-computation.

_Originally posted by @nsmith- in https://github.com/scikit-hep/coffea/issues/1470#issuecomment-3512514319_

nsmith-:
@btovar a few questions:
- Should the next element be guaranteed to be exactly `NewSizeRequest` big? At least as big? Always larger? Are we trying to avoid too large a task that it exhausts resources or too many small tasks?
- If a work element fails, do we just append it to a list of failed work elements? Then any retry will no longer be resizable. Is that OK?

[...]

PR #1512, opened 2026-01-26
chore: Try to factor out some common parts of threaded backend

nsmith-:
FYI @pfackeldey


nsmith-:
So here's a boiled down view of the motivation for having a background thread,
```python
import time
from collections.abc import Iterable
from concurrent.futures import Future, ThreadPoolExecutor
from dataclasses import dataclass

from coffea.compute.data.workelement import DataWorkElement
from coffea.compute.errors import TaskElement


@dataclass(frozen=True)
class SomeData:
    item: int

    def load(self) -> int:
        return self.item


SomeTaskElement = TaskElement[int, int]


def square(x: int) -> int:
    time.sleep(0.01)
    return x * x


@dataclass
class SimpleTaskImpl:
    """Simpler task implementation that just runs everything immediately.

    Without an active thread managing collection of results, the futures
    will hold onto the results directly, which is not ideal.
    """

    inflight: set[tuple[SomeTaskElement, Future]]

    def result(self) -> int:
        return sum(f.result() for _, f in self.inflight)

    def partial_result(self) -> tuple[int, list[SomeTaskElement]]:
        """This is something a standard Future does not have."""
        partial_sum = 0
        pending = []
        for w, f in self.inflight:
            if f.done():
                partial_sum += f.result()
            else:
                pending.append(w)
        return partial_sum, pending

    def wait(self) -> None:
        for _, f in self.inflight:
            f.result()

    def done(self) -> bool:
        return all(f.done() for _, f in self.inflight)

    def cancel(self) -> None:
        for _, f in self.inflight:
            f.cancel()


class BackendImpl:
    """A backend receives a computable, which is just a collection of work elements,
    and executes them, returning a Future-like object "TaskImpl" that eventually
    holds the summed result. The one extra feature we need is to be able to peek
    at the partial results while the task is still running, potentially stopping
    early. To be able to do so, the backend should not block on the entire task.
    """

    pool: ThreadPoolExecutor

    def __enter__(self) -> "BackendImpl":
        self.pool = ThreadPoolExecutor(max_workers=1)
        self.pool.__enter__()
        return self

    def __exit__(self, exc_type, exc_value, traceback) -> None:
        self.pool.__exit__(exc_type, exc_value, traceback)

    def compute(self, work: Iterable[SomeTaskElement]) -> SimpleTaskImpl:
        return SimpleTaskImpl(inflight={(el, self.pool.submit(el)) for el in work})


def test_basic_protocol() -> None:
    work = [TaskElement(i, DataWorkElement(square, SomeData(i))) for i in range(10)]

    with BackendImpl() as backend:
        task = backend.compute(work)

        time.sleep(0.05)  # let some of the work complete
        partial, pending = task.partial_result()
        assert partial > 0
        assert len(pending) < 10
        print(f"Partial result: {partial}, pending: {pending}")

        assert task.result() == sum(i * i for i in range(10))


if __name__ == "__main__":
    test_basic_protocol()
```

nsmith-:
A possible way to reduce the boilerplate in the task impl. is to make some composable generators:
```python
from functools import partial
from typing import Callable, Generator, Iterable, Iterator, Protocol


class CloneableIterator[T](Protocol):
    """An iterator that can be cloned, where the clone starts from the current position."""

    def __iter__(self) -> Iterator[T]: ...

    def clone(self) -> "CloneableIterator[T]": ...


class ResumeableIterable[T]:
    def __init__(self, input: Iterable[T], offset: int = 0):
        if hasattr(input, "__next__"):
            raise TypeError("Input must be an iterable, but not an iterator")
        self._input = input
        self._offset = offset

    def __iter__(self):
        for index, element in enumerate(self._input):
            if index < self._offset:
                continue
            yield element
            self._offset = index + 1

    def clone(self) -> "ResumeableIterable[T]":
        return ResumeableIterable(self._input, self._offset)


def retry[T](iterable: Iterable[T], retries: int):
    for item in iterable:
        attempt = 0
        while True:
            try:
                yield item
                break
            except Exception:
                attempt += 1
                if attempt > retries:
                    raise


def run[T, S](generator: Generator[Callable[[], T], S, None]):
    try:
        func = next(generator)
    except StopIteration:
        return
    while True:
        try:
            send_value = yield func()
        except Exception as e:
            generator.throw(e)
        else:
            try:
                func = generator.send(send_value)
            except StopIteration:
                return


failures = 0


def square(x: int) -> int:
    global failures
    if x == 5:
        failures += 1
        if failures <= 1:
            raise ValueError("Intentional failure on 5")
    return x * x


if __name__ == "__main__":
    work = [partial(square, i) for i in range(10)]

    gen = ResumeableIterable(work)
    print("Running normally with retries:")
    for val in run(retry(gen, retries=2)):
        print(f"Got {val}")

    failures = 0

    print("Running until exception:")
    gen = ResumeableIterable(work)
    try:
        for val in gen:
            print(f"Got {val()}")
    except ValueError:
        pass
    print("Cloning and resuming")
    gen2 = gen.clone()
    for val in gen2:
        print(f"Got {val()}")
```

nsmith-:
Ok new plan, I think this gets everything we need
```python
from __future__ import annotations
from concurrent.futures import FIRST_COMPLETED, Future, ThreadPoolExecutor, wait
from dataclasses import dataclass
from functools import partial, reduce
import operator
import time
from typing import Callable, Generator, Generic, Iterable, Protocol

from coffea.compute.protocol import (
    Addable,
    EmptyResult,
    ResultT,
)


class Pool(Protocol):
    def submit[T](self, item: Callable[[], T], /) -> Future[T]: ...


class Submitter:
    """Submit work items to the pool, yielding futures as they complete.

    Arguments:
        pool: the thread pool to submit work to
        max_inflight: maximum number of work items to have inflight at once
        poll_interval: how often to poll for cancel calls (in seconds)

    Note: max_inflight should be tuned according to the expected number of
    workers in the pool.
    """

    def __init__(
        self, pool: Pool, *, max_inflight: int, poll_interval: float = 0.1
    ) -> None:
        self.pool = pool
        self.max_inflight = max_inflight
        self.poll_interval = poll_interval
        self.running = True

    def cancel(self) -> None:
        self.running = False

    def __call__[T](
        self, work: Iterable[Callable[[], T]]
    ) -> Generator[Future[T], None, None]:
        workiter = iter(work)
        inflight: set[Future[T]] = set()
        while (element := next(workiter, None)) or inflight:
            if element:
                fut = self.pool.submit(element)
                inflight.add(fut)
            if len(inflight) >= self.max_inflight or element is None:
                done, inflight = wait(
                    inflight, timeout=self.poll_interval, return_when=FIRST_COMPLETED
                )
                yield from done
            if not self.running:
                break
        self.running = False
        for fut in inflight:
            fut.cancel()


def add(xs: Iterable[Future[ResultT]]) -> ResultT:
    return reduce(operator.add, (x.result() for x in xs))


class Reducer:
    """Merge futures into larger sums, yielding futures for the merged results.

    Arguments:
        pool: the thread pool to submit merge work to
        merge_every: how many futures to merge at once
    """

    def __init__(self, pool: Pool, *, merge_every: int):
        self.pool = pool
        self.merge_every = merge_every

    def __call__(
        self, work: Iterable[Future[ResultT]]
    ) -> Generator[Future[ResultT], None, None]:
        workiter = iter(work)
        ready: list[Future[ResultT]] = []
        while (item := next(workiter, None)) or ready:
            if item:
                ready.append(item)
            if len(ready) >= self.merge_every or item is None:
                # print(f"Submitting merge of {len(ready)} elements")
                yield self.pool.submit(partial(add, ready))
                ready = []


@dataclass
class ResultHolder[T]:
    """Box to hold the result of an accumulation.

    The data is set via callbacks from futures as partial results become ready.
    """

    result: T | None = None
    """The accumulated result, or None if not yet available"""
    final: bool = False
    """Whether the result is final (computation complete)"""
    error: Exception | None = None
    """Any error encountered during accumulation"""

    def set_result(self, fut: Future[T], *, final: bool = False) -> None:
        try:
            self.result = fut.result()
        except Exception as e:
            self.error = e
        self.final = final


class Accumulator:
    """Accumulate futures into a single final result.

    Returns partial sums as futures are added together.

    Arguments:
        pool: the thread pool to submit merge work to
    """

    def __init__(self, pool: Pool):
        self.pool = pool

    def __call__(
        self, work: Iterable[Future[ResultT]], holder: ResultHolder[ResultT]
    ) -> ResultT:
        workiter = iter(work)
        partial_sum = next(workiter, None)
        if partial_sum is None:
            raise RuntimeError("No work to accumulate")
        partial_sum.add_done_callback(holder.set_result)
        while (fut := next(workiter, None)) and holder.error is None:
            partial_sum = self.pool.submit(partial(add, (partial_sum, fut)))
            partial_sum.add_done_callback(holder.set_result)
        holder.set_result(partial_sum, final=True)
        return partial_sum.result()


@dataclass
class ResultWrapper[T: Addable]:
    result: T | EmptyResult
    successes: list[int]
    failures: list[tuple[int, Exception]]

    def __add__(self, other: ResultWrapper[T]) -> ResultWrapper[T]:
        return ResultWrapper(
            self.result + other.result,
            self.successes + other.successes,
            self.failures + other.failures,
        )


failures = 0


def square_bad(x: int) -> int:
    global failures
    print(f"Squaring {x}")
    time.sleep(0.05)
    if x == 5:
        failures += 1
        if failures <= 1:
            raise ValueError("Intentional failure on 5")
    return x * x


def square(x: int) -> int:
    # print(f"Squaring {x}")
    # time.sleep(0.05)
    return x * x


@dataclass
class WorkWrapper(Generic[ResultT]):
    index: int
    work: Callable[[], ResultT]

    def __call__(self) -> ResultWrapper[ResultT]:
        try:
            res = self.work()
        except Exception as e:
            return ResultWrapper(EmptyResult(), [], [(self.index, e)])
        else:
            return ResultWrapper(res, [self.index], [])


if __name__ == "__main__":
    work = [WorkWrapper(i, partial(square_bad, i)) for i in range(40)]
    # work = [partial(square_bad, i) for i in range(40)]

    # to make this not slow, increase max_inflight and merge_every (e.g. 64, 8)
    # work = (partial(square, i) for i in range(1_000_000))

    with ThreadPoolExecutor(max_workers=4) as tpool:
        submitter = Submitter(tpool, max_inflight=4)
        reducer = Reducer(tpool, merge_every=4)
        accumulator = Accumulator(tpool)
        res = ResultHolder[ResultWrapper[int]]()

        try:
            # accumulator(reducer(submitter(work)), res)
            tpool.submit(partial(accumulator, reducer(submitter(work)), res)).result()
        except KeyboardInterrupt:
            print("Cancelled by user")
            submitter.cancel()
        print("Total:", res)
```

nsmith-:
From discussion with @pfackeldey 
- Computable is an iterable of "Steps" (ranges of entries) always
- An optional Computable method would generate steps and receive some send type that tells if the range is too big
- Submitter might want a `pool.map` for efficiency over `pool.submit` one at a time
- A generator type that handles writing out array result types would be useful

nsmith-:
The compute interface changed a tiny bit: I introduced a `ResultWrapper` to handle returning partial results, and now `Computable` explicitly returns ranges of data through its `gen_steps` call. This gets us ready for #1483; should be straightforward after a refactoring of the data side. Otherwise the protocol is largely the same: `Task` is still a fancy future, and `WorkElement` is still a closure, though now the input type is erased.

pfackeldey (approved):
Looks good to me :+1:

We discussed this thoroughly in the last coffea dev meeting, and I think this is a good structure that avoids as much as possible diving into threading shenanigans, which ultimately should make it more accessible to write custom compute backends.

PR #1514, opened 2026-01-26
chore: fix treemaker schema uproot writing test to write TTree

ariostas:
Uproot switched to writing RNTuples by default, so this test needed to be updated since the `TTreeMakerSchema` doesn't work for RNTuples.

### DIFF
--- a/tests/test_nanoevents_treemaker.py
+++ b/tests/test_nanoevents_treemaker.py
@@ -90,7 +90,7 @@ def test_uproot_write():
     ).events()
 
     with uproot.recreate("treemaker_write_test.root") as f:
-        f["PreSelection"] = TreeMakerSchema.uproot_writeable(orig_events)
+        f.mktree("PreSelection", TreeMakerSchema.uproot_writeable(orig_events))
 
     test_events = NanoEventsFactory.from_root(
         {"treemaker_write_test.root": "PreSelection"},


ikrommyd:
I don't remember something here, what is the thing that doesn't work? You can't write treemaker structure or is it that the schema can't read rntuple? What's the error?

ariostas:
> You can't write treemaker structure or is it that the schema can't read rntuple? What's the error?

The latter. The schema can't read RNTuple. Here's the error: https://github.com/scikit-hep/coffea/actions/runs/21343519788/job/61426792760#step:9:2114

ikrommyd:
This says
```
>               raise AttributeError(f"no field named {where!r}")
E               AttributeError: no field named 'pt'. Did you mean: 'ptD'?
```
so I guess something is not being written properly?

ariostas:
If I make this change

```python
with uproot.recreate("treemaker_write_test.root") as f:
    array = TreeMakerSchema.uproot_writeable(orig_events)
    print(array["Jets"].fields)
    f["PreSelection"] = array
    
with uproot.open("treemaker_write_test.root") as f:
    print(f["PreSelection"].arrays().Jets.fields)
```
It prints the same before writing and after reading. So I don't know where `pt` is coming from.

ariostas:
Actually, it's because `TreeMakerSchema` renames `/.fPt` into `pt`. For some reason that part is not done correctly for RNTuples. Also, I just realized that Uproot should not be letting you write that because `/` and `.` are illegal characters in fields names.

ikrommyd:
https://github.com/scikit-hep/coffea/blob/9f4d839bcd9c7b78e85069e79ea1ed35679c22e7/tests/test_nanoevents_rntuple.py#L106-L126
we currently xfail rntuple reading with treemaker but this is with the root-converted file.

ikrommyd:
> Actually, it's because `TreeMakerSchema` renames `/.fPt` into `pt`. For some reason that part is not done correctly for RNTuples. Also, I just realized that Uproot should not be letting you write that because `/` and `.` are illegal characters in fields names.

So is it an uproot bug? If you shouldn't write this format as RNTuple, we should indeed do what you're doing here and write TTree but we should probably also fix and test with `pytest.raises` that an error is being raised properly if you try to write RNTuple.

ariostas:
Uproot letting you write illegal names is a bug, but that's not the reason why this test doesn't work. Something else in the renaming logic of `TreeMakerSchema` is not working. But I guess after the Uproot bug is fixed, it won't even let you write it as an RNTuple unless you do the renamings beforehand.

ikrommyd:
Let's merge this one then since we should be writing TTree either way here.
However, is this worth an issue if something's up with treemaker schema here?

PR #1521, opened 2026-02-27
feat: correctionlib (JSON-POG) adapter classes for CorrectedJetsFactory

alefisico:
Adds drop-in adapter classes that wrap correctionlib (JSON-POG) corrections to work with the existing CorrectedJetsFactory — no changes to the factory itself.
                                                                                                      
  New classes:                                                                                        
  - CorrectionLibJEC — compound JEC correction
  - CorrectionLibJER — JER pt-resolution
  - CorrectionLibJERSF — JER scale factors (evaluates nom/up/down, returns (N,3) array)
  - CorrectionLibJUNC — JES uncertainty sources
  - CorrectionLibJECStack — drop-in replacement for JECStack, with from_file() convenience constructor

  Usage:
  ```
  from coffea.jetmet_tools import CorrectionLibJECStack, CorrectedJetsFactory

  jec_stack = CorrectionLibJECStack.from_file(
      "jet_jerc.json.gz",
      jec_tag="Summer24Prompt24_V2",
      data_type="MC",
      jet_type="AK4PFPuppi",
      jer_tag="Summer23BPixPrompt23_RunD_JRV1",
      unc_sources=["Regrouped_FlavorQCD", "Regrouped_AbsoluteMPFBias"],
  )

  name_map = jec_stack.blank_name_map
  name_map["JetPt"] = "pt"
  # ... fill remaining fields

  factory = CorrectedJetsFactory(name_map, jec_stack)
  corrected_jets = factory.build(jets)
```

  Tests:
  - Unit tests with synthetic JSON-POG data (eager, jagged, dask with optimization on/off, end-to-end
  factory)
  - Cross-validation tests using real Summer22_22Sep2023_V3_MC corrections: verifies that
  correctionlib adapters produce numerically identical results to the txt-based correctors
  (FactorizedJetCorrector, JetCorrectionUncertainty, JetResolution, JetResolutionScaleFactor) for JEC,
   JUNC, JER, and JERSF

### CODE REVIEW

**Comment by lgray** [comment]
A cursory pass over this looks really quite nice - thanks so much for this!

I need to dig into it and see what it's doing a bit more to see if anything's not quite what we'd like to be doing.

File: src/coffea/jetmet_tools/correctionlib_adapters.py [hunk 1]
```diff
@@ -0,0 +1,454 @@
...
+    def __init__(self, correction, input_names):
+        self._correction = correction
+        self._input_names = input_names
+
+    def __call__(self, *args):
+        if _is_typetracer(*args):
+            return _typetracer_1d(args[0])
+        np_args = [_to_flat_numpy(a) for a in args]
+        result = self._correction.evaluate(*np_args)
+        return awkward.Array(result.astype(numpy.float32))
```
nsmith-: Since `evaluate()` works the same for both, I think we can just have one implementation and use it for both Correction and CompoundCorrection types

src/coffea/jetmet_tools/correctionlib_adapters.py
```diff
@@ -0,0 +1,454 @@
...
```
nsmith-: Correctionlib already handles awkward v2 and dask-awkward arrays, see https://github.com/cms-nanoAOD/correctionlib/blob/f3f02eb2256cb6cd96ed8356c993c3e1c6ef3759/src/correctionlib/highlevel.py#L190 and the tests in https://github.com/cms-nanoAOD/correctionlib/blob/master/tests/test_highlevel.py#L71
For the backport you can keep these helpers, but for this PR we should just use `correction.evaluate()` directly

src/coffea/jetmet_tools/correctionlib_adapters.py
```diff
@@ -0,0 +1,454 @@
...
+    ----------
+    correction : correctionlib.Correction or correctionlib.CompoundCorrection
+        The correction object (e.g. ``cset.compound["...L1L2L3Res..."]``).
+    """
+
+    def __init__(self, correction):
+        self._correction = correction
+        self._signature = [inp.name for inp in correction.inputs]
+        self._is_compound = (
+            isinstance(correction, correctionlib.highlevel.CorrectionSet)
```
nsmith-: How would `correctionlib.highlevel.CorrectionSet` be used here?

src/coffea/jetmet_tools/correctionlib_adapters.py
```diff
@@ -0,0 +1,454 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _dispatch(func_cls, correction, input_names, kwargs):
+    """Route evaluation through ``dask_awkward.map_partitions`` when the first
+    kwarg value is a dask array, otherwise evaluate eagerly."""
+    first_arg = kwargs[input_names[0]]
+    args = tuple(kwargs[name] for name in input_names)
+
+    if isinstance(first_arg, dask_awkward.Array):
```
nsmith-: As above, the dask-awkward map_partitions call is handled by correctionlib already. For the backport, we don't have dask-awkward so it can just be dropped there.

src/coffea/jetmet_tools/correctionlib_adapters.py
```diff
@@ -0,0 +1,454 @@
...
+            "JetMass",
+            "JetPt",
+            "METpt",
+            "METphi",
+            "JetPhi",
+            "UnClusteredEnergyDeltaX",
+            "UnClusteredEnergyDeltaY",
+        }
+        if self._jec is not None:
+            for name in self._jec.signature:
```
nsmith-, line 197: So, in the JME correctionlib json, the names match what is above (e.g. `JetPt` etc.) ? In case they change it in the future, maybe we can check that at least `JetPt` is in the signature as a canary?

src/coffea/jetmet_tools/correctionlib_adapters.py
```diff
@@ -0,0 +1,454 @@
...
+            Compound correction level, e.g. ``"L1L2L3Res"``.
+        unc_sources : list[str], optional
+            Uncertainty source names, e.g.
+            ``["Regrouped_FlavorQCD", "Regrouped_AbsoluteMPFBias"]``.
+        jer_tag : str, optional
+            JER campaign tag.  If ``None``, JER/JERSF are not loaded.
+
+        Examples
+        --------
+        Build a correctionlib-backed JEC stack and apply it to jets::
```
nsmith-, line 243: FYI here is how this renders: https://coffea-hep--1521.org.readthedocs.build/en/1521/api/coffea.jetmet_tools.CorrectionLibJECStack.html#coffea.jetmet_tools.CorrectionLibJECStack.from_file

Looks good!

tests/samples/jet_jerc.json.gz
nsmith-, line 1: Is this a stripped down version of `jet_jerc_Summer22_V3.json.gz`? Any reason not to use the latter?

tests/test_correctionlib_adapters.py
```diff
@@ -0,0 +1,621 @@
...
+    Rho_jag = ak.unflatten(test_Rho, counts)
+
+    corr_jag = jec.getCorrection(JetA=A_jag, JetEta=eta_jag, JetPt=pt_jag, Rho=Rho_jag)
+    # Result is flat (1D) since adapters flatten internally
+    assert len(corr_jag) == len(test_pt)
+
+
+@pytest.mark.parametrize("optimization_enabled", [True, False])
+def test_correctionlib_jec_dask(clib_stack, flat_jet_arrays, optimization_enabled):
+    with dask.config.set({"awkward.optimization.enabled": optimization_enabled}):
```
nsmith-: For my education, what are these dask-awkward optimizations?

tests/test_jetmet_correctionlib_adapters.py
```diff
@@ -0,0 +1,621 @@
...
+    jets_dict = {
+        "pt": test_pt.astype(np.float32),
+        "mass": (test_pt * 0.1).astype(np.float32),
+        "eta": test_eta.astype(np.float32),
+        "phi": np.zeros(n_flat, dtype=np.float32),
+        "area": np.full(n_flat, 0.5, dtype=np.float32),
+        "pt_raw": test_pt.astype(np.float32),
+        "mass_raw": (test_pt * 0.1).astype(np.float32),
+        "rho": np.full(n_flat, 30.0, dtype=np.float32),
+        "pt_gen": (test_pt * 0.95).astype(np.float32),
```
nsmith-, line 275: Do we want to set some of these 0 to test the stochastic smearing for unmatched jets as well?
Actually, I guess that is one missing feature here: we are still using the coffea stochastic smearing strategy. We'd have to create a new correctionlib version of https://github.com/scikit-hep/coffea/blob/0be2a9ce6c9d989d9a4b0b555925ad6265812fd9/src/coffea/jetmet_tools/CorrectedJetsFactory.py#L61 
Maybe CorrectedJetsFactory can just check the JECStack type and then dispatch to a correctionlib `jer_smear` implementation?

nsmith- (changes requested):
This is a nice way of implementing only the minimal changes in migrating from txt to json!

src/coffea/jetmet_tools/correctionlib_adapters.py
ikrommyd: I think even for the backport, if we have new enough correctionlib it will work. What we should do here is make sure that we pin correctionlib to >= whatever version supports jagged arrays. And if we can have this lower pin even in the backport, let's just do it.

tests/test_correctionlib_adapters.py
ikrommyd: Dask graph optimizations (making it smaller, merging nodes etc) + necessary columns optimization I think. We have this in other tests. I don't think we should have this parametrization here though. It's fine to leave the optimization to default to not repeat the test twice.

src/coffea/jetmet_tools/correctionlib_adapters.py
ikrommyd: Typetracing and touching is only required if you wanna pass dask-awkward arrays through a function that is not inherently typetracable (it uses some function calls that are not part of the dask-awkward API). correctionlib takes care of that for you so you don't need any typetracing code branches. Just call `correction.evaluate(my dask-awkward array)` and it works fine. All of typetracing can be removed here making the PR even nicer.

src/coffea/jetmet_tools/correctionlib_adapters.py
nsmith-: correctionlib doesn't support awkward v1 jagged arrays, afaik

ikrommyd: Oh I thought it would be agnostic and just rewrap.

ikrommyd: You are right, it doesn't. And the error is also not good. It's two separate errors actually. It should probably check the version and if it's awkward1, just raise and prompt the user to flatten and unflatten. I'll open an issue
```py
In [14]: ptweight.to_evaluator().evaluate(ak.Array([[1,2,3],[],[4,5]]))
ValueError                                Traceback (most recent call last)
File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/correctionlib/highlevel.py:229, in Correction.evaluate(self, *args)
    228 try:
--> 229     vargs = [
    230         numpy.asarray(arg)
    231         for arg in args
    232         if not isinstance(arg, (str, int, float))
    233     ]
    234 except NotImplementedError:

File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/correctionlib/highlevel.py:230, in <listcomp>(.0)
    228 try:
    229     vargs = [
--> 230         numpy.asarray(arg)
    231         for arg in args
    232         if not isinstance(arg, (str, int, float))
    233     ]
    234 except NotImplementedError:

File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/awkward/highlevel.py:1351, in Array.__array__(self, *args, **kwargs)
   1327 """
   1328 Intercepts attempts to convert this Array into a NumPy array and
   1329 either performs a zero-copy conversion or raises an error.
   (...)   1349 cannot be sliced as dimensions.
   1350 """
-> 1351 return ak._connect._numpy.convert_to_array(self.layout, args, kwargs)

File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/awkward/_connect/_numpy.py:13, in convert_to_array(layout, args, kwargs)
     12 def convert_to_array(layout, args, kwargs):
---> 13     out = ak.operations.convert.to_numpy(layout, allow_missing=False)
     14     if args == () and kwargs == {}:

File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/awkward/operations/convert.py:324, in to_numpy(array, allow_missing)
    323 elif isinstance(array, ak._util.listtypes):
--> 324     return to_numpy(array.toRegularArray(), allow_missing=allow_missing)
    326 elif isinstance(array, ak._util.recordtypes):

ValueError: in ListOffsetArray64, cannot convert to RegularArray because subarray lengths are not regular

(https://github.com/scikit-hep/awkward-1.0/blob/1.10.3/src/cpu-kernels/awkward_ListOffsetArray_toRegularArray.cpp#L22)

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
Cell In[14], line 1
----> 1 ptweight.to_evaluator().evaluate(ak.Array([[1,2,3],[],[4,5]]))

File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/correctionlib/highlevel.py:240, in Correction.evaluate(self, *args)
    238 except (ValueError, TypeError):
    239     if any(str(type(arg)).startswith("<class 'awkward.") for arg in args):
--> 240         return _wrap_awkward(self._base.evalv, *args)  # type: ignore
    241     raise
    243 if vargs:

File ~/micromamba/envs/coffea-07x/lib/python3.11/site-packages/correctionlib/highlevel.py:131, in _wrap_awkward(func, *args)
    122 arg_indices = array_indices + non_array_indices
    124 tocall = partial(
    125     _call_as_numpy,
    126     func=func,  # type: ignore
    127     non_array_args=non_array_args,
    128     arg_indices=arg_indices,
    129 )
--> 131 return awkward.transform(tocall, *array_args)

AttributeError: module 'awkward' has no attribute 'transform'
```

tests/samples/jet_jerc.json.gz
alefisico: The `jet_jerc.json.gz` is a dummy file with simple constant corrections (e.g. JEC returns `1 + 0.01*JetA`, JER returns flat `0.1`, JERSF returns `nom=1.1/up=1.2/down=1.0`). This lets the unit tests assert exact expected values (like `np.float32(1.005)**4`).

The real `jet_jerc_Summer22_V3.json.gz` is used separately in the cross-validation tests (`test_crossval_*`) where we compare correctionlib vs txt-based correctors.

I can add a comment at the top of the test file clarifying this distinction, or if you'd prefer I can rework the unit tests to use the real file with approximate checks instead. Let me know which you prefer.

tests/test_jetmet_correctionlib_adapters.py
alefisico: I wanted to make another MR with this, because the current way of making random numbers `rand_gauss`  depend on the partitions/chunks, and if for whatever reason I change the chunks I cannot reproduce the smearing. In our group one of our students developed a hack using the event number, eta and phi, but I was going to push this after this is MR.

src/coffea/jetmet_tools/correctionlib_adapters.py
alefisico: ok, I am pushing new changes to take care of most of your comments. So if you dont see my reply of your comment, I implemented it already.

tests/samples/jet_jerc.json.gz
nsmith-: Ok I see, yeah just a comment at the top of the test file would help!

tests/test_jetmet_correctionlib_adapters.py
nsmith-: As of https://gitlab.cern.ch/cms-analysis-corrections/JME/examples/-/issues/1 we have `/cvmfs/cms-griddata.cern.ch/cat/metadata/JME/JER-Smearing/latest/jer_smear.json.gz` available that could be used for deterministic smearing. I think it would even be fine to include this path as a default value for a new parameter in `CorrectedJetsFactory` that controls the smearing behavior. But yes, we can leave that to a follow-up pull request.

ikrommyd:
Btw @alefisico, I notice you are doing `assert np.allclose` in testing and providing a custom error message. It's generally much better to use the `numpy.testing` module which has much better error messages so https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html in this case.
Generally every here is very useful https://numpy.org/doc/stable/reference/routines.testing.html

nsmith-:
Personally I'm a big fan of `array == pytest.approx(expected_array)`, it shows a nice debug

ikrommyd:
> Personally I'm a big fan of `array == pytest.approx(expected_array)`, it shows a nice debug

Oh does this work for arrays too? I'm generally using this for numbers and the `numpy.testing` things for arrays cause I know they show max relative and absolute difference and % of mismatched elements.

Edit:
Yeah okay they both kinda show the same
image
vs
```py
In [1]: import numpy as np

In [2]: x = np.array([1,2,3])

In [3]: y = np.array([1,2,4])

In [4]: np.testing.assert_allclose(x, y)
AssertionError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 np.testing.assert_allclose(x, y)

    [... skipping hidden 1 frame]

File ~/micromamba/envs/coffea-202x/lib/python3.13/site-packages/numpy/testing/_private/utils.py:919, in assert_array_compare(comparison, x, y, err_msg, verbose, header, precision, equal_nan, equal_inf, strict, names)
    914         err_msg += '\n' + '\n'.join(remarks)
    915         msg = build_err_msg([ox, oy], err_msg,
    916                             verbose=verbose, header=header,
    917                             names=names,
    918                             precision=precision)
--> 919         raise AssertionError(msg)
    920 except ValueError:
    921     import traceback

AssertionError:
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 3 (33.3%)
Max absolute difference among violations: 1
Max relative difference among violations: 0.25
 ACTUAL: array([1, 2, 3])
 DESIRED: array([1, 2, 4])
 ```

**Review by nsmith- — APPROVED** [review]

**Comment by ikrommyd** [comment]
Thanks @nsmith-. I'll give it a once-over too and merge.

File: tests/test_correctionlib_adapters.py [hunk 9]
```diff
@@ -0,0 +1,619 @@
...
+    from coffea.jetmet_tools import CorrectedJetsFactory
+
+    name_map = clib_stack.blank_name_map
+    name_map["JetPt"] = "pt"
+    name_map["JetMass"] = "mass"
+    name_map["JetEta"] = "eta"
+    name_map["JetA"] = "area"
+    name_map["ptRaw"] = "pt_raw"
+    name_map["massRaw"] = "mass_raw"
+    name_map["Rho"] = "rho"
```
ikrommyd: `rho` clashes with `pt` in scikit-hep/vector. You have to use something else like `Rho` for example like
```
name_map["Rho"] = "Rho"
```
and also adjust anywhere else

ikrommyd: I did it

ikrommyd (approved):
Did the rho -> Rho change and added `np.testing` to the tests. LGTM!

lgray:
Doing a final run through.

lgray (approved):

lgray:
@alefisico Please take the changes from this PR and apply them to the backported version as well. Looks good, thanks for doing this!

Issue #1513, opened 2026-01-26
Make final coffea 0.7 release


ikrommyd:
We should cherry-pick whatever we want from master into the backports branch and make a final coffea 0.7 release.

ikrommyd:
cc @valsdav I'll probably need you to test this one once a PR is up.

PR #1522, opened 2026-02-27
feat(backport): correctionlib (JSON-POG) adapter classes for CorrectedJetsFactory

alefisico:
Same as #1521 but backporting to 0.7.


ikrommyd:
@alefisico Thanks a lot for this work! Regarding the backport, do you feel like there is a need for it? i.e. do you have people wanting this? I was kinda hoping to not do a coffea 0.7 release ever again 🤣
See https://github.com/scikit-hep/coffea/issues/1513, I was just hoping to reach a "good final state" for coffea 0.7 and not release again. I thought that the current state is good but if you feel like we need this PR, we can consider adding it first and then doing final release.

alefisico:
I know many, including me, and all my group. I actually need this more than the one to master

ikrommyd:
Have you thought about not using an unmaintained package with awkward1 as a dependency which has memory leaks and bugs? 🤣

alefisico:
I implemented the changes from #1521 
Let me know if something else is needed.

ikrommyd:
Thanks! FYI lowercase `rho` is acceptable in coffea 0.7 since it's not using scikit-hep/vector. But it's good to be consistent with newer coffea so I've changed it to use `Rho` in the other test already present in the repo too.

ikrommyd:
I did a couple of minor tweaks and it looks good to me. I'm only thinking of one thing. In the "regular" jet met tools that coffea 0.7 already has, we manually utilize awkward 1s highlevel virtual arrays to lazify things `(ak.virtual)`. I'm wondering if we should do something similar here or let it compute everything on the fly like it does now. Right now the `lazy_cache` argument is just passed around to maintain API compatibility and it's not being used and we don't explicitly construct virtual arrays at all.

ikrommyd:
@lgray what do you think here? The old (non correctionlib) factory is lazy through `ak.virtual` although one can argue that the correctionlib one is simpler and does not need laziness. The counter argument is consistency I guess?

lgray:
@ikrommyd @alefisico Since the correctionlib one should be much simpler (many fewer function calls, typically one entry point), and there's still some UX benefit to using the virtual arrays it may be nice to implement a fairly shallow virtual array wrapper here. 

Given that we're trying to obsolete 0.7 we should present something decent but not make too much extra effort.

ikrommyd:
Yeah if it's easy enough we should just do it for consistency I think but we should definitely not waste time hunting down memory problems if it's not trivial.

alefisico:
ok, added `ak.virtual` wrappers for consistency with the existing factory.  The correctionlib path is simpler so the wrappers stay shallow and easy to follow. `lazy_cache` is now wired up properly.

ikrommyd:
@alefisico I pushed a commit https://github.com/scikit-hep/coffea/pull/1522/changes/4c78a5c6df0a84837e2c5ad403ad7e0f00ff35ce to always return virtual arrays for conssitency with the old jetmet tools https://github.com/scikit-hep/coffea/pull/1522
Could you please test it out? It's good to merge otherwise I think.

alefisico:
I tested it locally and I think it is good from my side. I can also add the test to this MR in the `test_jetmet_correctionlib_adapters.py` script if you think is needed.

ikrommyd:
Do you have a better pytest locally?

alefisico:
I quikcly test what you said with something like this:
```
def test_correctionlib_junc_with_lazy_cache(clib_stack, flat_jet_arrays):
    """getUncertainty must accept and use lazy_cache kwarg (new behavior after virtualisation)."""
    counts, _, test_eta, test_pt, test_Rho = flat_jet_arrays

    cache = cachetools.Cache(np.inf)
    uncs = clib_stack.junc.getUncertainty(
        JetEta=ak.Array(test_eta),
        JetPt=ak.Array(test_pt),
        lazy_cache=cache,
    )
    assert len(uncs) == 2
    for name, vals in uncs:
        arr = np.asarray(vals)
        assert arr.shape == (len(test_eta), 2)
        np.testing.assert_allclose(arr[:, 0], 1.02)
        np.testing.assert_allclose(arr[:, 1], 0.98)
    # Cache should have been populated after materialisation
    assert len(cache) > 0
```
which is:
* Passes lazy_cache=cache to getUncertainty (would have failed before the change)
* Materializes each result with np.asarray(vals) and checks the values are still correct
* Asserts len(cache) > 0 to confirm the cache was actually wired up and populated on access

But It might be not needed.

PR #1526, opened 2026-03-05
chore: update uproot pin and use https for logo in README

ikrommyd:


pyproject.toml
```diff
@@ -45,7 +45,7 @@ classifiers = [
 ]
 dependencies = [
   "awkward>=2.8.11",
-  "uproot>=5.6.0",
+  "uproot>=5.7.0",
```
nsmith-, line 48: This update is for what feature?

ikrommyd: Andres's recommendation due to various improvements.

ikrommyd: And we are bleeding edge awkward pin >= 2.8.11 anyways and since all of them are pure python and support >= 3.10 it shouldn't cause any installation trouble. Do you think there might be an issue? We can go >= 5.6.1 if we wanna be very conservative but 5.6.0 caused issues in LCG releases as LCG releases installed that unfortunately and there's some attribute missing in 5.6.0 that we use in coffea as if it exists.

nsmith-: Ok just curious

PR #1533, opened 2026-03-19
feat: adding split fileset strategies to coffea functionality

### LINKED ISSUE #1532: Adding split fileset strategies to coffea functionality
This feature would give users more control over how analysis gets executed inside the coffea package.
Possible use cases:
- To potentially get a partial result if the whole result wasn't obtained from the first run attempt, then rerun the analysis for missing chunks only
- To only run the analysis on, let's say, 10% of the data as a test/trial run
- To only run the analysis on a certain part of the fileset because there aren't enough workers available today for the whole dataset, save them and proceed tomorrow, etc.

Example of how analysis execution could look:

```python
chunks = split_fileset(fileset, strategy="by_dataset", percentage=20)
result = None
for chunk in chunks:
    try:
        output, metrics = run(chunk, processor_instance=Processor())
        if result is None:
            result = output
        else:
            result += output
    except BaseException as e:
        print(f"Error: {e}")
        continue
```
**Additional context**
Examples of Jupyter notebook analysis and more details can be found [here](https://github.com/hooloobooroodkoo/demo-split/tree/master)

hooloobooroodkoo:
Closes #1532

This PR adds two simple functions to the `coffea.util` module: `split_fileset` and `hash_fileset.`
The idea is to give users more control over how they execute analysis on a fileset, and to get a partial result instead of nothing if something breaks (e.g., one file is broken). If this PR is accepted and merged (or the alternative one I'm going to open next), I'll write documentation with a kind of "Best practices" guide on how to use it, with the examples below.

1. `split_fileset()` allows choosing a strategy for how to split the fileset into parts (~chunks, but higher-level chunks — not the usual coffea chunks; one chunk will be a unique subset of files from the fileset). It returns a list of partial filesets. This function accepts:
- `fileset`:    {dataset: {"files": {path: treename, ...}}}
- `strategy`:   "by_dataset" — one dataset per one chunk; None — all datasets together
- `percentage`: an integer that divides 100 evenly (20, 25, 50...). If `strategy="by_dataset"`, each dataset is split into `100/percentage` chunks; otherwise the whole fileset is split into `100/percentage` chunks where each chunk gets that percentage of each dataset's files
 - `datasets`: list, callable or tuple of dataset names
 
 This gives users the ability to write analysis like this:
 ```python
 chunks = split_fileset(fileset, strategy="by_dataset", percentage=20)
result = None
for chunk in chunks:
    try:
        output, metrics = run(chunk, processor_instance=Processor())
        if result is None:
            result = output
        else:
            result += output
    except BaseException as e:
        print(f"Error: {e}")
        continue
 ```
If one or several chunks contained a broken file etc., a partial result will still be returned.
 
 2. `hash_fileset` allows creating a unique filename for a processed chunk based on its file paths and dataset names. This is useful when you want to preserve partial results and only rerun analysis on missing chunks. The recommended approach in a Jupyter notebook, for example, would be:
 ```python
 # concept example
chunks = split_fileset(fileset, strategy="by_dataset", percentage=20)

import os

result = None
cache_dir = "./chunk_cache"
os.makedirs(cache_dir, exist_ok=True)
```
After the first run, partial result will be saved. Then user can just rerun the same cell and partial result will be extracted from cache, while processor run will be only applied to missing chunks.

```python
for chunk in chunks:
    chunk_hash = hash_fileset(chunk)
    cache_path = os.path.join(cache_dir, f"{chunk_hash}.coffea")

    if os.path.exists(cache_path):
        print(f"Loading cached result for chunk {chunk_hash}")
        output = load(cache_path)
    else:
        try:
            output, metrics = run(chunk, processor_instance=Processor())
            save(output, cache_path)
            print(f"Saved result for chunk {chunk_hash}")
        except BaseException as e:
            print(f"Error processing chunk {chunk_hash}: {e}")
            continue

    if result is None:
        result = output
    else:
        result += output
```

Usage examples with full analysis can be found [here](https://github.com/hooloobooroodkoo/demo-split/tree/master).
Thanks, @pfackeldey and @alexander-held for the idea!

### DIFF
--- a/src/coffea/util.py
+++ b/src/coffea/util.py
@@ -12,6 +12,8 @@ import cloudpickle
 import dask_awkward
 import fsspec
 import hist
+import math
+import json
 import numba
 import numpy
 import uproot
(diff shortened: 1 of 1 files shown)


pfackeldey:
I think it's either this PR or https://github.com/scikit-hep/coffea/pull/1534, but nice!

ikrommyd:
I'm thinking about where this should live. It's not dask-awkward so not in dataset tools but it's also not that generic to be in `coffea.util`. Maybe some new module inside processor? Generally the things that are about executing with virtual arrays are inside `coffea.processor`.

lgray:
dataset tools is a fine place for it to live. Dataset tools only uses dask-awkward to farm out the metadata harvesting. There are other tools in there to manipulate the resulting datasets as well.

PR #1534, opened 2026-03-19
feat: adding split_fileset functionality and Result output

### LINKED ISSUE #1532: Adding split fileset strategies to coffea functionality

hooloobooroodkoo:
Closes #1532

a. This PR adds two simple functions to the `coffea.dataset_tools` module: `split_fileset` and `hash_fileset.`
The idea is to give users more control over how they execute analysis on a fileset, and to get a partial result instead of nothing if something breaks (e.g., one file is broken). If this PR is accepted and merged (or the alternative one I'm going to open next), I'll write documentation with a kind of "Best practices" guide on how to use it, with the examples below.

1. `split_fileset()` allows choosing a strategy for how to split the fileset into parts (~chunks, but higher-level chunks — not the usual coffea chunks; one chunk will be a unique subset of files from the fileset). It returns a list of partial filesets. This function accepts:
- `fileset`:    {dataset: {"files": {path: treename, ...}}}
- `strategy`:   "by_dataset" — one dataset per one chunk; None — all datasets together
- `percentage`: an integer that divides 100 evenly (20, 25, 50...). If `strategy="by_dataset"`, each dataset is split into `100/percentage` chunks; otherwise the whole fileset is split into `100/percentage` chunks where each chunk gets that percentage of each dataset's files
 - `datasets`: list, callable or tuple of dataset names
 
 This gives users the ability to write analysis like this:
 ```python
 chunks = split_fileset(fileset, strategy="by_dataset", percentage=20)
result = None
for chunk in chunks:
    try:
        output, metrics = run(chunk, processor_instance=Processor())
        if result is None:
            result = output
        else:
            result += output
    except BaseException as e:
        print(f"Error: {e}")
        continue
 ```
If one or several chunks contained a broken file etc., a partial result will still be returned.
 
 2. `hash_fileset` allows creating a unique filename for a processed chunk based on its file paths and dataset names. This is useful when you want to preserve partial results and only rerun analysis on missing chunks. The recommended approach in a Jupyter notebook, for example, would be:
 ```python
 # concept example
chunks = split_fileset(fileset, strategy="by_dataset", percentage=20)

import os

result = None
cache_dir = "./chunk_cache"
os.makedirs(cache_dir, exist_ok=True)
```
After the first run, partial result will be saved. Then user can just rerun the same cell and partial result will be extracted from cache, while processor run will be only applied to missing chunks.

```python
for chunk in chunks:
    chunk_hash = hash_fileset(chunk)
    cache_path = os.path.join(cache_dir, f"{chunk_hash}.coffea")

    if os.path.exists(cache_path):
        print(f"Loading cached result for chunk {chunk_hash}")
        output = load(cache_path)
    else:
        try:
            output, metrics = run(chunk, processor_instance=Processor())
            save(output, cache_path)
            print(f"Saved result for chunk {chunk_hash}")
        except BaseException as e:
            print(f"Error processing chunk {chunk_hash}: {e}")
            continue

    if result is None:
        result = output
    else:
        result += output
```

b. This PR also introduces an optional Rust-inspired Result return type for `coffea.processor.Runner`. Instead of returning Accumulatable or raising an error in case of failure, it now can return `Result` object that can be either `Ok(Accumulateble)` or `Err(Exception)`. 

In the Runner user should specify a new flag `use_result_type`:

```python
run = processor.Runner(executor=executor,
                        schema=schemas.NanoAODSchema,
                        savemetrics=True,
                        use_result_type=True
                      )
```

Then the user can decide what to do with errors, and the code would look the following way:

```python
result = run(chunk, processor_instance=Processor())
if result.is_ok():
    output, metrics = result.unwrap()
else:
    # user can implement additional logic here
    print(f"Error: {result.exception}")
    continue
```

### New API
```python
from coffea.processor import Ok, Err, Result

result = run(chunk, processor_instance=Processor())
result.is_ok()       # bool
result.is_err()      # bool
result.unwrap()      # returns Accumulatable, or raises if Err
result.exception     # the exception, if Result is Err
```


ikrommyd:
I'd say we should probably proceed with this one instead of #1533 but make this new result type opt-in via a keyword argument in `Runner`

ikrommyd:
@hooloobooroodkoo do you feel like that the new notebook you added may fit better as a part of the processing notebook?

ikrommyd (changes requested):
There are a few things to do here:

1. It needs tests
2. If I'm reading it correctly, it curreclty only works with `__call__`. It should work with `run` too.
3. I don't think this plays well with `skipbadfiles` at the moment. We should perhaps block some combinations of both `result_type` and `skipbadfiles` both being true. Or maybe `result_type` requires `skipbadfiles` to be true and they somehow the result type only catches the exceptions defined by skipbadfiles.
4. The docstring style does not match other docstrings in the project
5. I would probably move the result type definitions inside `executor.py` since `WorkItem` and `FileMeta` are there too instead of having a specific `result.py` for that.
6. Perhaps the notebook should be part of the processing notebook and not a separate thing.

hooloobooroodkoo:
> @hooloobooroodkoo do you feel like that the new notebook you added may fit better as a part of the processing notebook?

I think it’s better to leave it as is, because this is a completely different example, and it would look a bit confusing if we tried to stick it into the processing and scaling example. Besides, it’s like an explanation of a completely independent, optional concept. If anyone is interested in getting a partial result, it might be easier to find it just by looking at the examples tab among the titles than to guess that it would be in `Processing and Scaling`.

hooloobooroodkoo:
@ikrommyd 
- [x] Tests are added: both `__call__` and `run()`, and the invalid skipbadfiles combination.
- [x] skipbadfiles resolved: I think the best way would be just to raise a ValueError when both use_result_type and skipbadfiles are True and inform the user that they are mutually exclusive (you either want to catch all types of errors per subset of fileset and, maybe, apply your own strategies to handle them; or you just choose to ignore these specific file errors and it's enough for you as a user. So the combination is blocked in __post_init__.
- [x] docstrings fixed
- [x] Result class moved to executor.py

ikrommyd:
@hooloobooroodkoo I have pushed multiple fixes and changes. Firs off I fixed a few issues regarding hashing because filesets can have different formats too. But more importantly I made `skipbadfiles` and `use_result_type` play differently together. So `use_result_type` is an extra step on top of `skipbadfiles`. So `use_result_type` needs `skipbadfiles`, otherwise an error is raised. Basically I make `skipbadfiles` specify the errors we want to catch and `use_result_type` specifies whether we want to silently ignore them (if it's False) or actually wrap them in the `Err` type. Let me know what you think.

hooloobooroodkoo:
@ikrommyd 
Thank you for making it much more advanced! The hashing fixes and the `_normalize_fileset` step in particular are really useful. I'll definitely take some of those improvements into coffea-workflow. Regarding  `skipbadfiles` + `use_result_type` it makes sense and doesn't change a lot from the perspective of coffea-workflow. So whatever works the best from coffea perspective is the best solution. I was just trying to keep it simple initially and mix it minimally with the existing coffea concepts

ikrommyd:
@hooloobooroodkoo are you happy with the current state and how it works? I can go ahead and merge it if that's the case. I just thought this implementation is probably the most "normal" behavior.

hooloobooroodkoo:
@ikrommyd yes, it works, go ahead!

ikrommyd (approved):

PR #1535, opened 2026-03-26
feat: counter-based RNG for deterministic JER smearing

alefisico:
## Summary

The current JER smearing in `CorrectedJetsFactory` uses `numpy.random`, which produces different random numbers depending on how events are partitioned across chunks. This means the same jet can get different smearing corrections depending on the processing setup, making results non-reproducible.

This PR adds an **opt-in** counter-based RNG (CBRNG) using the Squares algorithm ([arXiv:2004.06278](https://arxiv.org/abs/2004.06278)) that generates deterministic random numbers based on per-jet physics quantities (event number, phi, eta). The same jet always gets the same random number regardless of partitioning, chunk size, or parallelization strategy. 

This developement was created by @chuyuanliu in our framework, and I am just putting it in the central coffea because I consider it important for other users.

**Key points:**
- New `event_number` parameter on `build()` — when provided, uses CBRNG; when omitted (`None`), falls back to the legacy RNG. Fully backwards compatible.
- Per-jet 128-bit counter: `[event_number(64), phi_bits(32) << 32 | eta_bits(32)]`
- Works with both eager awkward arrays and dask-awkward

## Files changed
- `src/coffea/jetmet_tools/CorrectedJetsFactory.py` — added `event_number`/`seeds` parameters to `build()`, CBRNG path alongside legacy RNG
- `src/coffea/jetmet_tools/cbrng.py` — new module implementing the Squares CBRNG algorithm
- `tests/test_jetmet_correctionlib_adapters.py` — added `test_corrected_jets_factory_cbrng` verifying determinism and partition independence

## Test plan
- [x] `test_corrected_jets_factory_cbrng` — verifies same jets produce same smearing across calls
- [x] All 17 existing tests pass unchanged (no regressions)

### DIFF
--- a/src/coffea/jetmet_tools/CorrectedJetsFactory.py
+++ b/src/coffea/jetmet_tools/CorrectedJetsFactory.py
@@ -49,8 +49,7 @@ def rand_gauss(event_number, phi, eta, rng):
     # Build 128-bit counter: [event_number(64), phi_bits(32) << 32 | eta_bits(32)]
     counter = numpy.empty((len(phi_arr), 2), dtype=numpy.uint64)
     counter[:, 0] = event_number
-    counter[:, 1] = numpy.round(phi_arr, 3).view(numpy.uint32).astype(numpy.uint64)
-    counter[:, 1] <<= 32
+    counter[:, 1] = numpy.round(phi_arr, 3).view(numpy.uint32).astype(numpy.uint64).byteswap()
     counter[:, 1] |= numpy.round(eta_arr, 3).view(numpy.uint32).astype(numpy.uint64)
 
     rand = rng.normal(counter).astype(numpy.float32)
--- a/src/coffea/jetmet_tools/CorrectedJetsFactory.py
+++ b/src/coffea/jetmet_tools/CorrectedJetsFactory.py
@@ -49,7 +49,9 @@ def rand_gauss(event_number, phi, eta, rng):
     # Build 128-bit counter: [event_number(64), phi_bits(32) << 32 | eta_bits(32)]
     counter = numpy.empty((len(phi_arr), 2), dtype=numpy.uint64)
     counter[:, 0] = event_number
-    counter[:, 1] = numpy.round(phi_arr, 3).view(numpy.uint32).astype(numpy.uint64).byteswap()
+    counter[:, 1] = (
+        numpy.round(phi_arr, 3).view(numpy.uint32).astype(numpy.uint64).byteswap()
+    )
     counter[:, 1] |= numpy.round(eta_arr, 3).view(numpy.uint32).astype(numpy.uint64)
 
     rand = rng.normal(counter).astype(numpy.float32)
@@ -334,7 +336,9 @@ class CorrectedJetsFactory:
             if event_number is not None:
                 # Counter-based RNG: deterministic per-jet, partition-independent
                 flat_event = awkward.flatten(
-                    awkward.broadcast_arrays(event_number, jets[self.name_map["JetPt"]])[0]
+                    awkward.broadcast_arrays(
+                        event_number, jets[self.name_map["JetPt"]]
+                    )[0]
                 )
                 rng = Squares(*seeds)
                 out_dict["jet_resolution_rand_gauss"] = maybe_map_partitions(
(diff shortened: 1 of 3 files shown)


src/coffea/jetmet_tools/CorrectedJetsFactory.py
```diff
@@ -29,7 +30,41 @@ def __call__(self, array, like_what):
...
+    )
+    phi_arr = numpy.asarray(
+        awkward.typetracer.length_zero_if_typetracer(phi), dtype=numpy.float32
+    )
+    eta_arr = numpy.asarray(
+        awkward.typetracer.length_zero_if_typetracer(eta), dtype=numpy.float32
+    )
+
+    # Build 128-bit counter: [event_number(64), phi_bits(32) << 32 | eta_bits(32)]
+    counter = numpy.empty((len(phi_arr), 2), dtype=numpy.uint64)
```
lgray, line 50: Not sure if this matters but you could implement building the counter this in fewer operations using array.byteswap() (it inverts the phi bits, but that doesn't matter much).

something like:
```python3
counter[:, 0] = event_number
counter[:, 1] = numpy.round(phi_arr, 3).view(numpy.uint32).astype(numpy.uint64).byteswap()
counter[:, 1] |= numpy.round(eta_arr, 3).view(numpy.uint32).astype(numpy.uint64)
```
I'm unsure if there's any major performance benefit or drawback to doing it this way, mostly just fewer lines eaten by the interpreter.

lgray:
Looks like I need to fix the macos CI environment.

ikrommyd:
This might be a dumb question, but can't you just seed already available PRNGs (numpy for example) with physics quantities so that you get deterministic random numbers?

chuyuanliu:
> This might be a dumb question, but can't you just seed already available PRNGs (numpy for example) with physics quantities so that you get deterministic random numbers?

This is actually a great question. When using stateful standard PRNGs (like MT19937 or PCG64), you can either create an instance for each partition or for each event/object.

In the former case, the seed will be partition-dependent, and so is the random sequence.

In the latter case, creating a huge number of RNG instances with different seeds is not vectorized in `numpy`, so you would have to introduce a Python-level `for` loop over the objects. Though, in practice, the quality of the random sequence generated by changing the seed and getting the first value is usually okay, the statistical properties are only guaranteed along a sequence generated from a single seed. In the worst case, it may introduce unexpected correlations. CBRNGs are designed to solve these exact issues.

Note that although `numpy` has an implementation of CBRNGs like `Philox`, the API only supports being used in the same schema as other stateful PRNGs (generating the next n values instead of mapping specific counters to values).

ikrommyd:
Well had I had in mind is that you create one PRNG instace per chunk and you seed it with a number coming from the first event of the chunk. It would be chunk-dependent yes but you won't need a loop over the events for PRNGs

NJManganelli:
Just a clarification, the HashPRNG node in correctionlib (e.g. `/cvmfs/cms-griddata.cern.ch/cat/metadata/JME/JER-Smearing/2025-11-03/jer_smear.json.gz`) doesn't fulfill the needs here because it falls short in what ways? Does it have this architectural dependency where MacOS (I guess Arm64 + compilers) and x86 + linux etc have different outputs?

ikrommyd:
The `stdnormal` normal should be good across architectures (unlike `normal` and that's why it was deprecated). I think you can make a correction object on the fly and use that one here no? It should hopefully be much faster too.

chuyuanliu:
Yes, I did a quick benchmark. In this setup, the `correctionlib` `HashPRNG` implementation is about 2x–4x faster depending on the input size. My current `numpy` based CBRNG implementation seems to hit memory bottleneck for larger inputs.

Although a counter-based construction could be statistically cleaner than reseeding `pcg32_oneseq` and taking the first generated value each time. That said, with the inputs first mixed by `XXH64`, this approach is probably good enough in practice for this use case.

So I’m fine with closing this PR and preparing a follow-up that uses `correctionlib` instead.

ikrommyd:
Yeah so given what you're saying, using `correctionlib` would be best I think. You don't have to close the PR, you casn just update but if you think a new PR is cleaner, I'm fine. Your call.

chuyuanliu:
I would prefer a new PR to give a cleaner commit history.

ikrommyd:
We squash merge anyways so it's gonna be one commit on master but whatever you prefer.

PR #1536, opened 2026-03-26
feat(backport): counter-based RNG for deterministic JER smearing

alefisico:
## Summary

The current JER smearing in `CorrectedJetsFactory` uses `numpy.random`, which produces different random numbers depending on how events are partitioned across chunks. This means the same jet can get different smearing corrections depending on the processing setup, making results non-reproducible.

This PR adds an **opt-in** counter-based RNG (CBRNG) using the Squares algorithm ([arXiv:2004.06278](https://arxiv.org/abs/2004.06278)) that generates deterministic random numbers based on per-jet physics quantities (event number, phi, eta). The same jet always gets the same random number regardless of partitioning, chunk size, or parallelization strategy.

This is the v0.7.x backport of the same feature on master #1535 . The key difference is that CBRNG calls are executed **eagerly** with `numpy.asarray()` instead of being wrapped in `awkward.virtual`, since v1's virtual arrays can produce stale data on sliced arrays.

This developement was created by @chuyuanliu in our framework, and I am just putting it in the central coffea because I consider it important for other users.

**Key points:**
- New `event_number` parameter on `build()` — when provided, uses CBRNG; when omitted (`None`), falls back to the legacy RNG. Fully backwards compatible.
- Per-jet 128-bit counter: `[event_number(64), phi_bits(32) << 32 | eta_bits(32)]`
- Eager evaluation to avoid awkward v1 `virtual` pitfalls

## Files changed
- `coffea/jetmet_tools/CorrectedJetsFactory.py` — added `event_number`/`seeds` parameters to `build()`, CBRNG path alongside legacy RNG
- `coffea/jetmet_tools/cbrng.py` — new module implementing the Squares CBRNG algorithm
- `coffea/jetmet_tools/correctionlib_adapters.py` — minor fix
- `tests/test_correctionlib_adapters.py` — added `test_corrected_jets_factory_cbrng` verifying determinism and partition independence

## Test plan
- [x] `test_corrected_jets_factory_cbrng` — verifies same jets produce same smearing across calls
- [x] All 16 existing tests pass unchanged (no regressions)


alefisico:
**Note:** This PR depends on #1522 (correctionlib adapters backport). The diff currently includes changes from that branch — once #1522 is merged into `backports-v0.7.x`, this PR will automatically show only the CBRNG commit.

PR #1539, opened 2026-03-30
ci: install `libomp` in macos ci

ikrommyd:


ikrommyd:
I made `OMP_THREAD_LIMIT` work on windows too by using the `env` block but why did you want to run this single-threaded in the first place @lgray? I don't care either way I just didn't see any harm from multiple threads in the runners.

lgray:
I'm trying to get a cleaner trace, the failure from multiple threads was unintelligible.

ikrommyd:
Oh now ci hits ` Fatal Python error: Aborted`. Is this what you wanted to understand?

ikrommyd:
Feels related to https://github.com/dmlc/xgboost/issues/11500 although limiting openmp threads does not fix it and the crash is a bit different. It does feel like a general openmp conflict though.

We can switch the ci environment to a conda-forge setup which is nicely packages and does not need a system-wide openmp installation. This should fix the problem

lgray:
Yeah, that's what I was thinking too. That's pretty annoying. Thanks for digging!

lgray:
Ah, that means we should be able to successfully test if we just test the xboost one separate from everyone else.

Let's try that first, it is much less work.

lgray:
Ah, you already did it. Hah. Out of curiosity let's see if this works.

lgray:
Nope - the test is angry. Let's go with conda.

PR #1540, opened 2026-03-31
ci: use conda for CI testing setup

ikrommyd:


lgray (approved):

lgray:
Tests look fine to me, I think we're good to go. I don't like that we're more dependent on conda, I'd prefer that we have some tests that are bare pypi.... It's close enough though.

lgray:
@ikrommyd anything else you wanna do here? I'm otherwise fine to merge.

ikrommyd:
Uhmm now that it works I had two more ideas. We can use `uv` instead of `pip` in conda to make the pip installation faster.
Additionally, I don't like that it compiles `grpcio` from source in CI. This may have always been the case though, I don't know if the conda CI somehow changed something.

lgray:
Is the compilation on all archs or just a few?

lgray:
I'm fine with uv instead of pip. It's kind of a micro-optimization compared to the test volume though.

ikrommyd:
> Is the compilation on all archs or just a few?

From the brief look I took, it's installing an old version of grpcio in python 3.14 which doesn't have 3.14 wheels. I don't know who needs grpcio though.

ikrommyd:
Oh it's due to `tritonclient[grpc,http]` that we request
```
Collecting grpcio<1.68,>=1.63.0 (from tritonclient[grpc,http]>=2.56.0; extra == "triton"->coffea==2025.12.1.dev21+g487c3ad66)
  Downloading grpcio-1.67.1.tar.gz (12.6 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 12.6/12.6 MB 144.4 MB/s  0:00:00
```
It has an upper pin that doesn't let it install newer `grpcio` versions which would have wheels. So it has been happening this whole time and has nothing to do with this PR.

lgray:
Opened an issue over on the tritonclient github.

ikrommyd:
Cool. I switched to `uv` here. It's a bit annoying that we have to have envs where half the packages are from pypi and half are from conda-forge but it's better if we want to test with pytorch and xgboost together. This is good to go from my side.

lgray:
FWIW an ideal situation would be to test conda and pypi separately and entirely within those package managers to the extent possible. This will do for now.

PR #1541, opened 2026-04-01
ci: start testing 3.14t

ikrommyd:
Numba 0.65.0 is out with free-threaded support/wheels.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,17 +72,20 @@ jobs:
     - name: Install coffea and dependencies (Linux)
       if: startsWith( matrix.os, 'ubuntu' )
       run: |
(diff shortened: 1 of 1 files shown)


ikrommyd:
Conda forge pytorch does not have 3.14t builds and tornado, lz4, and numcodecs need to be built from source in 3.14t but I think we can start testing. In the future we can try running the tests in parallel too.

lgray:
Excellent - this is a great result already!

lgray:
Triton has no wheels for 3.14t and won't build?

ikrommyd:
> Triton has no wheels for 3.14t and won't build?

Grpcio which is required by triton won’t build, not triton itself. It’s a known issue. It uses the GIL and they have done no effort to support free threaded python yet.

PR #1545, opened 2026-04-06
feat: add `uproot_options` in preprocessing

### LINKED ISSUE #1544: Propagate uproot_options to preprocess
In order to open files from presigned S3 URLs, `uproot.open()` needs `encoded=True` argument passed. This is possible in most places in coffea via `uproot_options`, except in `metadata_fetcher_root` here:
https://github.com/scikit-hep/coffea/blob/954f23350d93a2ae068fc335db3068ef27009b67/src/coffea/processor/executor.py#L1234

It would be great to also propagate the `uproot_options` there via `Runner.preprocess` to allow metadata extractions from files stored in S3.

Thanks!

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1544

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1229,9 +1229,14 @@ class Runner:
 
     @staticmethod
     def metadata_fetcher_root(
-        xrootdtimeout: int, align_clusters: bool, item: FileMeta
+        xrootdtimeout: int,
+        align_clusters: bool,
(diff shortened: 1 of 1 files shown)


ikrommyd:
@kondratyevd could you take a look if that's what you need?

kondratyevd:
Yes, that's what I had in mind. Thanks!

PR #1547, opened 2026-04-16
ci: move dask client parametrization inside the job

ikrommyd:
I think we can benefit from having less jobs for the github runners to run as the runners are limited.


ikrommyd:
I'm not sure if this is good or bad. It is slower if all runners run at the same time but often I see that the runners only run a few jobs at a time. In this case it's faster if we have less jobs.

ikrommyd:
I wanna try this thing and check it out for a bit. Because ci is often very slow when github can't get you runners quickly.
Can be reverted very easily if it's slower in the long run.

PR #1548, opened 2026-04-22
feat: restore support for reading parquet file in executor workflows

MoAly98:
## Summary

This is mostly a legacy fix. It brings parquet support back to `Runner(format="parquet", ...)`, which had been stubbed out with `NotImplementedError` during the calver refactor. The executor pieces are essentially what coffea 0.7 shipped, adapted to the current `NanoEventsFactory.from_parquet` signature. There is also a small parquet-factory change to match the root side.

## Changes

- `src/coffea/processor/executor.py`: parquet is again a valid `format=` for `Runner`. File opening uses `ParquetFileContext`, work items go through `NanoEventsFactory.from_parquet(...)`, and the preprocess step runs file-existence checks so missing files raise `FileNotFoundError` at preprocess time (same as root). `chunksize` and `maxchunks` also apply to parquet now, as a side effect of sharing the preprocess path with root.
- `src/coffea/nanoevents/factory.py` and `src/coffea/nanoevents/mapping/parquet.py`: `factory.file_handle` now returns the underlying `pyarrow.parquet.ParquetFile` for parquet factories instead of `None`. Matches the root side added in #1502.
- Tests: `tests/test_local_executors.py::test_nanoevents_analysis` is un-xfailed for parquet. A small malformed parquet fixture (`tests/samples/nano_dy_malformed.parquet`) is added to exercise `skipbadfiles=True` for parquet, since parquet has no equivalent of root's `UprootMissTreeError` for the bad-file scenarios in that test.

## Minimal example

```python
import awkward as ak

from coffea import processor
from coffea.nanoevents import NanoAODSchema

run = processor.Runner(
    executor=processor.IterativeExecutor(),
    schema=NanoAODSchema,
    format="parquet",
)

def sum_muon_pt(events):
    return {"total_muon_pt": float(ak.sum(events.Muon.pt))}

out = run(
    {"DY": {"treename": "Events", "files": ["tests/samples/nano_dy.parquet"]}},
    processor_instance=sum_muon_pt,
    treename="Events",
)
print(out["total_muon_pt"])
```

_(PR summary retouched by claude)_


tests/test_local_executors.py
```diff
@@ -29,8 +29,8 @@ def test_nanoevents_analysis(
             mode=mode, check_filehandle=True
         ).process
 
-    if filetype == "parquet":
-        pytest.xfail("parquet nanoevents not supported yet")
+    if filetype == "parquet" and skipbadfiles is True:
```
ikrommyd: I'd rather have this in this PR. It should very simple to catch exceptions in executors.

MoAly98: yes, i added it as a to-do. I think I just need to cme up with an Exception that is not a ROOT specific one

ikrommyd: `skipbadfiles` I think defaults to `OSError` if it's `True` but it can also be a tuple of exceptions to catch.

ikrommyd:
I see windows failures in CI. It looks like this is conflicting with windows drive letters which are also split with `:`.
```
                    elif format == "parquet":
                        skyhook_options = {}
                        if ":" in item.filename:
>                           (
                                ceph_config_path,
                                ceph_data_pool,
                                filename,
                            ) = item.filename.split(":")
E                           ValueError: not enough values to unpack (expected 3, got 2)
```

tests/test_local_executors.py
MoAly98: yesm the question here is to come up with a non-OSError for the parquet case which would raise an error when `skipbadfiles=True`

MoAly98:
> I see windows failures in CI. It looks like this is conflicting with windows drive letters which are also split with `:`.
> 
> ```
>                     elif format == "parquet":
>                         skyhook_options = {}
>                         if ":" in item.filename:
> >                           (
>                                 ceph_config_path,
>                                 ceph_data_pool,
>                                 filename,
>                             ) = item.filename.split(":")
> E                           ValueError: not enough values to unpack (expected 3, got 2)
> ```

@ikrommyd interesting that this was not happening in 0.7, no?

tests/test_local_executors.py
```diff
@@ -29,8 +30,19 @@ def test_nanoevents_analysis(
...
+    bad_second_file = (
+        "tests/samples/nano_dy_malformed.parquet"
+        if filetype == "parquet"
+        else f"tests/samples/nano_dy_SpecialTree.{filetype}"
+    )
+    bad_only_file = (
+        "tests/samples/nano_dy_malformed.parquet"
+        if filetype == "parquet"
+        else f"tests/samples/nano_dy.{filetype}"
+    )
```
MoAly98, line 49: @ikrommyd i don't know how much i like this and if there is a better way to not have this logic... some error both root and parquet could be affected by which is not an OSError? could write a bad processor for this,  but feels like an overkill.

ikrommyd: i think it's alright...Also splitting the test in two instead of parametrizing is also fine I think. Whatever you think is cleaner

ikrommyd:
> > I see windows failures in CI. It looks like this is conflicting with windows drive letters which are also split with `:`.
> > ```
> >                     elif format == "parquet":
> >                         skyhook_options = {}
> >                         if ":" in item.filename:
> > >                           (
> >                                 ceph_config_path,
> >                                 ceph_data_pool,
> >                                 filename,
> >                             ) = item.filename.split(":")
> > E                           ValueError: not enough values to unpack (expected 3, got 2)
> > ```
> 
> @ikrommyd interesting that this was not happening in 0.7, no?

yeah that's odd, CI passed last time I did a backport PR

lgray:
The logic around skyhook (which we may want to drop anyway) is very different in coffea 0.7:
https://github.com/scikit-hep/coffea/blob/backports-v0.7.x/coffea/nanoevents/factory.py#L230

I'd honestly be OK with dropping it, that R&D direction never really converged in a way compatible with HEP.

ikrommyd:
> The logic around skyhook (which we may want to drop anyway) is very different in coffea 0.7: https://github.com/scikit-hep/coffea/blob/backports-v0.7.x/coffea/nanoevents/factory.py#L230
> 
> I'd honestly be OK with dropping it, that R&D direction never really converged in a way compatible with HEP.

yeah that's reasonable. I'll do a PR towards master to remove skyhook.

ikrommyd:
Let's do https://github.com/scikit-hep/coffea/pull/1549 and rebase this one right after.

MoAly98:
ok @ikrommyd @nsmith-  I will just then wait for you to merge #1549 and approve this PR after I rebase :) Would be great if we can get a minor release afterwards to allow us to use this for the integration challenge.

ikrommyd:
@MoAly98 skyhook has been removed in master and I've rebased the PR branch here. Please continue with the PR. You don't need to care about skyhook at all in your changes.

ikrommyd:
I have improved the testing a bit.

Can we please move the `to_flat_columns` part outside of this PR and do it separately? I would like think about it a bit more and I would also like it to work for dask mode too. It's easily possible https://github.com/scikit-hep/coffea/discussions/1100. It currently just returns an empty dict
```
In [5]: events = NanoEventsFactory.from_root({fname: "Events"}, schemaclass=NanoAODSchema, mode="dask").events()
/Users/iason/Dropbox/work/pyhep_dev/coffea/src/coffea/nanoevents/schemas/nanoaod.py:289: RuntimeWarning: Missing cross-reference index for LowPtElectron_electronIdx => Electron
  warnings.warn(
/Users/iason/Dropbox/work/pyhep_dev/coffea/src/coffea/nanoevents/schemas/nanoaod.py:289: RuntimeWarning: Missing cross-reference index for LowPtElectron_photonIdx => Photon
  warnings.warn(
/Users/iason/Dropbox/work/pyhep_dev/coffea/src/coffea/nanoevents/schemas/nanoaod.py:328: RuntimeWarning: Branch Photon_mass already exists but its values will be replaced with 0.0
  warnings.warn(
/Users/iason/Dropbox/work/pyhep_dev/coffea/src/coffea/nanoevents/schemas/nanoaod.py:328: RuntimeWarning: Branch Photon_charge already exists but its values will be replaced with 0.0
  warnings.warn(

In [6]: NanoAODSchema.to_flat_columns(events)
Out[6]: {}
```

MoAly98:
@ikrommyd thanks for improving the testing! I moved the column flattening to #1557 :)

ikrommyd:
I added some better error messages and I should have made checkpointing work with parquet too.

Thanks for this!

ikrommyd (approved):
LGTM! Should be good to merge if CI is green. Thanks!

PR #1550, opened 2026-04-23
feat: awkward validation hooks to ensure the right fields are present for vector behaviors to work

### LINKED ISSUE #1473: Add behavior validation hook to ensure that all the right fields are present

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1473


pfackeldey (approved):
the validation part looks good to me! 
We briefly discussed offline that there's no performance regression adding those, even though they run pretty often. Can you confirm this is still the case, or do we need to rethink something here?

ikrommyd:
```py
import timeit
import numpy as np
import awkward as ak
from coffea.nanoevents.methods import vector

data = {"pt": np.array([1.0]), "eta": np.array([0.5]),
        "phi": np.array([0.1]), "mass": np.array([0.0])}
for i in range(46):
    data[f"extra{i}"] = np.array([float(i)])

arr = ak.zip(data, with_name="PtEtaPhiMLorentzVector", behavior=vector.behavior)

if hasattr(arr, "__awkward_validation__"):
    n = 10_000
    t = timeit.timeit(arr.__awkward_validation__, number=n)
    print(f"validation: {t/n*1e6:.2f} µs/call ({t*1000:.0f} ms for {n})")
else:
    print("no validation hook on this branch")
```
this gives me `validation: 1.28 µs/call (13 ms for 10000)` so I think we're good

ikrommyd:
There is a chance that the testing did not catch all the bugs so I'm going to merge this. I'd like to beta test it in my analysis for a bit and through the integration tests before we make a coffea release (which should be before CHEP).

PR #1557, opened 2026-05-06
feat: add `uproot_writeable` schema method for writing events to columnar files for NanoAOD and Base schemas

MoAly98:
Adds a uproot_writeable(events) classmethod to BaseSchema and NanoAODSchema — the inverse of what each schema does at construction. Mirrors TreeMakerSchema.uproot_writeable
  (see [discussion #735](https://github.com/scikit-hep/coffea/discussions/735#discussioncomment-9646917)) so the typical skim workflow is:

  with uproot.recreate(out_path) as f:
      f.mktree("Events", NanoAODSchema.uproot_writeable(events))

  Eager/virtual only.


MoAly98:
@ikrommyd i followed the discussion to support the dask mode and quick checks work. Let me know if thats what you would want.

src/coffea/nanoevents/schemas/base.py
```diff
@@ -131,3 +153,15 @@ def behavior(cls):
...
+    def to_flat_columns(events):
+        """Deconstruct ``events`` into a flat ``{branch: array}`` dict.
+
+        Must be overridden per schema to invert the schema-specific zipping
+        applied during construction. Output is intended for columnar writers
+        (``ak.to_parquet``, ``uproot.TTree``).
+        """
+        raise NotImplementedError(
+            "to_flat_columns is not implemented for this schema; override it on the subclass"
+        )
```
ikrommyd: I think we should make this work for base schema too (but guard subclasses). It can just spit back the columns one by one with no restructuring as it's already like a dict of field -> column.

ikrommyd:
Oh and this doesn't even have to be a schema method I think. It can just be a function that accepts `events` only as you can directly get the schema like this
```py
In [6]: events.attrs["@events_factory"]
Out[6]: <coffea.nanoevents.factory.NanoEventsFactory at 0x11f01b4d0>

In [7]: events.attrs["@events_factory"]._schema
Out[7]: <coffea.nanoevents.schemas.nanoaod.NanoAODSchema at 0x11f01b0e0>
```

Do you think a function is better instead of a schema method?

Edit:
I noticed that you don't have this in the dask case but you'd just need to add the factory to the metadata here: https://github.com/scikit-hep/coffea/blob/3dd78a82ac9ea93f08e25efa289a8a63692606ea/src/coffea/nanoevents/factory.py#L767

Edit 2:
It may actually be bad or not even possible to add the factory to the attrs in the dask case because we want arrays to be picklable fast.

ikrommyd:
@lgray Is that a reason to want the dask awkward case attrs to be minimal? Cause they would be pickled and unpickled to share arrays between workers so adding the factory there too makes pickling and unpickling heavier. I guess it might just be nothing compared to pickling the array anyways.

ikrommyd:
the dask case is painfully slow to write. The graph is insanely big and takes minutes to optimize. Can reproduce this with coffea 2024 even and all the depencneis from that time. We will not fix it here. I will make it eager only.

I also renamed the method to `uproot_writeable` as the treemakerschema already had such a method.

PR #1558, opened 2026-05-06
feat: make `dask`, `distributed`, `dask-awkward`, and `dask-histogram` optional dependencies

### LINKED ISSUE #1527: Make dask-awkward optional for coffea
Because of issues with [dask](https://github.com/dask-contrib/dask-awkward/issues/598), `dask-awkward` was pinned to use `dask<2025.4.0,>=2023.04.0`, and it makes almost impossible to upgrade any dask related dependency, e.g. `dask-distributed` without upgrading dask and after which it breaks everything.

It would work if we can make the dependency optional:

```
pip install coffea[awkward-dask]
```

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1527

We make dask awkward and dask histogram optional. `isinstance` checks are move into a helper `_isinstance`, for map partitiions we use a helper `maybe_map_partitions` or the guarded import function, imports are guarded with a nice error message, and we vendor `dask_property` and `dask_method` from the dask awkward source code.

Disclaimer: This was mostly generated with Claude Code and Codex in multiple passes of editing and reviewing.


pyproject.toml
```diff
@@ -47,8 +47,8 @@ dependencies = [
   "awkward>=2.8.11",
   "uproot>=5.7.0",
   "dask[array]>=2024.3.0",
-  "dask-awkward>=2025.9.0",
-  "dask-histogram>=2025.2.0",
+  "distributed>=2024.3.0",
```
lgray: dask itself doesn't need to be a core dependency of coffea.

pyproject.toml
```diff
@@ -81,9 +81,8 @@ Homepage = "https://github.com/coffeateam/coffea"
 
 [project.optional-dependencies]
 dask = [
-  "distributed>=2024.3.0",
-  "bokeh!=3.0.*,>=2.4.2",
-  "blosc2",
+  "dask-awkward>=2025.9.0",
+  "dask-histogram>=2025.2.0",
```
lgray, line 87: I would recommend, instead, that we have optional dependencies `dask` and then `dask_awkward`. `dask` was supposed to mean the which scheduler you're using, similar to the `parsl` optional dep.

pyproject.toml
ikrommyd: Yeah that would take more import checks. Alright. I'm writing up CI run now modifying the tests to be able to test everything in CI

lgray (approved):

ikrommyd:
@lgray Thanks! I'll wait a bit with this and then put in a release (probably in June). I'll update the integration testing repo and perhaps ask some people to beta test this by just uninstalling dask-awkward from their envs. We just want to see we didn't break anyone that doesn't use any dask-awkward functionality before releasing.

PR #1563, opened 2026-05-19
fix: handle empty arrays for buffer cache

lauridsj:
This implements a workaround for an error observed in `numcodecs` by @pfackeldey , https://github.com/zarr-developers/numcodecs/issues/831 . When compressing and decompressing empty numpy arrays, an error is thrown for unknown reasons. This works around the issue by only compressing arrays with more than 16 bytes (proposed by @pfackeldey ).

### DIFF
--- a/src/coffea/nanoevents/mapping/buffer_cache.py
+++ b/src/coffea/nanoevents/mapping/buffer_cache.py
@@ -10,6 +10,7 @@ class ShapeDTypeStruct:
     dtype: np.dtype
     shape: tuple[int, ...]
     strides: tuple[int, ...]
+    compressed: bool
 
 
 ByteBuffer: tp.TypeAlias = bytes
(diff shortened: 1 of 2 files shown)


pfackeldey:
@lauridsj thanks! this looks good to me :) 

@ikrommyd the CI failure seems to be unrelated - any idea what's going on?

ikrommyd:
Yeah, micromamba latest release is broken on windows. Should be fixed now with the pin I added on master.

ikrommyd (commented):
Why is the check on the number of bytes and now simply on whether the array is empty or not? If it's a one element array, it still doesn't work?

ikrommyd:
Perhaps worth a very short test in `tests/test_buffer_cache.py`

pfackeldey:
> Why is the check on the number of bytes and now simply on whether the array is empty or not? If it's a one element array, it still doesn't work?

it's at the same time a tiny performance benefit to avoid calling into a codec for so few data that you can't really compress / benefit from compression anyways.

ikrommyd (changes requested):
Just add a small test that shows we can indeed add empty arrays to such a cache and we should be good.

ikrommyd (approved):

lauridsj:
Thanks a lot for adding the test & merging, and sorry for being offline for a few days!

PR #1567, opened 2026-06-01
docs: fix various typos and old links

ariostas:
I noticed that there are multiple places in the docs where there are typos that cause the docs to be rendered weirdly. For example, look at the `preload` parameter [here](https://coffea-hep.readthedocs.io/en/latest/api/coffea.nanoevents.NanoEventsFactory.html#coffea.nanoevents.NanoEventsFactory.from_root).

I used Codex to go through it and fix typos and links. But I still need to go through the changes to make sure that they are correct (especially for links). I'll mark it as ready for review when I'm done.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@
 
 * Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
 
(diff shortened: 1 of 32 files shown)


ikrommyd:
A user found this old link in contributing.md too: https://github.com/scikit-hep/coffea/issues/1568#issuecomment-4650551145
Would be nice to fix it with this PR too.

ariostas:
I looked through the changes and fixed a few things. Now it's ready to go.

ikrommyd (approved):

Issue #1568, opened 2026-06-08
`torch_wrapper` does not do inference properly on GPU


EthanColbert8:
**Describe the bug**
When using an `ml_tools.torch_wrapper` instance to perform inference with a PyTorch model, the model itself is sent to GPU if `torch.cuda.is_available()`, but the input tensors are not, leading to a PyTorch error with tensors on different devices.

**To Reproduce**
1. Use analysis code such as the following (adapted from the processor example), paired with a simple MLP saved via TorchScript (that accepts 5 input features):
```
import logging
import numpy as np
import awkward as ak
from coffea import processor, ml_tools
from coffea.nanoevents import NanoEventsFactory, BaseSchema
from coffea.nanoevents.methods import candidate

class SimpleMLModule(ml_tools.torch_wrapper):
    def __init__(self, model_path):
        super().__init__(model_path)
    
    def prepare_awkward(self, events):
        muon = events.muon

        features = [
            muon.px,
            muon.py,
            muon.pz,
            muon.energy,
            muon.charge,
        ]

        inputs = {
            "x": ak.values_astype(ak.concatenate([x[:,np.newaxis] for x in features], axis=-1), "float32")
        }
        return (), inputs

class SimpleProcessor(processor.ProcessorABC):
    def __init__(self, mode="virtual"):
        assert mode in {"eager", "virtual", "dask"}
        self._mode = mode

        logging.debug("Instantiating Torch module.")
        self.model = SimpleMLModule("simple_mlp_torchscript.pt")
        logging.debug("Model instantiated.")
    
    def process(self, events):
        dataset = events.metadata["dataset"]
        muons = ak.zip(
            {
                "pt": events.Muon_pt,
                "eta": events.Muon_eta,
                "phi": events.Muon_phi,
                "mass": events.Muon_mass,
                "charge": events.Muon_charge,
            },
            with_name="PtEtaPhiMCandidate",
            behavior=candidate.behavior
        )

        cut = ak.num(muons) >= 1

        muon = muons[cut][:, 0]
        events = ak.with_field(events[cut], muon, "muon")

        logging.debug("Running ML model on events.")
        ml_fields = self.model(events)
        logging.debug(f"Model finished. Type of result: {ak.type(ml_fields)}")

        return {dataset: {"ml_stuff": ml_fields}}

    def postprocess(self, accumulator):
        pass

def main():
    filename = "Run2012B_DoubleMuParked.root"

    logging.debug("Loading events.")

    access_log = []
[...]

PR #1569, opened 2026-06-08
fix: send tensors to GPU if using it in `torch_wrapper`

EthanColbert8:
This PR addresses issue #1568 

Currently, the `torch_wrapper` class in `ml_tools` initializes Torch models on the GPU when loading them, but doesn't send input tensors there at inference time (see the issue). This PR simply sends the tensors to the GPU if the current device type is `"cuda"`.

Opening as a draft initially, until I fully test everything :)


ikrommyd:
Superseded by https://github.com/scikit-hep/coffea/pull/1570

PR #1573, opened 2026-06-25
fix: adapt to Uproot hiding AsGrouped branches in form

ariostas:
This is the companion PR to https://github.com/scikit-hep/uproot5/pull/1668.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -130,6 +130,9 @@ class _map_schema_uproot(_map_schema_base):
             },
             "form_key": None,
         }
+        typenames = form.parameters.get("typenames")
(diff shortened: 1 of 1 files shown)


mrzimu (commented):
It's good to me, and should we specify a new version of `uproot` in the requirement section of `pyproject.toml` ?

ariostas:
> should we specify a new version of uproot in the requirement section of pyproject.toml ?

The change is backwards compatible, so it's not needed

ikrommyd:
Let me know if we should just merge this one. Just wanna make sure you're not going to change anything from the uproot side that needs adjustments here.

ariostas:
@ikrommyd it's ready to go. I just merged the Uproot PR.

PR #1574, opened 2026-06-27
perf(tests): share one dask Client across the dask_client suite and parallelize it, remove useless test coverage

lgray:
- Updated latest commit to use a real compression level.

> [!WARNING]
> **DO NOT MERGE** — this is a maintenance/perf experiment opened for review and to exercise CI.

This PR speeds up **both** legs each test matrix job runs, without changing the code paths the suite exercises.

---

## Part 1 — the `dask_client` leg (startup-bound)

The `dask_client`-marked tests each spun up a **fresh** `distributed.Client()` (a new `LocalCluster` with worker processes), and CI ran them **serially** (`-m "dask_client"`, no `-n`). Across a leg that's ~20+ cluster startups back-to-back, and cluster spawn dominated the wall time.

- **`tests/conftest.py`**: new session-scoped, resource-bounded `dask_client` fixture (1 worker × 2 threads, `processes=True`). One cluster is reused per xdist worker instead of one per test.
- **19 `dask_client` tests** across 5 files now request the fixture and use `Client.as_current()` (or the default client) instead of constructing their own `Client()`. Tests that called `client.close()` are updated to leave lifecycle management to the fixture (the previous per-test `close()` would otherwise tear down the shared client mid-session). Dead `try: from distributed import Client` blocks and unused inline imports removed.
- **`.github/workflows/ci.yml`**: the `dask_client` leg now runs under pytest-xdist (`-n 4`), matching the non-dask leg.
- **`pyproject.toml`**: registered the `dask_client` and `network` markers (silences `PytestUnknownMarkWarning`).

**Measured** (10-core box, py3.14, `distributed` 2025.3.1), `-m dask_client`, **26 passed / 4 skipped** every config:

| Config | Wall | Speedup |
|---|---|---|
| Baseline: per-test `Client()`, serial | 169.5 s | 1.00× |
| Shared session client, serial | 138.4 s | 1.22× |
| Shared session client, `-n 4` | **69.3 s** | **2.45×** |

---

## Part 2 — the `not dask_client` leg (compute-bound)

This leg already runs `-n 4`. Profiling showed it is **CPU-bound and well balanced** (the dominant file, `test_local_executors.py` = 191 s of the 314 s total, runs at ~3.74 of 4 cores busy — only ~7% idle), so harness-level tricks (more workers, work-stealing) have little headroom. The win is removing genuinely redundant work.

`test_nanoevents_analysis` parametrized `compression` over `[None, 0, 2]` — a 3× multiplier on a 432-combination matrix. coffea's executor only branches on `compression is None` (every check in `processor/executor.py` is `is None`/`is not None`, never truthy), so `0` and `2` run the **identical** coffea path — both forward the level straight to `lz4f` and differ only in lz4's internal ratio, which coffea never inspects. Collapsing to `[None, 0]` keeps both distinct code paths (no-compress + compress) and the fastest lz4 level.

- `test_local_executors.py`: **453 → 309 tests** (−144).

**Measured** at `-n 4`:

| Scope | Before | After | Speedup |
|---|---|---|---|
| `test_local_executors.py` (isolated) | 191.1 s | 134.5 s | −30% |
| full `-m "not dask_client"` leg | 314.0 s | 249.3 s | **−21%** (1089 passed, all green) |

---

## Combined impact (per matrix leg, local box)

| Leg | Before | After |
|---|---|---|
| `not dask_client` | 314 s | 249 s |
| `dask_client` | 169 s (serial) | 69 s (`-n 4`) |
| **Total** | **~484 s** | **~318 s (−34%)** |

…multiplied across the 12-leg `test` matrix.

## Same-code-path / non-vacuous notes

- The `dask_client` tests still go through the distributed scheduler with cross-process pickling; only the number/lifetime of clusters changed. A shared client surfaced a real latent coupling — 5 tests called `client.close()`, which with a shared client closed it for everyone (`Exception: Tried sending message after closing`); removing those is the correct fix, and the leg is green serially **and** at `-n 4`.
- The compression collapse drops zero coffea code paths (verified against the branch conditions); `compression=0` is distinct from `None` and exercises the compress/`lz4f` path. All 309 file-local and 1089 leg tests pass.

## Reviewer notes

- `-n 4` for the `dask_client` leg mirrors the existing non-dask leg; if the 4-vCPU runners show memory pressure from 4 concurrent bounded clusters, dial to `-n 2` with no other change.
- If you'd rather retain a "real" lz4 level in the matrix, `[None, 2]` gives the identical speedup (the win is dropping one of the three values, not which one is kept).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
           - python-version: "3.14t"
             python-conda-pkg: "python-freethreading=3.14"
 
-    name: Test (${{ matrix.os }}) - py ${{ matrix.python-version }}
+    name: Test w/o dask client (${{ matrix.os }}) - py ${{ matrix.python-version }}
 
     steps:
     - uses: actions/checkout@v6
@@ -94,10 +94,10 @@ jobs:
       run: |
         docker run -d --rm -p 8000:8000 -p 8001:8001 -p 8002:8002 -v ${{ github.workspace }}/tests/samples/triton_models_test:/models nvcr.io/nvidia/tritonserver:25.11-pyt-python-py3 tritonserver --model-repository=/models
 
-    - name: Test with pytest (without dask Client - run in parallel)
+    - name: Test with pytest (without dask Client)
       run: |
         python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "not dask_client" -n 4
-    - name: Test with pytest (with dask Client - run in parallel)
+    - name: Test with pytest (with dask Client)
       run: |
         python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "dask_client" -n 4
     - name: Upload codecov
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -97,9 +97,9 @@ jobs:
     - name: Test with pytest (without dask Client - run in parallel)
       run: |
         python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "not dask_client" -n 4
-    - name: Test with pytest (with dask Client)
+    - name: Test with pytest (with dask Client - run in parallel)
       run: |
-        python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "dask_client"
+        python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "dask_client" -n 4
```
ikrommyd, line 172: if we can use pytest-xdist on the dask client tests now, can't we just merge them into one task instead of having non dask_client and dask_client testing?

tests/test_local_executors.py
```diff
@@ -15,7 +15,10 @@
 @pytest.mark.parametrize("filetype", ["ttree", "rntuple", "parquet"])
 @pytest.mark.parametrize("skipbadfiles", [False, True, _exceptions])
 @pytest.mark.parametrize("maxchunks", [None, 1000])
-@pytest.mark.parametrize("compression", [None, 0, 2])
+# compression None vs not-None are the two distinct coffea code paths (executor.py
+# branches on `compression is None`); the integer level is only forwarded to lz4 and
+# does not change which coffea branch runs, so a single non-None level suffices.
```
ikrommyd: not a big fan of these large comments LLMs leave around often

tests/conftest.py
```diff
@@ -6,3 +6,29 @@
...
+    ``dask.compute`` dispatches through it; they do not depend on cluster
+    isolation. Spinning up a fresh ``LocalCluster`` per test (the previous
+    pattern) cost several seconds each and dominated the serial ``dask_client``
+    leg. Reusing one bounded cluster exercises the identical distributed
+    code path (cross-process task serialization + scheduler execution) while
+    paying the startup cost once per worker.
+
+    Bounded to a small, fixed footprint so the leg can run under pytest-xdist
+    without oversubscribing CI runners.
+    """
```
ikrommyd: I feel like this is LLM oververbosity

.github/workflows/ci.yml
```diff
@@ -97,9 +97,9 @@ jobs:
     - name: Test with pytest (without dask Client - run in parallel)
       run: |
         python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "not dask_client" -n 4
-    - name: Test with pytest (with dask Client)
+    - name: Test with pytest (with dask Client - run in parallel)
```
ikrommyd: Not really need this name change if we keep the two tasks. If we merge the tasks just "Test with pytest" should be good.

lgray: these should stay separate so we can run them in parallel, even with the time savings they're pretty chunky.

ikrommyd: yeah but we run the non dask client tests in parallel too so why shouldn't those be 1 task now?

lgray: The two subsets of tests were originally split because of how long they took, just making two shards of tests so the workflow itself runs in parallel on more CI nodes.

lgray: And given that things still take a while, probably best to leave them split.

.github/workflows/ci.yml
lgray: This was and still is a test-length issue. Honestly we should probably shard a bit further so the turn around is < 10 minutes, but then we also have  concurrency issues. :-/

lgray:
Not merging the two parallel testing runs since it still saves us time (especially in py3.10 on windows, what's even going on there?). Otherwise addressed.

.github/workflows/ci.yml
lgray: Ah - lol reading things wrong. Let me stick the dask tests into their own CI job then. We shouldn't be making people wait an hour for results.

ikrommyd:
```suggestion
    - name: Test with pytest (with dask Client)
```

ikrommyd (approved):
Just a minor suggestion cause I don't like it when LLMs change names of things to refelct a change in the code/config

.github/workflows/ci.yml
lgray: That's fine. You OK with me splitting out dask client as a separate CI run? it'll ~double the number of CI jobs but cut the time in half.

ikrommyd: Yeah sure I actually did not know which one is better so I wanted to try it out for a bit. I think more jobs that take less time is better. At some point I thought that less jobs that take longer would be better because we request fewer workers from github but I think that it's not true. You effectively want to revert https://github.com/scikit-hep/coffea/pull/1547 so go ahead. I just wanted to test it.

lgray: I'll set them both to just "Test with pytest" - the run in parallel doesn't really add any useful context any more.

lgray:
@ikrommyd Wild - so somehow it takes 46 minutes on windows in py3.10... That's pretty terrible.  Anyway - now ready for final review!

.github/workflows/ci.yml
```diff
@@ -94,12 +94,89 @@ jobs:
...
         docker run -d --rm -p 8000:8000 -p 8001:8001 -p 8002:8002 -v ${{ github.workspace }}/tests/samples/triton_models_test:/models nvcr.io/nvidia/tritonserver:25.11-pyt-python-py3 tritonserver --model-repository=/models
 
-    - name: Test with pytest (without dask Client - run in parallel)
+    - name: Test with pytest (without dask Client)
       run: |
         python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "not dask_client" -n 4
     - name: Test with pytest (with dask Client)
       run: |
-        python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "dask_client"
+        python -m pytest --cov-report=xml --cov=coffea --ignore=tests/test_taskvine_dask.py --ignore=tests/test_taskvine_virtual.py -m "dask_client" -n 4
```
ikrommyd: We don't need this part (The whole Test with pytest (with dask Client)) now at all right since there is a separate `test-dask-client` task.

.github/workflows/ci.yml
```diff
@@ -114,14 +191,6 @@ jobs:
       run: |
         cd docs && make html
         touch build/html/.nojekyll
```
ikrommyd: All this is now duplicated right? It should only be in the main `test` task and not in `test-dask-client` I think.

lgray: Since it builds in both cases we can lose one, yeah. Largely doesn't matter, it's fast.

.github/workflows/ci.yml
lgray: Ah, missed it.

ikrommyd: Probably the reason for the super slow CI times.

ikrommyd (approved):

ikrommyd:
From my side this is good to go so just hit the button whenever you're ready

ikrommyd:
@lgray would you mind telling your agent to send this patch over to https://github.com/scikit-hep/integration-tests too? We could speed up the coffea part of the CI using that which is actually the longest one. I'd rather keep the testing there with and without task though in the same CI job like it is now as there are many other packages we don't want to duplicate testing for.

Issue #1578, opened 2026-07-02
Claude Fable review of coffea


henryiii:
As requested on Slack. Ran on xhigh, otherwise followed https://iscinumpy.dev/post/claude-code-reviews/.

:robot: _AI text below_ :robot:

# Coffea project review

Seven parallel reviewers covered all of `src/coffea` plus packaging/CI (~29k lines). Many findings were **reproduced at runtime** against the repo's own test samples and the installed venv (marked ✓); I independently re-verified the highest-severity code-trace claims. Summary: **19 high-severity bugs** (several silently corrupt physics results), a large tail of medium/low bugs concentrated in copy-paste-duplicated code, one CI gate that can let red PRs merge, and a lint stack that could collapse into ruff.

> **Progress tracking:** each finding below is a checkbox; items whose fix has been **merged** are checked and annotated with the PR. Open fix PRs are annotated in parentheses so they can be tracked down. Merged so far: #1580, #1582, #1584.

## Critical bugs — silently wrong results

These produce wrong numbers without an exception, which for an analysis framework is the worst class:

- [x] **JER "down" variation corrupts `mass_orig`** — `jetmet_tools/CorrectedJetsFactory.py:385-392`: both `with_field` calls in the down block use `up` as the base record (the up block correctly uses its own fresh flatten). Final `JER.down.mass_orig` equals the **up-smeared** mass instead of the original. MET propagation is unaffected (reads only `pt_orig`), but anyone reading `mass_orig` on the down variation gets wrong values. _(fixed in #1580)_
- [ ] **`FsrPhoton.matched_muon` resolves against `Jet` in dask mode** — `nanoevents/methods/nanoaod.py:547`: the eager path uses `Muon._apply_global_index`, the dask path uses `Jet._apply_global_index` — silently wrong objects in one of three modes. _(#1581)_
- [x] **TreeMaker schema truncates field names to one character** ✓ — `schemas/treemaker.py:142`: `k[len(cname) + 1]` is missing the slice colon (`:`). On the shipped sample, `events.JetsJECup.fields == ['j', 'o']` instead of `['jerFactor', 'origIndex']`, and two variables sharing a first letter collide so one branch is **silently dropped**. _(fixed in #1582)_
[...]

PR #1580, opened 2026-07-06
fix(jetmet): JER down variation no longer corrupts mass_orig

NJManganelli:
**Part of #1578** — critical bug 1: *JER "down" variation corrupts `mass_orig`*.

In `CorrectedJetsFactory`, the JER "down" variation forwarded the original pre-smearing `pt_orig`/`mass_orig` by calling `awkward.with_field` with the already up-smeared `up` record as the base, rather than a fresh flatten of the original jets. Because the mass forward read its value from the previous (`up`-based) intermediate, `JER.down.mass_orig` silently held the **up-smeared mass** instead of the original mass; anyone reading `mass_orig` on the down variation got wrong values (MET propagation was unaffected). In practice `pt_orig` stayed correct only by ordering luck; both forwards are repaired to use the freshly flattened `down` record for robustness.

A regression test asserts that both `JER.up` and `JER.down` expose `pt_orig`/`mass_orig` equal to the untouched original arrays (and that the down originals differ from the up-smeared mass); it fails on `master` and passes with this fix. The full `tests/test_jetmet_tools.py` suite (17 tests, including the dask-awkward variants) passes, and `pre-commit run --all-files` is clean.

### DIFF
--- a/src/coffea/jetmet_tools/CorrectedJetsFactory.py
+++ b/src/coffea/jetmet_tools/CorrectedJetsFactory.py
@@ -383,10 +383,12 @@ class CorrectedJetsFactory:
             down = awkward.flatten(jets)
             # always forward the original (likely corrected) pt/mass
             down = awkward.with_field(
-                up, down[self.name_map["JetPt"]], where=self.name_map["JetPt"] + "_orig"
+                down,
+                down[self.name_map["JetPt"]],
+                where=self.name_map["JetPt"] + "_orig",
             )
             down = awkward.with_field(
-                up,
+                down,
                 down[self.name_map["JetMass"]],
                 where=self.name_map["JetMass"] + "_orig",
             )
--- a/tests/test_jetmet_tools.py
+++ b/tests/test_jetmet_tools.py
@@ -915,6 +915,20 @@ def test_corrected_jets_factory():
     assert name_map["JetPt"] + "_orig" in corrected_jets.JES_AbsoluteStat.up.fields
     assert name_map["JetPt"] + "_orig" in corrected_jets.JER.up.fields
 
+    # Regression test for scikit-hep/coffea#1578: the JER "down" variation must
+    # forward the original (pre-smearing) pt/mass in pt_orig/mass_orig, not the
+    # up-smeared values. Compare against the untouched original arrays.
+    orig_pt = ak.flatten(corrected_jets.pt_orig)
+    orig_mass = ak.flatten(corrected_jets.mass_orig)
+    for var in ("up", "down"):
+        assert ak.all(ak.flatten(corrected_jets.JER[var].pt_orig) == orig_pt)
+        assert ak.all(ak.flatten(corrected_jets.JER[var].mass_orig) == orig_mass)
+    # The down variation's forwarded originals must differ from the up-smeared values.
+    assert ak.any(
+        ak.flatten(corrected_jets.JER.down.mass_orig)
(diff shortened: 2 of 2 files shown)


lgray (commented):

**Summary:** This PR fixes a real physics-correctness bug and the fix is correct, minimal, and verified. On master, the JER down-variation block flattened the jets into `down` but then passed the already-built **`up`** record as the base of both `awkward.with_field` calls (old lines 385–391). The first call's result (a correct `pt_orig` written onto an `up` base) was immediately discarded because the second call based on `up` again; meanwhile the second call read `down[JetMass]` from that first result — i.e. from the up-variation record, whose `mass` had already been replaced with the up-smeared value at lines 369–370 — so `JER.down.mass_orig` silently held the **up-smeared mass**. `JER.down.pt_orig` stayed correct only because the final record inherited `up`'s own (correctly set) `pt_orig` — the "ordering luck" described in #1578 is accurate. The fix chains both `with_field` calls on `down` (`CorrectedJetsFactory.py:385-394`), making the down block symmetric with the up block (lines 349–356). Since the code path is shared, both eager and dask-awkward users were affected. **Approve** — I have no blocking findings.

I independently verified the blast-radius claims:
- **MET propagation unaffected — confirmed.** `corrected_polar_met` uses only jet `pt`/`phi`/`pt_orig` (`CorrectedMETFactory.py:8-12`), never `mass_orig`, and `pt_orig` was correct on master.
- **JES/junc variations unaffected — confirmed.** They are built from `template = awkward.zip(in_dict, ...)` (`CorrectedJetsFactory.py:468-473`), whose `_orig` fields come from the untouched fresh flatten (lines 240–243), not from the JER records.
- **No other instance of the pattern.** All remaining `with_field` chains in both factories are correctly based.

Runtime proof (uv venv, awkward 2.10.0, dask-awkward 2026.2.1): the new regression test passes at the PR head; with only `CorrectedJetsFactory.py` reverted to master it fails at `tests/test_jetmet_tools.py:925` on exactly the `mass_orig` assertion (`assert np.False_`) while the `pt_orig` assertion passes — so the test is non-vacuous and discriminating. Full `tests/test_jetmet_tools.py`: 17 passed, 0 failed at the PR head.

**Nits / Optional**
- `tests/test_jetmet_tools.py:918-928` — the regression asserts were added only to the eager `test_corrected_jets_factory`; the dask variant `test_corrected_jets_factory_dak` (line 975) has no `_orig` assertions at all. The fix lives in shared code so the eager test fully discriminates the fix, but mirroring the four asserts (plus `.compute()`) in the dak test would guard the dask path against future divergence. Fine to defer.
- `src/coffea/jetmet_tools/CorrectedJetsFactory.py:384` — optional: a one-line comment noting that the base must be the fresh flatten (not the `up` record) would inoculate against reintroducing the aliasing; the now-symmetric up/down blocks make this a nice-to-have rather than a need.

**Test coverage:** Good. The new test asserts both `pt_orig` and `mass_orig` on both `up` and `down`, compares against the untouched nominal `_orig` arrays, and includes a non-vacuity guard (`ak.any(down.mass_orig != up.mass)`) ensuring the smeared values genuinely differ from the originals in the sample. Discriminating property proven by the master-revert failure above. Only gap is the dask-path mirror noted in the nit.

**Pydantic:** N/A — the diff touches no pydantic code (grep: 0 matches in the diff and in `jetmet_tools/`).

lgray:
No further comments from me - a first review by eye looked fine.

PR #1582, opened 2026-07-06
fix(nanoevents): TreeMaker schema no longer truncates subbranch field names

NJManganelli:
**Part of #1578** — critical bug 3: *TreeMaker schema truncates field names (missing slice colon)*.

The TreeMaker schema's single-collection branch used `k[len(cname) + 1]` (a single-character index) rather than a slice, truncating each subbranch field name to one character. When two subbranches of a collection share a first letter — e.g. `jerFactor` and `origIndex` under `JetsJECup` — both collapse to the same one-character key and one is silently dropped, producing wrong-but-not-erroring data.

This changes the index to the slice `k[len(cname) + 1:]`, matching the already-correct nested-collection code path a few lines below. On the shipped `tests/samples/treemaker.root`, `events.JetsJECup.fields` goes from `['j', 'o']` to the correct `['jerFactor', 'origIndex']`. A parametrized regression test asserts the full field names are present, the truncated names are absent, and the values remain readable; it fails on the old code and passes with the fix (17/17 in `tests/test_nanoevents_treemaker.py`). Pre-commit clean.

### DIFF
--- a/src/coffea/nanoevents/schemas/treemaker.py
+++ b/src/coffea/nanoevents/schemas/treemaker.py
@@ -139,7 +139,7 @@ class TreeMakerSchema(BaseSchema):
 
             if cname not in branch_forms:
                 collection = zip_forms(
-                    {k[len(cname) + 1]: branch_forms.pop(k) for k in items}, cname
+                    {k[len(cname) + 1 :]: branch_forms.pop(k) for k in items}, cname
                 )
                 branch_forms[cname] = collection
             else:
--- a/tests/test_nanoevents_treemaker.py
+++ b/tests/test_nanoevents_treemaker.py
@@ -84,6 +84,26 @@ def test_nested_collection(collection, subcollection, arr_type, element, events)
         )
 
 
+@pytest.mark.parametrize(
+    "collection",
+    ["JetsJECup", "JetsJECdown"],
+)
+def test_subbranch_field_names_not_truncated(events, collection):
+    # Regression test for scikit-hep/coffea#1578: the schema previously used
+    # ``k[len(cname) + 1]`` (single index) instead of a slice, truncating
+    # subbranch field names to a single character. That silently collapsed
+    # distinct subbranches sharing a first letter (e.g. "jerFactor" and
+    # "origIndex" -> "j" and "o"), dropping one of them.
+    fields = events[collection].fields
+    assert "jerFactor" in fields
+    assert "origIndex" in fields
+    assert "j" not in fields
+    assert "o" not in fields
+    # Values must remain readable through the full field names.
+    assert ak.all(ak.num(events[collection].jerFactor.compute(), axis=1) >= 0)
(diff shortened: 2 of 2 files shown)


lgray (commented):

This PR is correct and I recommend merging. The one-character change at `src/coffea/nanoevents/schemas/treemaker.py:142` fixes a genuine truncation bug: in the single-collection zip path, `k[len(cname) + 1]` took one character instead of slicing off the `<collection>_` prefix. Since `items` only contains keys satisfying `k.startswith(cname + "_")` (treemaker.py:119), `len(cname) + 1` is exactly the right offset, and the fix makes this path consistent with the two sibling paths that already sliced correctly (treemaker.py:130 and :153). I verified at runtime that the new regression test passes at the PR head (17/17 for the whole file), and that reverting only the schema file to master makes both parametrizations fail at `assert 'jerFactor' in ['j', 'o']` — the test is non-vacuous and discriminating. A grep across `src/coffea/nanoevents/schemas/` found no other instance of this bug class (fcc.py and nanoaod.py already use slices), so the blast radius is TreeMakerSchema only.

**Nits/Optional**

- The comment in the new test (tests/test_nanoevents_treemaker.py:94–96) and the PR description present `jerFactor`/`origIndex` as the first-letter-collision example that "drops one of them" — but they truncate to *distinct* letters `j`/`o`, so both survive (renamed) and nothing is dropped. I swept the shipped sample: 8 collections were truncated under master (`Jets{,AK8}JEC{up,down}` → `['j','o']`, `Jets{,AK8}JER{up,down}` → `['o']`), and none of them actually loses a field, since no collection has two subbranches sharing a first letter. The silent-drop failure mode is real in general (dict comprehension last-write-wins, and `branch_forms.pop` consumes both source branches), it just isn't exhibited by this sample. Suggest rewording the comment so the example matches what actually happens.
- Optional: the sample has 8 affected collections but only `JetsJECup`/`JetsJECdown` are parametrized. Since a single code path fixes all of them this is adequate, but the `JetsAK8JEC*` variants could be added to the same parametrize for free; the `*JER*` variants carry only `origIndex`, so they'd need a slightly different assertion set.

**Test coverage**: good — asserts presence of full names, absence of truncated names, and readability, and the revert run proves it discriminates on exactly this bug. The test matches the file's existing conventions (module-scoped dask-mode `events` fixture, `.compute()` in assertions); the file does not parametrize over eager/virtual modes anywhere, and the schema's form-building is mode-independent, so that's fine. Minor note: the `ak.num(...) >= 0` assertions are tautologically true once `.compute()` succeeds — they prove the renamed fields are computable, which is their stated purpose, but the real discriminating power is in the field-name assertions.

Pydantic: not applicable — the diff touches no pydantic code (verified by grep).

lgray:
Scan by eye is fine - review bot confirms. Test prevents regression. Gonna merge.

PR #1584, opened 2026-07-06
fix(nanoevents): request !loadallowmissing branches so dask mode reads maybe-missing columns

NJManganelli:
**Part of #1578** — critical bug 5: *dask mode + saved union forms fabricates all-None columns*.

`keys_for_buffer_keys` translated buffer keys to branch names with an exact `== "!load"` comparison, but saved/union forms (from `preprocess(..., save_form=True)` across files where some lack a branch) tag maybe-missing branches with the `!loadallowmissing` token. Those branches never matched and were never requested from the file, so dask mode backfilled them as **all-None even for files that actually contain the branch** — silently wrong data.

The fix matches `startswith("!load")`, covering both tokens, consistent with the token dispatch in `nanoevents/mapping/base.py`. Adds a fast unit regression test on `keys_for_buffer_keys` and an end-to-end dask test that preprocesses a two-file union form and confirms the present branch returns its real values (both fail without the fix). Full `tests/test_nanoevents.py` passes; pre-commit clean.

Note: a *genuinely* missing branch in dask mode can still surface a `KeyInFileError` from uproot's direct `tree.arrays` read — that is addressed in a follow-up PR stacked on this one. The same `== "!load"` pattern in `trace._form_keys_to_columns` is fixed in a separate companion PR.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -53,7 +53,15 @@ class _map_schema_base:  # ImplementsFormMapping, ImplementsFormMappingInfo
                 [
                     name
                     for name, maybe_transform in zip(operands, it_operands)
-                    if maybe_transform == "!load"
+                    # Match both "!load" and "!loadallowmissing": a saved/union form
+                    # marks branches that may be missing in some files with the
+                    # "!loadallowmissing" token (see nanoevents.mapping.base, which
+                    # dispatches on node.startswith("!load")). Both tokens denote a
+                    # real column read, so both branches must be requested from the
+                    # file. Matching only "!load" silently drops the maybe-missing
+                    # branches, causing dask mode to fabricate them as all-None even
+                    # in files that actually contain them.
+                    if maybe_transform.startswith("!load")
                 ]
             )
         return base_columns
--- a/tests/test_nanoevents.py
+++ b/tests/test_nanoevents.py
@@ -2,6 +2,7 @@ import os
 from pathlib import Path
 
 import awkward as ak
+import numpy as np
 import pytest
 import uproot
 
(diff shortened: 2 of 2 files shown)


lgray (commented):

**Summary.** This PR fixes a real silent-wrong-data bug: `keys_for_buffer_keys` (src/coffea/nanoevents/factory.py:43) translated buffer keys to branch names with an exact `== "!load"` match, so branches tagged `!loadallowmissing` by a saved/union form (built by `_lazify_form`, src/coffea/nanoevents/mapping/uproot.py:64–68) were never requested by uproot's dask column projection (`uproot/_dask.py` calls `keys_for_buffer_keys` at its projection/report sites). `load_buffers` then read only the requested branches (factory.py:170), and `PreloadedSourceMapping.extract_column` fabricated an all-None `IndexedOptionArray` for the unread branch (src/coffea/nanoevents/mapping/preloaded.py:87–91) — even for files that contain it. The one-line fix (`startswith("!load")`, factory.py:64) matches exactly the right token set — the full `!*` token inventory contains no other `!load*` token and no transform named `load*` — and mirrors the authoritative dispatcher in mapping/base.py:122/128. I verified by measurement: at the PR head both new tests pass and the has-branch file returns real values in dask mode; with only factory.py reverted to the merge-base, both tests fail (`assert set() == {'flag'}`; `assert 'flag' in set()`) and a direct reproduction returns 20/20 None from a file that contains the branch. Eager/virtual modes are genuinely unaffected (they never call `keys_for_buffer_keys`; `from_root` forwards `known_base_form` only in the dask branch, factory.py:369). Full `tests/test_nanoevents.py` passes 21/21. This is correct, minimal, well-commented, and mergeable — the notes below are about sequencing and polish, not the fix itself.

**Should fix**
- **Merge sequencing with the stacked KeyInFileError PR.** Measured behavior change: before this PR, a dask read of the maybe-missing branch from a file *lacking* it returned the semantically correct all-None column; after this PR it raises `uproot.KeyInFileError: not found: 'flag'` at compute time (because the branch is now passed to `tree.arrays` in `load_buffers`, factory.py:170, and uproot raises before the `allow_missing` machinery in preloaded.py:87 is reached). Since heterogeneous filesets are exactly the use case for union forms, this PR alone trades "wrong data on present-branch files + working missing-branch files" for "correct present-branch files + hard error on missing-branch files". A loud error beats silent wrong data, so this is the right trade — but the stacked PR restoring the missing-branch path should land in the same release, and it's worth stating this behavior change explicitly in the PR description/changelog.

**Nits / Optional**
- The deferred missing-branch behavior is invisible in the test file — the e2e test only loads the file that has the branch. A short comment (or an `xfail`/`raises` assertion pinning the current `KeyInFileError`, to be flipped by the stacked PR) would document the known gap where a future reader will look for it (tests/test_nanoevents.py:376ff).
- tests/test_nanoevents.py:404–410: the dict-vs-`FilesetSpec`/`DatasetSpec` dance can collapse to `ds = DatasetSpec.model_validate(available["ds"])` — the `FilesetSpec` RootModel defines `__getitem__` (src/coffea/dataset_tools/filespec.py:401) and pydantic v2 `model_validate` accepts model instances, so both branches are redundant.

**Test coverage.** Both tests are non-vacuous and discriminating, proven by the revert run (unit fails on `set() == {'flag'}`, e2e fails on the necessary-columns assertion; the underlying symptom — 20/20 fabricated Nones from a file containing the branch — was reproduced separately). The unit test exercises `!loadallowmissing` via both `!index` and `!content` buffer variants plus a plain `!load` key and a mixed set, and the synthetic keys match the exact production format from `_lazify_form`. The e2e test builds the union form the way users hit it (`preprocess(..., save_form=True)` across a file lacking the branch), asserts the form is actually an `IndexedOptionArray`, and checks real values (`ak.all(computed)`), not just non-None-ness. It doesn't pin eager/virtual for the same fileset, but there is no direct analogue to pin: `from_root` only consumes `known_base_form` in dask mode (factory.py:369), and I verified both modes return correct values from the has-branch file regardless.

**Pydantic.** No pydantic code is touched by the diff; the e2e test consumes preprocess output idiomatically (`DatasetSpec.model_validate` + the `form` computed field, filespec.py:596).

lgray:

The earlier "should fix" on this PR — the merge-sequencing concern that a dask read of a maybe-missing branch from a file lacking it now raises `uproot.KeyInFileError` at compute time — is **resolved by the stacked #1592** (verdict: **YES**). Measured in an isolated env at #1592's head: reverting only `src/coffea/nanoevents/factory.py` back to this PR's head reproduces the predicted failure — `test_union_form_genuinely_missing_branch_dask` fails with `uproot.KeyInFileError('flag')` at `factory.py:170` (`tree.arrays(...)`); on #1592's head that same read returns an all-None `option[bool]` column (no exception), matching the union form / eager-virtual mapping semantics, while a genuinely-required absent branch still fails loudly with `KeyError`. #1592 also adds the missing-branch test coverage that was absent here.

Recommend merging in order — **#1584 first, then #1592, in the same release** — so master never carries the KeyInFileError-on-missing-branch window.

PR #1585, opened 2026-07-06
fix(nanoevents): Candidate + Candidate no longer silently drops charge

NJManganelli:
**Part of #1578** — critical bug 6: *Candidate + Candidate silently drops charge*.

Candidate-derived four-vector addition was silently dropping the `charge` field for same-class sums. The cause was ordering: a module-level `awkward._util.copy_behaviors` call ran **before** the `@awkward.mixin_class` decorator, and because `mixin_class` registers ufunc behaviors with `setdefault`, the copied charge-less `LorentzVector.add` pre-empted `Candidate`'s own charge-propagating `add` for the `(add, Candidate, Candidate)` key. The same anti-pattern in `nanoaod.py` made `GenVisTau + GenVisTau` drop charge even though `GenVisTau + Muon` kept it. (Classes copying from `PtEtaPhiMCandidate` — Muon, Tau, etc. — were unaffected because that source `add` already propagates charge.)

This PR moves the `copy_behaviors` merges to run after each class is defined and merges them with `setdefault` semantics, so cross-class behaviors are still copied while the charge-aware `add` wins for same-class sums. Regression tests cover `Candidate`, `PtEtaPhiMCandidate`, `PtEtaPhiECandidate`, and `GenVisTau` (from a NanoAOD sample), verified to fail before the fix and pass after (35/35 in `tests/test_nanoevents_vector.py`, 19/19 in `tests/test_nanoevents.py`). Pre-commit clean.

### DIFF
--- a/src/coffea/nanoevents/methods/candidate.py
+++ b/src/coffea/nanoevents/methods/candidate.py
@@ -12,8 +12,6 @@ from coffea.nanoevents.methods import vector
 
 behavior = dict(vector.behavior)
 
-behavior.update(awkward._util.copy_behaviors("LorentzVector", "Candidate", behavior))
-
 
 @awkward.mixin_class(behavior)
 class Candidate(vector.LorentzVector):
@@ -57,6 +55,19 @@ class Candidate(vector.LorentzVector):
         parent = super()
         if hasattr(parent, "__awkward_validation__"):
             parent.__awkward_validation__()
+
+
+# Copy the cross-class LorentzVector behaviors (e.g. Candidate + TwoVector) onto
+# Candidate, but only for keys the ``@mixin_class`` decorator has not already
+# registered. This MUST run after the decorator: ``copy_behaviors`` would
+# otherwise pre-register ``(add, Candidate, Candidate)`` -> LorentzVector.add via
+# the ``setdefault`` used by ``mixin_class``, shadowing Candidate's own
+# charge-propagating ``add`` (see scikit-hep/coffea#1578).
+for _key, _value in awkward._util.copy_behaviors(
+    "LorentzVector", "Candidate", behavior
+).items():
+    behavior.setdefault(_key, _value)
+del _key, _value
 
 
 @awkward.mixin_class(behavior)
--- a/src/coffea/nanoevents/methods/nanoaod.py
+++ b/src/coffea/nanoevents/methods/nanoaod.py
@@ -199,10 +199,6 @@ GenParticleRecord.ProjectionClass3D = vector.ThreeVectorRecord  # noqa: F821
 GenParticleRecord.ProjectionClass4D = GenParticleRecord  # noqa: F821
 GenParticleRecord.MomentumClass = vector.LorentzVectorRecord  # noqa: F821
 
-behavior.update(
-    awkward._util.copy_behaviors("PtEtaPhiMLorentzVector", "GenVisTau", behavior)
-)
-
 
 @awkward.mixin_class(behavior)
 class GenVisTau(candidate.PtEtaPhiMCandidate, base.NanoCollection):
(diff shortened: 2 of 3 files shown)


lgray (commented):

## Summary

This is a correct, well-targeted fix. The diagnosis holds up under verification: `awkward.mixin_class` registers ufunc behaviors with `dict.setdefault` (`awkward/behaviors/mixins.py:55,58,60`), so on master the module-level `copy_behaviors(...)`/`behavior.update(...)` that ran *before* the `@mixin_class` decorator pre-seeded `(numpy.add, "Candidate", "Candidate") -> LorentzVector.add`, which then blocked `Candidate`'s own charge-propagating `add`. I reproduced the bug on master (`Candidate + Candidate` → fields `['x','y','z','t']`, charge silently dropped) and confirmed the fix restores it (`['x','y','z','t','charge']`, `charge = [0]`). Same story for `GenVisTau + GenVisTau`. The reorder-then-`setdefault` approach is exactly right, and the accompanying comments explain the non-obvious ordering constraint clearly. Recommend merge.

Key points I verified:
- **No cross-class behavior lost.** Comparing the full `(ufunc, *names)` key sets of `candidate.behavior` and `nanoaod.behavior` between master and PR head, the sets are **identical** — the fix changes only the *value* bound to the same-class `add` keys, so `Candidate + TwoVector`, `GenVisTau + Muon`, etc. still resolve. The new `setdefault` loop correctly fills only missing cross-class keys and does not clobber the decorator's own `add` (it's `setdefault`, not `update`).
- **Blast radius is fully covered for this bug class.** The only classes with a custom charge-carrying `add` are `Candidate` and `MomentumCandidate` (edm4hep/fcc). All nanoaod leptons/jets copy from `PtEtaPhiMCandidate`, whose `add` already carries charge, so they were never affected. `GenParticle` (copies from the charge-less `PtEtaPhiMLorentzVector`) is not charge-bearing, so charge-less `add` is correct there.
- **Full suites pass** (`test_nanoevents_vector.py` 31 passed; `test_nanoevents.py` 18 passed), no regressions.

## Nits / Optional

- **edm4hep / fcc `MomentumCandidate` (out of scope, worth a follow-up).** `edm4hep.py:259` and `fcc.py:77` call `copy_behaviors(vector.LorentzVector, MomentumCandidate, behavior)` with **class objects**, but `copy_behaviors` matches names by string — so those calls return **0 entries and are silent no-ops**. `MomentumCandidate + MomentumCandidate` still keeps charge (the decorator binds it and nothing overwrites), so there's no charge bug here — but the *intended* cross-class behavior copies never actually happen. Not caused by this PR; consider a separate issue to pass `"LorentzVector"`/`"MomentumCandidate"` strings there.
- **Minor duplication.** The `for _key, _value in copy_behaviors(...).items(): behavior.setdefault(...)` / `del _key, _value` block plus its explanatory comment is duplicated verbatim in `candidate.py:66` and `nanoaod.py:225`. Four lines twice is fine as-is; only worth a tiny shared helper if a third copy site appears (they could drift otherwise).

## Test coverage

The new tests are non-vacuous and discriminating: reverting both source files to master makes both tests **fail** with `charge` absent from `.fields` (not present-but-wrong), and they pass on PR head. They assert `charge` is both present **and** numerically equal to the summed input charges, cover all four named classes (`Candidate`, `PtEtaPhiMCandidate`, `PtEtaPhiECandidate`, `GenVisTau`), include a same-sign sum (`c1 + c1` → `2`), and guard the never-regressed cross-class path (`GenVisTau + Muon`). Coverage is eager-mode only, which is sufficient here: behavior registration is mode-independent — the same `behavior` dict drives ufunc dispatch, and dask just wraps the identical behavior — so there's no distinct dask code path to exercise for this fix.

## Pydantic

Not applicable — the diff contains no pydantic-related changes (`grep -i pydantic` over the diff is empty).

lgray:
Tests cover addition appropriately and in a way that shouldn't miss any further drift later. Fix is concise. I'm adding another issue to pick up the edm4hep / fcc comment, since I'd like to avoid silent no-ops in the code base.

PR #1586, opened 2026-07-06
fix(nanoevents): ThreeVector.unit returns a genuine unit vector (normalize by 3D magnitude)

NJManganelli:
**Part of #1578** — critical bug 7: *ThreeVector.unit returns non-unit vector*.

`ThreeVector.unit` divided by `self.rho`, but for a 3D vector `rho` is only the transverse magnitude sqrt(x²+y²), so the returned vector was not a unit vector — e.g. `(3,4,12).unit` had 3D magnitude 13/5 = 2.6 instead of 1. This changes the normalization to `self.p` (the full 3D magnitude), so `unit` now returns a vector of length 1 pointing in the same direction. `SphericalThreeVector`, which inherits this property, is fixed by the same change.

`TwoVector.unit` is intentionally left alone (there `rho` *is* the 2D norm), and `LorentzVector.unit` is unaffected because it resolves to the scikit-hep `vector` library's own unit-mass method rather than this property. The existing test asserted `unit.rho == 1`, which codified the bug; it is updated to assert `unit.p == 1` with direction (theta, phi) preserved, and an explicit `(3,4,12)` regression case is added (33/33 in `tests/test_nanoevents_vector.py`). Pre-commit clean.


lgray (commented):

**Summary.** This is a correct, minimal fix for a real bug. `ThreeVector.unit` divided by `self.rho` — the *transverse* magnitude √(x²+y²) — so the returned vector was not unit length in 3D; e.g. `(3,4,12).unit` had 3D magnitude 13/5 = 2.6. The one-line change to divide by `self.p` (full 3D magnitude √(x²+y²+z²)) is the right normalization, preserves direction (positive scalar ⇒ θ, φ unchanged), and returns the same 3D-vector type. **Recommend merge.**

I verified the definitions against the actual code and at runtime (scikit-hep `vector` 1.8.1): `rho`=5.0 vs `p`=13.0 for (3,4,12), so `/self.p` is correct. The fix is confirmed non-vacuous by reverting only `vector.py` to master — the updated tests then fail with `unit.p` ≈ 2.03/1.87 instead of 1.

**Blast radius — all checked, all good:**
- `SphericalThreeVector` inherits `ThreeVector.unit` (no override) and is fixed by the same change — probe gives `unit.p`=1.0 with θ,φ preserved.
- `TwoVector.unit` is correctly left alone: it divides by `self.r`=`self.rho`, which for a 2D vector *is* the full norm √(x²+y²).
- `LorentzVector.unit` is unaffected — it resolves to `vector._methods.Lorentz` (a callable method, `v.unit()`), not this module's property; `'unit'` is not in `LorentzVector`'s class dict.
- No consumer of `ThreeVector.unit` exists elsewhere in `src/coffea/`, so no downstream double-correction.

**Nits/Optional**
- Nice bonus (worth mentioning in the PR description): the fix also removes a div-by-zero on purely longitudinal vectors. `(0,0,5).unit` previously gave `z=inf` (rho=0); it now correctly gives `z=1.0`. This is the most extreme manifestation of the bug and would be a strong regression case to add explicitly.
- The zero vector `(0,0,0).unit` still yields `nan` under both old and new (0/0) — unchanged, not introduced here, and arguably out of scope; just noting it isn't guarded.
- The added docstring already inoculates against the exact rho-vs-p confusion that caused the bug — good call.

**Test coverage.** Strong. The old assertion `unit.rho == 1` (which codified the bug) is correctly *replaced*, not duplicated. The new tests assert `unit.p == 1`, preserve both `theta` and `phi`, add an explicit `(3,4,12)` regression checking each component (3/13, 4/13, 12/13), and cover the inheriting `SphericalThreeVector`. Proven discriminating: reverting the source flips both tests to failing. One optional gap: no test for the longitudinal `(0,0,z)` case where the old code produced `inf` — the sharpest discriminator, verified by my probe but not codified.

**Pydantic.** Not applicable — the diff touches no pydantic code (`grep -ci pydantic` on the diff = 0).

src/coffea/nanoevents/methods/vector.py
```diff
@@ -315,8 +315,12 @@ def delta_phi(self, other):
 
     @property
     def unit(self):
-        """Unit vector, a vector of length 1 pointing in the same direction"""
-        return self / self.rho
+        """Unit vector, a vector of length 1 pointing in the same direction
+
+        For a 3D vector the length is the full 3D magnitude ``p``
+        (:math:`\\sqrt{x^2+y^2+z^2}`), not the transverse magnitude ``rho``.
```
ikrommyd: we definitely don't want the previous bug in the docstring "not the transverse magnitude ``rho``" is bad

ikrommyd (changes requested):

ikrommyd:
I suspect this is a relic of migration to scikit-hep/vector

lgray:
Hmm this may be for a part of vector that wasn't there when we did the migration (though it's hard to think of a vector package not having unit())... It appears to be there now... I'd be fine with just removing unit() (though iirc it would demote the class to the vector class)

nsmith-:
This brings to mind that we should collect any user-behavior changes for the release notes. People may have started to rely on the bugged behavior.

NJManganelli:
Addressed @ikrommyd's note: the docstring no longer carries the old "not the transverse magnitude `rho`" wording (that line described the bug it was fixing — removed in `aa4fcfaf`), and master is merged in.

On keeping vs. removing `unit()`: I'd lean **keep**. It's now correct (normalizes by the 3D magnitude), and since scikit-hep/vector does provide `unit()` on its 3D types, removing it here would be a silent API regression for anyone already calling it — including those who may have worked around the bug. Happy to remove instead if you'd rather, but keeping-and-fixing seems lower-risk.

Per @nsmith-'s point, this is a user-visible behavior change, so it wants a release-note line: *"`ThreeVector.unit` now returns a true unit vector (normalized by the 3D magnitude `p`); previously it divided by the transverse magnitude `rho`, so results were wrong for vectors with a z-component."* I'll leave that for a batched release-notes pass rather than adding it here.

ikrommyd:
I just saw in the vector docs that scikit-hep vector defines a unit behavior too. Do we overwrite it here? In that case, do we even need it? 😄 can’t we just delete it in that case and inherit it from vector?

NJManganelli:
If we are fine with demotion to vector as @lgray noted

ikrommyd:
if vector does the same and correct calculation, I suggest we do that. I think we shouldn't add more than we need on top of vector anywhere.

lgray:
I think then the question really becomes "do we remove the vector classes entirely and have candidates inherit directly from vector" -> some names would change but I don't think we lose any functionality at this point?

NJManganelli:
Hmmm, if this is going to get upscoped (and not a criticism, sounds like a good thing to improve the code now that things have evolved) to that kind of refactor, someone else must take over (if needed, copy this branch and close this PR)

ikrommyd:
We lose things like `metric_table` no? And particle specific things like `matched_jet`. We can inherit from vector I think but we need the coffea specific stuff unless they are all upstreamed to vector. But `unit` does not fall into that scope and if it's something that vector has for us, then I thought we may just want to delete it insead of fixing it.

lgray:
Those things are specific to candidates - not vectors. So, technically, no we do not lose them.

lgray:
Ah, no metric_table is on the vector classes, ah well.

nsmith-:
IIRC the coffea-vector lives on top of vector to add new functions like metric_table but also a bunch of aliases for backwards compatibility. We could do a deprecation round for all the aliases, but that still leaves metric_table.

lgray:
Yeah I'm not sure what the typical usage patterns for `metric_table` are: off of candidates usually or off of actual vectors?

Another possibility is we turn it into a free function.

lgray:
Holding final review of this PR since we want to coordinate with vector about it.

ikrommyd:
scikit-hep/vectors `unit()` method is correct and our vectors were already inheriting that for lorentz vectors. I don't think it's nice to have a property for 2d/3d vectors and a method for 4d vectors. Let's just go with the method inherited from vector for all of them.

ikrommyd (approved):
I think the right way to do this is at is it right now and remove the `unit` property and just have the `unit()` method our vectors already inherit from scikit-hep/vector. Our 4D vectors already have this and it is correct. It's not nice to have a property for 2D/3D and a method for 4D. It's a tiny breaking change since users might need to add `()` if they used the unit property (which was wrong anyways) but that makes it consistent with lorentz vectors so I say that's fine.

lgray:
Yeah I think this is a reasonable way to go. We should announce the breaking change in the release notes.

PR #1589, opened 2026-07-06
fix(nanoevents): make NanoAODSchema work with from_preloaded (guard __doc__ access)

NJManganelli:
**Part of #1578** — bug 15: *NanoAODSchema breaks from_preloaded*.

Building NanoAOD events from a preloaded column source (`NanoEventsFactory.from_preloaded`, e.g. via `SimplePreloadedColumnSource`) raised `KeyError: '__doc__'` in `NanoAODSchema._build_collections`: preloaded mappings don't inject a `"__doc__"` entry into the offsets branch parameters that uproot-backed forms carry, and the access was unguarded. The breakage was hidden because the relevant test was marked `xfail`.

This PR guards the lookup with `.get("__doc__", "no docstring available")`, mirroring the pattern already used in the Delphes schema. It also un-xfails `test_preloaded_nanoevents` (whose "no dask yet" skip reason was stale for the eager path it exercises) and adds the `run`/`luminosityBlock`/`event` columns the test's column source needs to satisfy the newer `error_missing_event_ids` validation — a test-fixture gap, not a source bug. Verified the test fails with `KeyError: '__doc__'` before the change and passes after. Pre-commit clean.

### DIFF
--- a/src/coffea/nanoevents/schemas/nanoaod.py
+++ b/src/coffea/nanoevents/schemas/nanoaod.py
@@ -375,7 +375,9 @@ class NanoAODSchema(BaseSchema):
                 )
                 output[name]["content"]["parameters"].update(
                     {
-                        "__doc__": offsets["parameters"]["__doc__"],
+                        "__doc__": offsets["parameters"].get(
+                            "__doc__", "no docstring available"
+                        ),
                         "collection_name": name,
                     }
                 )
--- a/tests/test_preloaded.py
+++ b/tests/test_preloaded.py
@@ -9,8 +9,6 @@ from coffea.processor.test_items import NanoEventsProcessor
 
 
 def test_preloaded_nanoevents():
-    pytest.xfail("preloaded nanoevents doesn't support dask yet")
-
     columns = [
         "nMuon",
         "Muon_pt",
@@ -20,6 +18,10 @@ def test_preloaded_nanoevents():
         "Muon_charge",
         "nJet",
         "Jet_eta",
+        # event ID fields are required by NanoAODSchema.error_missing_event_ids
+        "run",
+        "luminosityBlock",
+        "event",
     ]
     p = NanoEventsProcessor(columns=columns)
 
--- a/tests/test_preloaded.py
+++ b/tests/test_preloaded.py
@@ -23,7 +23,7 @@ def test_preloaded_nanoevents():
         "luminosityBlock",
         "event",
     ]
-    p = NanoEventsProcessor(columns=columns)
+    p = NanoEventsProcessor(columns=columns, mode="eager")
 
     rootdir = uproot.open(os.path.abspath("tests/samples/nano_dy.root"))
     tree = rootdir["Events"]


lgray (commented):

Thanks for chasing down bug 15. The core fix is correct and I verified it end-to-end: reverting only `nanoaod.py` to master reproduces `KeyError: '__doc__'` at `nanoaod.py:378` during `.events()` schema-build (no dask involved), and the `.get("__doc__", "no docstring available")` guard resolves it — matching the Delphes-schema pattern at `delphes.py:330-335`. The added `run`/`luminosityBlock`/`event` columns are genuinely required (the default `error_missing_event_ids=True` check at `nanoaod.py:271-279` would otherwise abort before the `__doc__` line). Blast radius is clean: line 378 is the *only* `__doc__` access in `nanoaod.py`, so this is a complete fix for the preloaded NanoAOD path, not a partial one.

One blocker: un-xfailing the test made it red in the "Test without dask" CI cell.

**Should fix**
- **CI failure in the no-dask matrix cell.** `NanoEventsProcessor(columns=columns)` defaults to `mode="dask"` (`NanoEventsProcessor.py:8`), so `p.process(events)` hits `import hist.dask` (`NanoEventsProcessor.py:34`), which raises `ModuleNotFoundError` when `dask_histogram` isn't installed. Reproduced exactly:
  ```
  NanoEventsProcessor.py:34: in accumulator
      import hist.dask as dah
  ModuleNotFoundError: for hist.dask, install the 'dask_histogram' package ...
  ```
  Note the `__doc__` bug you're fixing is exercised earlier at `from_preloaded(src).events()` and needs *no* dask — only the histogram fill does. So the cleanest fix keeps the whole test running everywhere by using the processor's eager mode (see Proposed fix). This also happens to be the more honest test: `from_preloaded(...).events()` produces eager arrays, and filling a `dask.Hist` with eager arrays (what the current `mode="dask"` default does) is a slightly odd mix.

**Nits / Optional**
- `delphes.py:346` still has an unguarded `offsets["parameters"]["__doc__"]` (the sibling of the guarded 330-335 block). Out of scope for this PR and not on the NanoAOD path, but if Delphes ever grows a `from_preloaded` test it'll hit the same KeyError — worth a follow-up.

**Test coverage**
The test is genuinely discriminating (proven: master `nanoaod.py` → KeyError, PR → pass) and non-vacuous. With the recommended `mode="eager"` fix it exercises the full path — schema build (`__doc__` regression), `p.process`, the cutflow assertions (`pt==18`, `mass==6`), and the `events.Muon.matched_jet` AttributeError — in *every* CI cell including no-dask. If you instead take the `importorskip` route, be aware the `__doc__` regression stops being covered in the no-dask cell (still covered in the ~24 with-dask cells), which is a small but real coverage gap given the bug itself doesn't need dask.

**pydantic**: N/A — no pydantic in this diff or in the nanoevents schema code path.

### Proposed fix

Recommended (minimal, zero coverage loss — verified: passes both with and without `dask_histogram`):
```diff
--- a/tests/test_preloaded.py
+++ b/tests/test_preloaded.py
@@ -23,7 +23,7 @@ def test_preloaded_nanoevents():
         "luminosityBlock",
         "event",
     ]
-    p = NanoEventsProcessor(columns=columns)
+    p = NanoEventsProcessor(columns=columns, mode="eager")
 
     rootdir = uproot.open(os.path.abspath("tests/samples/nano_dy.root"))
     tree = rootdir["Events"]
```
Rationale: preloaded events are eager, so the eager processor path avoids `import hist.dask` entirely and keeps the whole test (incl. the `__doc__` regression) running in the no-dask cell.

Alternative (matches the `dask_awkward` importorskip idiom used elsewhere in `tests/`, but skips the entire test — and thus the `__doc__` coverage — in the no-dask cell):
```diff
--- a/tests/test_preloaded.py
+++ b/tests/test_preloaded.py
@@ -9,6 +9,8 @@ from coffea.processor.test_items import NanoEventsProcessor
 
 
 def test_preloaded_nanoevents():
+    pytest.importorskip("dask_histogram")
+
     columns = [
```
I'd take the `mode="eager"` one. (A stronger importorskip variant — split the test so `.events()` + the AttributeError assertion run unconditionally and only `p.process` is guarded — also works but is a larger change and still filling dask hists with eager arrays; `mode="eager"` gets the same coverage more simply.)

lgray:

Pushed a one-line follow-up (`177d0832`) to fix the failing **Test without dask** cell.

`test_preloaded_nanoevents` was un-xfailed but calls `p.process(events)`, whose `accumulator` does `import hist.dask` under the default `mode="dask"`. That import requires `dask_histogram`, which isn't installed in the no-dask CI env, so the test failed there with `ModuleNotFoundError`. The `__doc__` regression this PR actually guards is exercised earlier at `from_preloaded(...).events()` and needs no dask.

Fix: construct the processor with `mode="eager"` (preloaded events are eager anyway). Verified in an isolated env:

- **no `dask_histogram` + fix** → passes (was the failing cell)
- **no `dask_histogram`, fix reverted** → fails with the exact `ModuleNotFoundError: hist.dask` CI error
- **with `dask_histogram` + fix** → still passes

This keeps the whole test — including the `__doc__` path — green in every cell, with no coverage loss. pre-commit clean.

PR #1591, opened 2026-07-06
fix(nanoevents): include !loadallowmissing branches in typetracer column reporting

NJManganelli:
**Part of #1578** — reporting-side companion to the `keys_for_buffer_keys` fix for *dask mode + saved union forms fabricates all-None columns*.

The typetracer-based necessary-columns reporting in `nanoevents/trace.py` (`_form_keys_to_columns`, used by all `trace_*` entry points) matched the coffea-DSL load token with an exact `== "!load"` comparison. Branches that may be absent in some files are lazified with the `!loadallowmissing` token (from `nanoevents/mapping/uproot.py`), so the exact match silently dropped them from the reported necessary columns.

This changes the comparison to `startswith("!load")`, consistent with how the token is consumed in `nanoevents/mapping/base.py` and with the `keys_for_buffer_keys` fix. A unit-level regression test constructing both a normal `!load` node and a `!loadallowmissing` node (using the real quoted-node buffer-key format, verified against a virtual-mode `data_touched` report) fails on the old code and passes on the new. Pre-commit clean.

### DIFF
--- a/src/coffea/nanoevents/trace.py
+++ b/src/coffea/nanoevents/trace.py
@@ -97,11 +97,16 @@ def _make_length_zero_one_tracer(
 def _form_keys_to_columns(touched: list) -> frozenset[str]:
     # translate the touched buffer keys to branch names
     keys = set()
-    # each buffer key encodes the necessary branches through a "!load" instruction in the coffea DSL
+    # each buffer key encodes the necessary branches through a "!load" instruction in the coffea DSL.
+    # Branches that may be missing in some files carry a "!loadallowmissing" token instead of a
+    # plain "!load" (generated in nanoevents/mapping/uproot.py), so match with startswith("!load")
+    # to stay consistent with nanoevents/mapping/base.py.
     for _buffer_key in touched:
         elements = unquote(_buffer_key.split("/")[-1]).split(",")
         keys |= {
-            elements[idx - 1] for idx, instr in enumerate(elements) if instr == "!load"
+            elements[idx - 1]
+            for idx, instr in enumerate(elements)
+            if instr.startswith("!load")
         }
     return frozenset(keys)
 
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -193,3 +193,28 @@ def test_tracing_nanoevents():
     _untypetracable_analysis(events)
     access_log = [x.branch for x in access_log]
     assert sorted(set(access_log)) == ["MET_sumEt"]
+
+
+def test_form_keys_to_columns_loadallowmissing():
+    # Regression test for scikit-hep/coffea#1578: branches that may be missing in
+    # some files carry the "!loadallowmissing" token (generated in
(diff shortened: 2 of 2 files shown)


lgray (commented):

**Summary.** This is the reporting-side companion to #1578: `_form_keys_to_columns` in `nanoevents/trace.py` extracted necessary columns by matching the load token with an exact `instr == "!load"`, which silently dropped any branch lazified with the `!loadallowmissing` token (produced in `mapping/uproot.py:65-70` for possibly-missing branches). The one-line fix to `instr.startswith("!load")` makes the typetracer necessary-columns report include those branches, matching the consumer's own dispatch in `mapping/base.py:122`. **The change is correct, minimal, and well-covered. Recommend merge.**

I verified the fix and the test end-to-end (evidence in the appendix): the token inventory across `src/coffea/nanoevents` is exactly `{!load, !loadallowmissing}` as the only `!load*` tokens, so `startswith("!load")` cannot over-match; the test format matches real buffer keys captured from a virtual-mode trace; and reverting only `trace.py` to master makes the new test fail with `Muon_missing` dropped from the reported set (`frozenset({'Muon_pt'})`), confirming it is discriminating.

**Bugs/Blockers**
- None.

**Should fix**
- None blocking. One thing worth stating explicitly in the PR/commit for reviewers tracking #1578: the sibling exact-match `keys_for_buffer_keys` at `src/coffea/nanoevents/factory.py:56` still uses `== "!load"` on current master — that's the **request side** handled by #1584, which is not yet merged. So this PR alone fixes the *report*, but possibly-missing/union branches only flow end-to-end once #1584 also lands. Not a defect in this PR (the split is intentional), just a merge-ordering note so the report side isn't assumed to be sufficient on its own.

**Nits/Optional**
- The `"!load"` prefix is now string-matched in three places (`trace.py:109`, `mapping/base.py:122`, and `factory.py:56` after #1584). A shared constant would remove drift risk, but for a 6-character literal across three call sites that's over-engineering — the new comment pointing at `mapping/uproot.py`/`mapping/base.py` is a better, lighter guard against drift. Leave as-is.
- The expanded comment already mirrors the two-token explanation style; good and consistent.

**Test coverage.** The new `test_form_keys_to_columns_loadallowmissing` is non-vacuous and discriminating (proven by the revert). It constructs **both** a plain `!load` node and a `!loadallowmissing` node (index + content variants, matching the real producer output byte-for-byte) and asserts the **exact** reported set `frozenset({"Muon_pt", "Muon_missing"})` — so it guards against both regressions: dropping allowmissing branches *and* accidentally breaking normal `!load`. It's unit-level against the real quoted-node format rather than an end-to-end trace over a union form; that's a reasonable scope here since the token extraction is the whole logic, but a future e2e trace over an actually-missing branch (the real path exercised together with #1584) would be a nice-to-have to close the last coverage gap.

**Pydantic.** Not applicable — the diff touches no pydantic code.

Posted on behalf of Lindsey Gray.

PR #1592, opened 2026-07-06
fix(nanoevents): backfill genuinely-missing !loadallowmissing branches as None in dask mode

NJManganelli:
**Part of #1578** — completes the dask-mode half of *saved union forms* handling. **Stacked on #1584 — merge that first** (its commit is included here).

Once maybe-missing branches are correctly requested (#1584), `_map_schema_uproot.load_buffers` passed them straight to `tree.arrays`, so reading a file that genuinely lacks such a branch raised `uproot.KeyInFileError` at compute time instead of honoring the `!loadallowmissing` semantics.

This change filters the requested keys to branches actually present in the file (same membership pattern as `mapping/uproot.py`); absent maybe-missing branches then fall through to `PreloadedSourceMapping`'s `allow_missing` path and are backfilled as an all-None `IndexedOptionArray`, exactly matching eager/virtual behavior. Genuinely-required (`!load`) branches that are absent still fail loudly (a `KeyError` from the mapping rather than uproot's `KeyInFileError` — both `KeyError` subclasses).

Adds an end-to-end regression test covering the missing-branch file (all-None, no exception), a mixed two-file dataset (real values + None backfill, with and without `allow_read_errors_with_report`), and the required-branch loud-error path; the has-the-branch test from #1584 continues to pass. Full `tests/test_nanoevents.py` passes; pre-commit clean.

### DIFF
--- a/src/coffea/nanoevents/factory.py
+++ b/src/coffea/nanoevents/factory.py
@@ -167,8 +167,17 @@ class _map_schema_uproot(_map_schema_base):
             f"{start}-{stop}",
         )
         uuidpfn = {partition_key[0]: tree.file.file_path}
+        # A saved/union form can mark branches as maybe-missing
+        # ("!loadallowmissing"); such branches may be genuinely absent from
+        # this particular file, so only request the branches that are present.
+        # Absent branches are then simply not in the preloaded column source,
+        # and the PreloadedSourceMapping backfills them as all-None through its
+        # allow_missing path (matching eager/virtual semantics), while a
+        # genuinely-required ("!load") branch that is absent still raises
+        # loudly at buffer-access time.
+        present_keys = [key for key in keys if key in tree]
         arrays = tree.arrays(
-            keys,
+            present_keys,
             entry_start=start,
             entry_stop=stop,
             ak_add_doc=interp_options["ak_add_doc"],
--- a/tests/test_nanoevents.py
+++ b/tests/test_nanoevents.py
@@ -373,6 +373,41 @@ def test_keys_for_buffer_keys_loadallowmissing():
     assert mapper.keys_for_buffer_keys({index_key, load_key}) == {"flag", "x"}
 
 
+def _preprocessed_form(files):
+    """Preprocess ``files`` ({path: object_path}) and return the saved form."""
+    from coffea.dataset_tools import preprocess
+    from coffea.dataset_tools.filespec import DatasetSpec
+
+    fileset = {"ds": {"files": files}}
+    available, _all = preprocess(fileset, save_form=True, skip_bad_files=False)
+
+    # preprocess preserves the input type: dict in -> dict out.
+    ds_entry = available["ds"] if isinstance(available, dict) else available.root["ds"]
+    ds = (
+        ds_entry
+        if isinstance(ds_entry, DatasetSpec)
+        else DatasetSpec.model_validate(ds_entry)
+    )
+    return ds.form
+
+
+def _write_union_flag_files(tmp_path, n=20):
+    """Write two flat ROOT files where only the first has a bool ``flag`` branch."""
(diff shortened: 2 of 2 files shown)


lgray (commented):

**Summary.** This is the stacked follow-up to #1584. #1584 made dask mode actually request `!loadallowmissing` branches, which introduced a hard `uproot.KeyInFileError` at compute time when such a branch is genuinely absent from a file. #1592 fixes that with a one-line, token-blind filter in `_map_schema_uproot.load_buffers` — `present_keys = [key for key in keys if key in tree]` (`factory.py:178`) — so absent branches never reach `tree.arrays`, fall through to `PreloadedSourceMapping`'s `allow_missing` path, and get backfilled as an all-None `IndexedOptionArray`. I read the mechanism end-to-end and verified the stacked behavior at runtime. **The change is correct and well-tested — no blocking issues.**

**Correctness (verified).** The filter is deliberately blind to the `!load`/`!loadallowmissing` token, but that's safe: the required-vs-allowmissing decision is re-applied downstream in `mapping/base.py:128` (`allow_missing = node == "!loadallowmissing"`) and enforced in `preloaded.py` — an absent *allowmissing* branch returns `None` → all-None array, while an absent *required* (`!load`) branch hits `columnsource[name]` → `KeyError`. So this does **not** silently swallow a truly-required missing branch. I confirmed the required path still raises (`KeyError('flag')`) and the all-None path yields `option[bool]` matching the union form. The all-None construction in `preloaded.py:88-91` (dask) is byte-identical to `mapping/uproot.py:231-234` (eager/virtual), so "matching eager/virtual semantics" holds at the shared mapping layer.

**Test coverage.** Strong. The new `test_union_form_genuinely_missing_branch_dask` covers (a) missing-branch file → all-None + present branches unaffected, (b) mixed two-file dataset → real values + None backfill with real value assertions, and (c) required-branch → loud `KeyError`. Assertions check real values, lengths, and none-counts (not just "no exception"). I proved they are discriminating: reverting only `factory.py` to #1584's head makes test (a) fail with the exact `uproot.KeyInFileError('flag')` at `factory.py:170` that #1584 introduced; restoring makes it pass. The `-25` deletions are a clean refactor of #1584's has-branch test into `_preprocessed_form`/`_write_union_flag_files` helpers — no coverage lost, and #1584's has-branch test (with its `necessary_columns` assertion) is preserved and passing. Full-file run: **22 passed**.

**Pydantic.** The diff adds no pydantic code; the test's `DatasetSpec.model_validate(...)` fallback in `_preprocessed_form` is idiomatic and matches how `preprocess(save_form=True)` output is consumed.

**Nits / optional (non-blocking):**
- The PR frames the check as "the same membership pattern as `mapping/uproot.py`," but that file enumerates branches via `tree.iteritems()` while this uses `key in tree`. Both are correct; the wording just slightly overstates the parallel. A one-line note that `key in tree` is uproot's `HasBranches.__contains__` would age well if anyone later tries to "unify" the two.
- Worth stating in the comment/description that eager/virtual mode ignores `known_base_form` (`factory.py:462` rebuilds the form from the tree), so the "matching eager/virtual" equivalence is a mapping-code guarantee, not something reproducible by calling `from_root(known_base_form=..., mode="eager")` — a future reader who tries the latter will hit `FieldNotFoundError`, not an all-None column.

**Stacked-effect note (measured).** Combined #1584+#1592 gives the intended end state: present-branch files return real values, genuinely-missing `!loadallowmissing` branches return an all-None `option[bool]` column with no exception, and genuinely-required absent branches still fail loudly (`KeyError`). #1584 alone raised `uproot.KeyInFileError` on the missing-branch read; #1592 closes exactly that gap. These two should land in the same release, #1584 first.

Reviewed on behalf of Lindsey Gray.

PR #1593, opened 2026-07-06
fix(tests): stop the shared dask client from leaking as the process default scheduler

NJManganelli:
The session-scoped `dask_client` fixture introduced in #1574 creates its `distributed.Client` with the default `set_as_default=True`, which makes it the **process-wide default scheduler** from the first `dask_client` test onward. In serial (non-xdist) runs, every later `.compute()` in the suite then silently routes through the cluster. Concretely, the FCC nanoevents tests (`test_nanoevents_fcc_*.py`) fail with `TypeError: Could not serialize HighLevelGraph` / `Can't pickle <lambda> on vector.backends.awkward` whenever any `dask_client` test runs earlier in the same process (reproduce on master with `pytest tests/test_lumi_tools.py tests/test_nanoevents_fcc_spring2021.py`). CI masks this because xdist spreads files across worker processes.

This keeps the shared session cluster (now `set_as_default=False`) and routes computations through it via a function-scoped `dask_client` wrapper fixture using `dask.config.set(scheduler=client.get)` plus `client.as_current()`, so the routing is scoped to each `dask_client` test and restored afterwards. All existing `dask_client` consumers (`test_dataset_tools.py`, `test_jetmet_tools.py`, `test_lumi_tools.py`, `test_ml_tools.py`) pass unchanged (25 passed, 4 skipped for missing optional ML deps).

A regression test file exercises both properties deterministically: inside a `dask_client` test, plain `.compute()` resolves to the shared cluster's scheduler; afterwards, the process default scheduler is not any live client's. The previously-failing pairing (`test_lumi_tools.py` + `test_nanoevents_fcc_spring2021.py`) is green with this change.

Discovered while validating the fixes for #1578.


lgray (commented):

## Summary

This is a clean, well-targeted fix. The session-scoped `dask_client` fixture from #1574 created its `distributed.Client` with the library default `set_as_default=True`, which (per `distributed/client.py:1190-1191,1425-1426`) mutates the **global** dask config scheduler and registers the client as the process-global default for the whole session — so in serial runs every later bare `.compute()` silently ships graphs to the cluster. The PR splits this into a session-scoped `_dask_session_client` created with `set_as_default=False` and a **function-scoped** `dask_client` wrapper that routes computations through the cluster only for the duration of each `dask_client` test. Both routing mechanisms are entered as context managers and unwound on teardown, which is exactly the property that makes the fix correct.

**Overall assessment: approve.** I verified the refactor by reading and by running, in an isolated venv (coffea resolved under the worktree; nothing installed into the conda env).

The teardown is the crux, and it's right (`tests/conftest.py:38-40`):
```python
with dask.config.set(scheduler=_dask_session_client.get):
    with _dask_session_client.as_current():
        yield _dask_session_client
```
Both `dask.config.set(...)` and `as_current()` are `with` context managers, so the config override and the current-client stack entry are both restored/popped after each `dask_client` test. Neither is redundant: `dask.config.set(scheduler=client.get)` is what routes a *bare* `.compute()` (via `dask.base.get_scheduler()`), while `as_current()` covers code paths that resolve the client through `get_client()`/`default_client()`. Existing consumers that also call `dask_client.as_current()` in their bodies nest fine.

### Discriminating proof (serial, `-p no:xdist` — xdist masks the bug)

- **End-to-end (2b):** at PR head, `pytest tests/test_lumi_tools.py tests/test_nanoevents_fcc_spring2021.py` → **62 passed**. Reverting only `tests/conftest.py` to master and rerunning the same pairing → **4 FCC tests fail** with `TypeError: ('Could not serialize object of type HighLevelGraph', ...)` → `TypeError: cannot pickle '_DaskProperty' object` (vector's unpicklable dask property). Restoring → 62 passed again. This is the exact leak the PR describes.
- **Unit (2c):** with master conftest, the new `test_default_scheduler_not_leaked` fails (`assert get not in {<bound method Client.get ...>}` — the session client *is* the process default); at PR head it passes.
- Consumers unchanged: `test_dataset_tools.py` 62 passed, `test_lumi_tools.py` passed, new file 2 passed.

## Test coverage

`tests/test_dask_client_fixture.py` is non-vacuous and discriminating (proven by 2b/2c above). It asserts both required properties: (i) inside a `dask_client` test `dask.base.get_scheduler() == dask_client.get` and a real `dask.delayed(...).compute()` executes on the cluster; (ii) a following non-`dask_client` test sees a process default scheduler that is not any session client's `.get`. Relying on definition order to place the "leak-detector" test after a `dask_client` test is a reasonable, deterministic way to exercise the failure vector. Good coverage of the actual bug.

## Nits / Optional

- `tests/test_dask_client_fixture.py:33` — the filter `for client in distributed.Client._instances if client.status` reads as "live clients", but `Client.status` is a `Status` enum string (`"running"`, `"closed"`, …) and any non-empty string is truthy, so closed clients aren't actually excluded. It doesn't weaken the assertion (a larger exclusion set can only make it stricter, and the test still catches the real leak), but `if client.status == "running"` would match the comment's intent.
- Optional: a one-line comment on the wrapper noting *why both* `dask.config.set` and `as_current()` are needed (bare-compute routing vs. `get_client()` resolution) would save a future maintainer the reasoning. The docstrings are already good; the `set_as_default=False` rationale in `_dask_session_client` is especially helpful.

## Pydantic

N/A — tests-only change (`git diff` touches only `tests/conftest.py` and `tests/test_dask_client_fixture.py`; no pydantic usage).

NJManganelli:
Updated. Anything else?

NJManganelli:
Testing itself that the scheduler doesn't leak is a bit meta, but I don't think it hurts to validate this remains true. Anything else needed or can we get this integrated?

PR #1596, opened 2026-07-06
fix: pass strings instead of class objects to copy_behaviors()

### LINKED ISSUE #1594: edm4hep/fcc: copy_behaviors called with class objects instead of strings is a silent no-op

### Summary

While reviewing #1585 (the `Candidate + Candidate` charge fix), I found that the `awkward._util.copy_behaviors(...)` calls in `edm4hep.py` and `fcc.py` are **silent no-ops**: they are called with **class objects** as the `from_name`/`to_name` arguments, but `copy_behaviors` matches purely by **string**. So the intended cross-class behavior copies onto `MomentumCandidate` (and the other EDM4hep/FCC classes) never actually happen.

This is a follow-up to #1585 / part of #1578. It is **latent**, not the charge bug fixed in #1585: `MomentumCandidate + MomentumCandidate` still preserves `charge` (the `@mixin_class` decorator binds that `add` directly and nothing overwrites it). What silently fails is the *cross-class* behavior copying these calls are meant to perform.

### Root cause (proven)

`copy_behaviors` matches names by string equality only ([`awkward/_util.py:108-120`](https://github.com/scikit-hep/awkward/blob/main/src/awkward/_util.py), awkward 2.10.0):

```python
def copy_behaviors(from_name: str, to_name: str, behavior: dict):
    output = {}
    for key, value in behavior.items():
        if isinstance(key, str):
            if key == from_name:
                output[to_name] = value
        else:
            if from_name in key:          # element-equality against a tuple like (numpy.add, "LorentzVector", "TwoVector")
                new_tuple = tuple(to_name if k == from_name else k for k in key)
                output[new_tuple] = value
    return output
```

Behavior-registry keys are strings (e.g. `(numpy.add, "MomentumCandidate", "MomentumCandidate")`, registered via `@awkward.mixin_class_method(numpy.add, {"MomentumCandidate"})` at `edm4hep.py:217`). A **class object** never equals a string, so `from_name in key` is always `False` → the function returns `{}` → the wrapping `behavior.update({})` does nothing.

Self-contained demonstration (awkward only, no coffea):

```python
import numpy
from awkward._util import copy_behaviors

class LorentzVector: ...
class MomentumCandidate: ...
behavior = {(numpy.add, "LorentzVector", "TwoVector"): "the_add_impl"}

copy_behaviors(LorentzVector, MomentumCandidate, behavior)        # -> {}   (class objects: no-op)
copy_behaviors("LorentzVector", "MomentumCandidate", behavior)    # -> {(numpy.add, "MomentumCandidate", "TwoVector"): "the_add_impl"}  (strings: works)
```

The correct pattern is already used in `candidate.py:15`, which passes **strings**: `copy_behaviors("LorentzVector", "Candidate", behavior)`.

### Affected call sites

All pass class objects and are therefore no-ops (line numbers as of master `595647d1`):

**`src/coffea/nanoevents/methods/edm4hep.py`**
- `:206` — `copy_behaviors(base.NanoCollection, edm4hep_nanocollection, behavior)`
- `:260` — `copy_behaviors(vector.LorentzVector, MomentumCandidate, behavior)`  ← motivating case from #1585
- `:324` — `copy_behaviors(MomentumCandidate, MCParticle, behavior)`
- `:438` — `copy_behaviors(MomentumCandidate, ReconstructedParticle, behavior)`

**`src/coffea/nanoevents/methods/fcc.py`**
- `:77` — `copy_behaviors(vector.LorentzVector, MomentumCandidate, behavior)`  ← motivating case from #1585
- `:182` — `copy_behaviors(MomentumCandidate, MCParticle, behavior)`
- `:253` — `copy_behaviors(MomentumCandidate, ReconstructedParticle, behavior)`
- `:449` — `copy_behaviors(MomentumCandidate, MCParticle, behavior)`
- `:504` — `copy_behaviors(MomentumCandidate, ReconstructedParticle, behavior_edm4hep1)`

### Impact

- **No `charge` regression for same-class sums** — verified in #1585 (`MomentumCandidate + MomentumCandidate` keeps `charge`).
- The **cross-class** behaviors these calls intend to propagate (e.g. `MomentumCandidate` + a plain `LorentzVector`/`TwoVector`, and the `MCParticle`/`ReconstructedParticle` copies) are never registered, so such operations silently fall back to awkward defaults. The concrete user-facing consequences per call site have **not** been measured yet and should be assessed as part of the fix.

### Proposed fix

Pass **strings** to `copy_behaviors` (mirroring `candidate.py`), e.g. `copy_behaviors("LorentzVector", "MomentumCandidate", behavior)`, at every site above, and add a regression test that asserts the copied cross-class keys actually appear in the resulting `behavior` dict (a discriminating test: it fails on the current class-object calls and passes once strings are used). This is a mechanical, well-scoped change.

Found during review of #1585; part of #1578.

molloyzak13:
`copy_behaviors()` in awkward matches by string equality only. When you pass class objects (like `vector.LorentzVector` or `MomentumCandidate`), the internal lookup never matches them against the string keys or tuple elements in the behavior registry. The function returns `{}`, and `behavior.update({})` is a no-op.

All 9 `copy_behaviors()` calls in `edm4hep.py` and `fcc.py` were passing class objects, so none of them actually worked. Cross-class behavior propagation (LorentzVector methods onto MomentumCandidate, MCParticle inheriting from MomentumCandidate, the EDM4hep1 FCC copies) was silently broken. The issue has a standalone awkward-only reproducer showing `copy_behaviors(LorentzVector, MomentumCandidate, behavior)` returns `{}` while the string version does the right thing.

The fix converts every call site to pass string class names, matching the pattern already used in `candidate.py:15`. Same 9 call sites, same behavior dicts, just string arguments instead of class objects. No imports changed, no refactors.

Tests pass.

Closes #1594

### DIFF
--- a/src/coffea/nanoevents/methods/edm4hep.py
+++ b/src/coffea/nanoevents/methods/edm4hep.py
@@ -203,7 +203,7 @@ class edm4hep_nanocollection(base.NanoCollection):
 
 
 behavior.update(
-    awkward._util.copy_behaviors(base.NanoCollection, edm4hep_nanocollection, behavior)
+    awkward._util.copy_behaviors("NanoCollection", "edm4hep_nanocollection", behavior)
 )
 
 
@@ -257,7 +257,7 @@ class MomentumCandidate(vector.LorentzVector):
 
 
 behavior.update(
-    awkward._util.copy_behaviors(vector.LorentzVector, MomentumCandidate, behavior)
+    awkward._util.copy_behaviors("LorentzVector", "MomentumCandidate", behavior)
 )
 MomentumCandidateArray.ProjectionClass2D = vector.TwoVectorArray  # noqa: F821
 MomentumCandidateArray.ProjectionClass3D = vector.ThreeVectorArray  # noqa: F821
@@ -321,7 +321,7 @@ class MCParticle(MomentumCandidate, edm4hep_nanocollection):
 
 
 _set_repr_name("MCParticle")
-behavior.update(awkward._util.copy_behaviors(MomentumCandidate, MCParticle, behavior))
+behavior.update(awkward._util.copy_behaviors("MomentumCandidate", "MCParticle", behavior))
 MCParticleArray.ProjectionClass2D = vector.TwoVectorArray  # noqa: F821
 MCParticleArray.ProjectionClass3D = vector.ThreeVectorArray  # noqa: F821
 MCParticleArray.ProjectionClass4D = MCParticleArray  # noqa: F821
@@ -435,7 +435,7 @@ class ReconstructedParticle(MomentumCandidate, edm4hep_nanocollection):
 
 _set_repr_name("ReconstructedParticle")
 behavior.update(
-    awkward._util.copy_behaviors(MomentumCandidate, ReconstructedParticle, behavior)
+    awkward._util.copy_behaviors("MomentumCandidate", "ReconstructedParticle", behavior)
 )
 ReconstructedParticleArray.ProjectionClass2D = vector.TwoVectorArray  # noqa: F821
 ReconstructedParticleArray.ProjectionClass3D = vector.ThreeVectorArray  # noqa: F821
--- a/src/coffea/nanoevents/methods/edm4hep.py
+++ b/src/coffea/nanoevents/methods/edm4hep.py
@@ -321,7 +321,9 @@ class MCParticle(MomentumCandidate, edm4hep_nanocollection):
 
 
 _set_repr_name("MCParticle")
-behavior.update(awkward._util.copy_behaviors("MomentumCandidate", "MCParticle", behavior))
+behavior.update(
+    awkward._util.copy_behaviors("MomentumCandidate", "MCParticle", behavior)
+)
 MCParticleArray.ProjectionClass2D = vector.TwoVectorArray  # noqa: F821
 MCParticleArray.ProjectionClass3D = vector.ThreeVectorArray  # noqa: F821
 MCParticleArray.ProjectionClass4D = MCParticleArray  # noqa: F821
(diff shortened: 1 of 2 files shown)


lgray:

Pushed a follow-up commit (`1704d71b`) to fix the failing FCC tests (`test_KaonParent_to_PionDaughters_Loop` in the edm4hep1/spring2021/winter2023 suites, `AttributeError: no field named 'get_daughters'`).

**Root cause.** Converting the calls to string names was the right fix for #1594 — but it exposed a second, ordering bug. These `copy_behaviors()` calls run *after* each class is defined, and `awkward.mixin_class` registers a class's own record via plain assignment/overwrite (`behaviors/mixins.py:42`), whereas ufunc methods use `setdefault`. While the calls were silent no-ops (class objects), nothing was copied. Now that they actually copy, `behavior.update(copy_behaviors("MomentumCandidate", "MCParticle", ...))` overwrites `behavior["MCParticle"]` (the real `MCParticleRecord`) with the parent's `MomentumCandidateRecord`. Slicing an `MCParticle` array then returns a `MomentumCandidateArray`, which lacks `get_daughters`.

Measured before the follow-up: `edm4hep.behavior["MCParticle"].__name__ == "MomentumCandidateRecord"` (should be `MCParticleRecord`); same for `ReconstructedParticle` and `edm4hep_nanocollection` across all three behavior dicts.

**Fix.** Merge the copied entries with `setdefault` instead of `update`, so genuinely-new cross-class keys are still added while each subclass's own record/array/ufunc registrations are preserved. This mirrors the pattern in `candidate.py`. After the change every record resolves to its own class again, and `copy_behaviors("MomentumCandidate", "MCParticle", behavior)` still returns its 13 keys (so #1594 stays fixed — the copies do real work now).

**Verified** in an isolated env (eager + dask): the 3 previously-failing tests pass, and `test_nanoevents_fcc_*` + `test_nanoevents_edm4hep` + `test_nanoevents_vector` are all green (250 passed) — including the charge-propagation tests, so no regression of the #1585 fix. pre-commit clean.

PR #1605, opened 2026-07-24
docs: add AGENTS.md + CLAUDE.md agent guide with coffea migration section

NJManganelli:
> 🤖 **AI-generated notice:** this branch and pull request were authored by Claude Code (an AI coding agent). Please review accordingly.

## What this adds

Two repository guide files for AI coding agents (and humans) working in coffea:

- **`AGENTS.md`** — a tool-agnostic guide: environment & commands, source layout,
  and current coffea idioms — `NanoEventsFactory` `mode="eager"/"virtual"/"dask"`
  (virtual default), awkward-2 setitem field assignment, scikit-hep `vector` and
  `hist` usage, pydantic `dataset_tools` preprocessing with the dict-in/dict-out
  contract, and the `ProcessorABC`/`Runner`/executor and `apply_to_fileset` paths.
  API details are taken from the current source on `master`.
- **`CLAUDE.md`** — imports `AGENTS.md` via `@AGENTS.md` so shared guidance lives
  in one place, plus a few Claude Code-specific reminders.

## Migration section

`AGENTS.md` includes a dedicated migration guide across coffea's eras:

- **coffea 0.7 (awkward-1 lazy arrays) → latest** — awkward 1→2, lazy arrays →
  explicit `mode=`, `coffea.hist` → `hist`/`mplhep`, and scikit-hep `vector`
  float32→float64 and reserved-coordinate-name gotchas.
- **2023–2025 dask-awkward / `delayed=` → latest** — `delayed=` → `mode=` (no
  shim), dask-free-by-default execution, dask now optional (v2026.7.0),
  `preprocess` returning `(available, all)`, and RNTuple/buffer-cache/`split_fileset`
  additions.
- **Pydantic dataset specification** — intentionally left as a *placeholder*
  advising against pushing users off plain-dict filesets until the staged
  extensions land in a tagged release.

Cross-references: discussions #1529, #705, #1368, and the release notes.

## Notes

- Docs-only change; no code or tests touched.
- The migration target and version floors are written to be updated as new
  releases ship; the guide flags this explicitly.


lgray:
Hold this PR - I want to read it - this kind of stuff needs care, doing brain surgery later is harder than getting it right now.

AGENTS.md
```diff
@@ -0,0 +1,273 @@
...
+
+Coffea 0.7 is the last `awkward`-1 line (Python 3.8/3.9, `coffea.hist`, coffea's
+built-in vector behaviors, lazy `NanoEvents`). It is unmaintained; migrate to the
+CalVer line. The large-scale changes:
+
+1. **awkward 1 → awkward 2.** Field assignment moves to setitem:
+   `events.Electron.myvar = x` → `events["Electron", "myvar"] = x`. Many
+   `ak.*` signatures changed; re-test every array manipulation.
+2. **Lazy arrays → explicit modes.** 0.7's implicit lazy `events` becomes a
+   `mode=` choice. `mode="eager"` is the closest in-memory analogue;
```
NJManganelli: Should be clarified: virtual is the closest analogue to lazy from before, eager is as said the in-memory analogue, ...

AGENTS.md
```diff
@@ -0,0 +1,273 @@
...
+   `mode="virtual"` (the default) reads columns on demand without a dask graph.
+3. **`coffea.hist` → `hist` + `mplhep`** (#705): `Cat(...)` →
+   `axis.StrCategory([], growth=True)`; `Bin(...)` → `axis.Regular/Variable`;
+   sparse categorical axes become dense (watch memory for many categories);
+   `storage="weight"` is opt-in; `weight`/`sample` are reserved axis names;
+   `h.sum("x")` → `h[{"x": sum}]`.
+4. **Vector → scikit-hep `vector`** (#1529): upcast `float32`→`float64` before
```
NJManganelli: It's a bit more nuanced, it's more that scikit-hep/vector keeps to original coordinate systems more stringently which can accidentally blowup, so the "workaround" is upcasting, but it kinda needs testing rather than blind adherence

AGENTS.md
```diff
@@ -0,0 +1,273 @@
...
+   `apply_to_fileset` only for genuine distributed scale-out.
+3. **dask is optional (2026.7).** If you keep the dask path, install it
+   explicitly: `pip install 'coffea[dask,dask-awkward]'`.
+4. **Preprocessing.** `preprocess()` returns `(available, all)` and saves forms
+   by default; `steps_per_file`/`step_size` replace "chunks"; pass
+   `{"allow_read_errors_with_report": True}` in `uproot_options` for access
+   reports. RNTuple inputs are supported (2025.11); buffer caches (2026.4) and
+   `split_fileset`/`Result` (2026.6) are available.
+5. **Schemas & tools** added along the way: EDM4HEP/updated FCC, Scouting,
+   weighted N-1, correctionlib adapter classes for `CorrectedJetsFactory`.
```
NJManganelli: We probably need a general clarification that dask remains as a FIRST-TIER job executor, it's the DAG-functionality we've deprecated

AGENTS.md
```diff
@@ -0,0 +1,273 @@
...
+
+> **Do not yet instruct users to fully migrate off plain-dict filesets.** Several
+> extensions to the pydantic dataset tools — union forms via `DatasetSpec`
+> addition with per-file field bitsets, user metadata extraction during
+> preprocessing, mutable/resizable steps, switchable execution backends
+> (`iterative`/`futures`), RNTuple/Parquet form extraction, and ServiceX /
+> RDataFrame / `universal_pathlib` interop — are staged on **feature branches not
+> yet merged upstream**. Fill in concrete, versioned migration steps here as those
+> branches land in a tagged release, and only then recommend the typed models as
+> the default fileset representation.
```
NJManganelli: This also might need a note to not encourage users to try the ModelFactory, it's more an example now (and a leftover of a very different dataclass-based tooling); all the pydantic stuff can go through direct input conversions, by e.g. wrapping a dict class with appropriate pydantic class 'constructor'

AGENTS.md
```diff
@@ -0,0 +1,273 @@
...
+
+---
+
+## Version targets
+
+- **Latest / recommended:** `v2026.7.0` (July 2026). Track new releases and
```
NJManganelli: I hope this is sufficiently self-referential to update accordingly...

CLAUDE.md
```diff
@@ -0,0 +1,33 @@
...
+- **Read `AGENTS.md` first.** It is the source of truth; the notes below only add
+  Claude-workflow reminders. Do not duplicate its content here — extend
+  `AGENTS.md` instead so both stay in sync.
+- **Before changing behavior**, locate the code with a search, read the relevant
+  module and its mirrored test in `tests/`, then make the smallest change that
+  satisfies the request. Prefer editing existing modules over adding new ones.
+- **Validate locally** the way CI does: `pre-commit run --all-files` and
+  `pytest` (use `pytest -n auto` for speed, or a `-k` selection while iterating).
+  Both must pass before you propose changes.
+- **Match the house comment style**: comments state current behavior and non-obvious
```
NJManganelli: This deserves to be in AGENTS.md, so we keep things sufficiently clean

CLAUDE.md
```diff
@@ -0,0 +1,33 @@
...
+  rationale — no issue/PR numbers, no "used to…/before the fix…" narration.
+- **The dask stack is optional** (v2026.7.0+). Don't assume `dask`/`dask-awkward`
```
NJManganelli: Same: "upstream"?

lgray:
We also need to set security policies so drive-by PRs don't change the agent behaviors by modifying skills.

lgray:
I also want to make sure we implement a proper plan/review/implement/review loop so that we get high quality contributions.

lgray:
@NJManganelli can you get your agent to fab-up a response to that commentary in the skills (it should make individual skills for planning, planning review, implementation, and implementation review) and then I'll take a look?

planning / planning review should repeat until all LOWs or NITs, with a final folding round to get the LOWs and NITs in the plan.

implementation / review should go until clean, if there's an unrecoverable problem in the implementation it should go back to planning/plan-review.

SUPER IMPORTANT -> each of the steps of the loop should be in a *fresh* sub-agent to prevent train-of-thought pollution and keep contexts small and focused.

Just as a sketch.

NJManganelli:
🤖 Posted by NJManganelli's agent (Claude Code). Pushed as `c9731358`.

@lgray @NJManganelli — thanks for the careful review. Understood on holding this; it stays a **draft**, no rush to merge.

### Inline review nits — addressed
- **virtual vs eager vs lazy**: reworded so *virtual* (default) is called out as the closest analogue to 0.7's lazy arrays (materialize-on-access), *eager* as the full in-memory analogue.
- **scikit-hep `vector`**: reframed — it adheres more strictly to the input coordinate system, so `float32` can blow up; upcasting to `float64` is a fix to **verify per-analysis, not follow blindly**.
- **dask**: added an explicit callout that **dask/distributed remain a first-tier job executor**; what's de-emphasized is the `dask-awkward` **DAG** (`mode="dask"`) as the *default* data model.
- **ModelFactory**: now shows direct `DataGroupSpec.model_validate(dict)` conversion and flags `ModelFactory` as an example/leftover, not the conversion API to steer users to.
- **self-referential version section**: strengthened to say a stale target here is a defect to fix in-place.
- **CLAUDE.md cleanup / "upstream"**: folded the tool-specific notes into a tool-agnostic **Agent workflow** section in `AGENTS.md`; `CLAUDE.md` is now just a pointer that `@`-imports it; dropped the ambiguous "upstream" wording.

### Security policy for the agent files (your 2nd comment)
Added a first cut so drive-by PRs can't silently change agent behavior:
- **`.github/CODEOWNERS`** covering `/AGENTS.md`, `/CLAUDE.md`, `/.claude/`, and `CODEOWNERS` itself.
- A **"Provenance & protection"** section in `AGENTS.md`: treat these as a supply-chain surface; recommend branch protection (*Require PR* + *Require review from Code Owners* + *dismiss stale approvals*); **treat PR-branch agent files as untrusted** when an agent reviews an external PR (prompt-injection / staged-attack defense — load operating instructions from the trusted base ref, never the PR head); and optional stronger isolation (keep action-capable skills in a maintainers-only location, plus a CI guard labeling any PR that touches these paths).

**Needs a maintainer decision:** the CODEOWNERS entry uses `@lgray @NJManganelli` as placeholders — please point it at the right maintainer team, and the branch-protection toggles have to be set in repo settings for it to actually bite.

### plan → review → implement → review skills (your sketch)
Happy to build these, but they're behavior-defining, so I'd like to confirm the design before writing them. Proposed:

- **Four skills**: `planning`, `planning-review`, `implementation`, `implementation-review`.
- **Severity taxonomy** to drive the loops — propose `BLOCKER / HIGH / MEDIUM / LOW / NIT` (aligns with the code-review convention). Loops key off "nothing above LOW remains."
- **Plan loop**: `planning` → `planning-review`, repeat until only LOW/NIT remain, then one **folding** round that incorporates the LOW/NITs into the plan.
- **Implement loop**: `implementation` → `implementation-review`, repeat until clean; an **unrecoverable** implementation problem kicks back to `planning`/`planning-review`.

Open questions for you:
1. Should these live **in-repo** under `.claude/skills/` (CODEOWNER-gated per above), or in the **isolated** maintainers-only location we're discussing for action-capable skills?
2. Authoring format — Claude Code **Skills** (`SKILL.md` each), or tool-agnostic markdown so other agents can use them too?
3. Any hard cap on loop iterations before it escalates to a human, and where the plan/review artifacts should be written?

If you confirm those, I'll implement the four skills as a **separate PR** (not folded into this docs PR) so this one stays reviewable.

lgray:
@NJManganelli's agent: The review loops step should live in `.claude/skills` as suggested. Use tool agnostic markdown so multiple agents may use the skills. No hard cap on loop iterations, reviews must go until clean.

ikrommyd:
I think this is too big. It takes a lot of context from smaller models. Typically people do smaller AGENTS.md

lgray:
I prefer high-quality contributions to caring for the risk of blowing lesser models' contexts.

Also - we can ask it to create a more concise set of instructions and skills at the end of this. Let's do this properly first. I would really tend to avoid premature optimization here.

nsmith-:
I agree with Iason.. not all of us want to spend so much tokens on this.
If we're concerned about quality, we have to decide: is this going to become a vibe-coded project or not?
I will **never** review PRs that the author did not themselves review. So if we're going vibe, then yes a very large AGENTS is necessary, and contributions will only be coming from those with the budget to allow it.

lgray:
Then we should set a token budget for skills and agents.md that we all agree to and optimize for that. I don't want to turn away contributions that are vibe coded and also good. I think a reasonable target is that someone with a $20 subscription (or $100 in api tokens) should be able to make a high quality contribution (like a new NanoEvents schema or something). Is that a reasonable price range?

We can also help with hinting which model/effort-level to use so the most cost/token efficient path is chosen.

@NJManganelli @ikrommyd @nsmith-

nsmith-:
One solution is to break things out into separate files. e.g. we can have a ARCHITECTURE.md. The agents should be able to `head ARCHITECTURE.md` (or parse it? idk what's builtin.. one can do `grep -n "^#"` for example) to get a table of contents and read portions of it quite easily to keep context under control. Only AGENTS always gets read in its entirety iiuc

NJManganelli:
That sounds good to me. Yes, it's a very good point that we shouldn't gate things on current models we have access to, even we may lose that as the heavily subsidized era ends.

NJManganelli:
🤖 Posted by NJManganelli's agent (Claude Code). Pushed as `22164729`.

Thanks all — I think @nsmith-'s "factor it out" resolves the size debate, and I've restructured accordingly.

**The thing that actually costs tokens is the *always-loaded* file.** `AGENTS.md`/`CLAUDE.md` get pulled into context every session (and `CLAUDE.md` `@`-imported the whole thing). Everything else is **pay-per-read** — an agent loads it only when the task needs it. So "comprehensive" and "cheap to load" aren't in tension; that's just progressive disclosure. This reconciles the two camps: keep the depth (@lgray's "do it properly"), pay for it only on demand (@ikrommyd / @nsmith-'s context concern).

**What changed (`22164729`):**
- **`AGENTS.md` → a lean ~77-line index** (~550 words): one-liner, environment, the hard rules, a routing table, version targets, and the always-relevant agent-file protection rules. This is all that loads by default now, down from ~380 lines.
- **`ARCHITECTURE.md`** — subsystem map + idioms (NanoEvents modes, awkward/hist, preprocessing, processors), conventions, gotchas, full protection policy. Headings are a TOC: `grep -n '^#' ARCHITECTURE.md`, read one section.
- **`docs/agents/migration.md`** — the 0.7 / dask-DAG / virtual / pydantic guide (the biggest chunk), read only when migrating.
- `CLAUDE.md` imports only the lean index. `CODEOWNERS` + the guard workflow extended to cover the new spokes.

**On scatter vs. index:** I went hub-and-spoke (one lean index + a couple of on-demand docs), *not* per-directory `AGENTS.md` scattered through the tree. Context economy comes from on-demand reads, which the index already gives us; scattering mostly adds maintenance surface and — relevant here — more CODEOWNER-gated paths / injection targets. If a subsystem later proves it needs its own note, we can add one or two then.

**On @lgray's budget target ($20 sub / $100 API to add, e.g., a NanoEvents schema):** I think that's very reachable *because* of this structure — a small always-on index + a task skill that points at just the relevant `ARCHITECTURE.md` section and one worked example, with a model/effort hint. Happy to add a short `docs/agents/budget.md` stating the target so we optimize against it.

**Skills:** per @lgray's answers (in `.claude/skills/`, tool-agnostic markdown, no hard cap, run until clean) I'll build `planning` / `planning-review` / `implementation` / `implementation-review` as a **separate PR** so this one stays reviewable.

Does the lean-index-plus-spokes shape work for everyone before I wire up the skills?

lgray:
@ikrommyd @nsmith- reasonable to you?

AGENTS.md
```diff
@@ -0,0 +1,77 @@
...
+pip install -e '.[dask,dask-awkward]'   # optional: distributed / mode="dask"
+                                        # other extras: parsl, rucio, xrootd, triton
+pre-commit run --all-files              # lint (black, flake8, …) — must pass
+pytest                                  # tests (testpaths=tests/); -n auto to parallelize
+```
+
+## Hard rules (always apply)
+
+- **awkward-2 field assignment is setitem**, not attribute:
+  `events["Muon", "pt2"] = …` — `events.Muon.pt2 = …` is silently dropped.
```
ikrommyd: setattr does not even work for fields in awkward 2

AGENTS.md
```diff
@@ -0,0 +1,77 @@
...
+- **The dask stack is optional** (v2026.7.0+). Don't assume `dask`/`dask-awkward`
+  import; guard dask-only paths with the `coffea/util.py` lazy imports
+  (`_import_dask`, `_import_dask_awkward`).
+- **Comments state current behavior** — no issue/PR numbers, no "used to…/before
```
ikrommyd: I would also like it to have minimal comments only when required in the same way humans write code. LLMs often tend to write full essays. Also a good rule is "if you need a wall of text to explain a change, it's probably not the write change, fix the code"

ARCHITECTURE.md
```diff
@@ -0,0 +1,208 @@
...
+Available schemas: `NanoAODSchema`, `PFNanoAODSchema`, `ScoutingNanoAODSchema`,
+`PHYSLITESchema`, `TreeMakerSchema`, `FCCSchema`, `EDM4HEPSchema`,
+`DelphesSchema`, `PDUNESchema`, and the `BaseSchema` they derive from.
+
+---
+
+## Manipulating awkward arrays
+
+- **Assign fields with setitem, not attribute assignment:**
+  `events["Muon", "pt2"] = events.Muon.pt ** 2` — **not** `events.Muon.pt2 = ...`
```
ikrommyd: repeated from AGENTS.md. Do we need both.

ARCHITECTURE.md
```diff
@@ -0,0 +1,208 @@
...
+  (attribute assignment does not persist on awkward-2 records).
+- Prefer vectorized awkward operations (`ak.num`, `ak.mask`, `ak.combinations`,
+  `ak.firsts`, broadcasting) over Python loops.
```
ikrommyd: I wouldn't even say prefer, I would say for loops over arrays are forbidden unless in a numba jitted function.

ARCHITECTURE.md
```diff
@@ -0,0 +1,208 @@
...
+- Vectors use **scikit-hep `vector`** behaviors. Watch reserved coordinate names
+  (`rho`, `phi`, `eta`, `theta`, `tau`, …): attaching a field named `rho` to a
```
ikrommyd: We also have the `__awkward_validation__` nowadays to enforce that so maybe worth a mention?

AGENTS.md
NJManganelli: Yeah, the issues related to that were always a nuisance

ARCHITECTURE.md
```diff
@@ -0,0 +1,208 @@
...
+
+available, allfiles = preprocess(fileset, step_size=100_000, save_form=True)
+
+# Scale a processor / analysis over the fileset (builds a dask graph):
+out, report = apply_to_fileset(
+    MyProcessor(), max_chunks(available, 5),
+    schemaclass=NanoAODSchema,
+    uproot_options={"allow_read_errors_with_report": True},
+)
+```
```
ikrommyd: It should be very very clear that this is the dask-based way right now. The executors do not use this API at all so any virtual array preprocessing shouldn't know anything about this at the moment until we have a unified API

ikrommyd:
Left a few comments, the smaller AGENTS.md looks reasonable now and for my ignorance, do we need both AGENDS.md and CLAUDE.md ?

NJManganelli:
Claude looks for the latter, so I think we need to treat it as "symlinking" Agents with potential for claude-specific overlay/specification

NJManganelli:
More detailed followup:
My query: 

> Is there anything CLAUDE.md adds substantively (today, or could in the future), when it both has the @AGENTS.md pointer and additional contents? If so it's reasonable to keep additional content there. If the @AGENTS.md fully covers everything Claude might need, we can always keep it trim.

Claude: 

> no, nothing substantive today — and one broken pointer turned up.
> 
> CLAUDE.md's entire non-import content is already in AGENTS.md:
> 
> 
> Since @AGENTS.md pulls all of that in anyway, it's the same duplication ikrommyd flagged at ARCHITECTURE.md:81 — just at file scope.
> 
> Could it earn its keep later? Yes, but only in narrow categories that don't exist yet:
> 1. @-imports — the syntax is Claude-specific. If you ever want Claude only to auto-load a second spoke, that must live here.
> 2. Model/effort hints — lgray's own ask ("hinting which model/effort-level to use so the most cost/token-efficient path is chosen"). Tiers and effort knobs are harness-specific and can't be written tool-agnostically.
> 3. .claude/ mechanics — skills, subagent types, hooks, permissions.
> 4. Symmetry — when GEMINI.md / copilot-instructions.md arrive, each is a thin shim. Keeping this one thin stops Claude-specific guidance from quietly becoming the real source of truth.
> 
> My read: trim now, add back when the content exists. Reserving space costs always-loaded tokens in every Claude session — precisely nsmith-'s and ikrommyd's objection — and what's there today is pure restatement. I kept a 3-line guardrail, which is the one genuinely CLAUDE.md-specific thing (what is allowed in this file):

Which raises a question regarding other agents. It seems that most use AGENTS.md, but github copilot and Gemini may benefit from symlinks. Claude is now an embedded 'symlink' in it's own @ syntax, but could be made a literal symlink...

```

# 1. Gemini CLI / Gemini Code Assist (Expects GEMINI.md at root)
ln -s AGENTS.md GEMINI.md

# 2. GitHub Copilot (Expects .github/copilot-instructions.md)
mkdir -p .github
ln -s ../AGENTS.md .github/copilot-instructions.md
```

NJManganelli:
I'll avoid having Claude add the review skills here, so dropping that line in the router and a separate PR for that will be opened in draft mode. Minimal codeowners added, but @andrzejnovak should you be included too? Anyone else?

ARCHITECTURE.md
```diff
@@ -0,0 +1,298 @@
...
+That trade is what buys the memory. A nominal value, one statistical variance for
+it, and 13 weight-based systematics is `nelem = 15` — 15 doubles per bin. The
+`Weight` equivalent spreads the 14 variations across a categorical systematics
+axis and stores a `WeightedSum` — value *and* variance — in every cell: 2 × 14 =
+28 doubles per nominal bin. The saving comes from paying for the variance once,
+where it is actually needed, instead of once per variation.
+
+When proposing `MultiCell` to an analyst, spell that trade out rather than
+assuming it is understood: they gain one fast fill and roughly half the memory,
+and they give up a statistical uncertainty per variation. Since `nelem` is fixed
```
NJManganelli:
```suggestion
and they can remove statistical uncertainties on variations not requiring them
Since `nelem` is fixed
```

ARCHITECTURE.md
```diff
@@ -0,0 +1,299 @@
...
+## Reading data: `NanoEventsFactory` and execution modes
+
+`from_root` / `from_parquet` take a **`mode`** argument — `"eager"`, `"virtual"`,
+or `"dask"` — that selects the backend:
+
+- **`"eager"`** — arrays are materialized immediately into memory (awkward 2).
+  Simplest mental model; closest to old in-memory workflows.
+- **`"virtual"`** — *the default* for `from_root`/`from_parquet`. Arrays are
+  virtual: columns are read from disk on first access and cached, with no task
+  graph. Memory-efficient and dask-free.
```
NJManganelli:
```suggestion
  graph. Memory-efficient and dask-free, closest to lazy awkward from coffea 0.7
```

ikrommyd:
We can actually just soft link or full copy AGENTS.md to CLAUDE.md. Recommended here https://learn.scientific-python.org/development/guides/ai/

lgray:
@NJManganelli @ikrommyd @nsmith- I gave this a thorough agentic review on Fable 5.1 (asking it to make sure token overhead is minimized, and that it's good for human(-in-the-loop) contributions), and have looked over it. I think it's sized reasonably well for a wide variety of model tiers and coding methodologies (including human only). 

Let me know what you think.

docs/agents/migration.md
```diff
@@ -0,0 +1,95 @@
...
+4. **Vectors go through scikit-hep `vector`.** It keeps to the input coordinate
+   system, so `float32` fields can lose precision or overflow to `inf` where
+   coffea's older behaviors did not; upcasting to `float64` is the usual fix, to be
+   verified against your own kinematics. A custom field that collides with a
+   coordinate its vector already carries is now rejected at construction rather
+   than silently shadowing it (see *Common gotchas* in `ARCHITECTURE.md`).
+5. **Processors and executors** still exist, but several executor arguments are
+   keyword-only and `postprocess` is optional.
+6. **`PackedSelection` and `Weights`** moved from `coffea.processor` to
+   `coffea.analysis_tools`.
```
ikrommyd: they are in `coffea.analysis_tool` in coffea 0.7 too

docs/agents/migration.md
ikrommyd, line 1: There should be a migration from dask-first to virtual arrays too, which is the most difficult one migration.

AGENTS.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+Python >= 3.10, releases are CalVer (`vYYYY.M.P`), dependency floors live in
+`pyproject.toml`. The `0.7.x` line is awkward-1 and unmaintained.
+
+## Hard rules
+
+- **Assign fields with setitem**: `events["Muon", "pt2"] = ...`. Attribute
+  assignment raises `AttributeError` in awkward 2.
+- **No Python loop over an array's `axis=0`** outside a numba-jitted function —
```
ikrommyd: why on axis 0 only?

Edit: I think also vectorization versus writing numba kernels should be preferred unless the logic of vectorization is complicated or it's reasonably slower versus vectorized awkward code.

ikrommyd (commented):
Left a couple of comments from a quick read.

lgray:
@ikrommyd looked over the output - your issues are addressed! Thanks!

AGENTS.md
NJManganelli: Agreed on vectorization over numba. For the axis=0, I think that was it's attempt to distill my instructions: a task that calls vectorized calls over up to 10 candidates along axis=1 is just fine, when it's doing all 100k options along axis=0. Basically just trying to make sure it doesn't do something really stupid like refuse any for loops anywhere, and target the principally bad path.

docs/agents/migration.md
```diff
@@ -37,15 +37,17 @@ behaviors, lazy `NanoEvents`) and is unmaintained.
...
 5. **Processors and executors** still exist, but several executor arguments are
    keyword-only and `postprocess` is optional.
-6. **`PackedSelection` and `Weights`** moved from `coffea.processor` to
-   `coffea.analysis_tools`.
+6. **`PackedSelection` and `Weights`** import from `coffea.analysis_tools`; the
+   `coffea.processor` re-export is gone.
 
 ---
 
 ## B. From the dask-first CalVer releases (2023 to early 2025, `delayed=`)
```
ikrommyd, line 45: Now I"m more confused as to what this migration is meant to cover. Is it early stage dask-awkward to current dask-awkward or dask-awkward to virtual arrays. Cause it feels like a mixture of the two right now.
For the latter, we need to talk about hist.dask -> hist, correctionlib cause old correctionlib versions did not support dask-awkward and we had to use this correctionlib wrapper in coffea, + we need to talk about undoing the things that the user had to do with the `if ak.backend(foo) == "typetracer"` etc to do manual touching when wanting to call non-awkward code. Also largely the ml tools are not needed anymore when moving to virtual arrays cause everything just works on plain eager arrays and dataframes and tensors. It's basically the inverse of https://github.com/scikit-hep/coffea/discussions/775 but modernized + what things were missing from this guide.

lgray: Will give this a another proper look later in the day. Thanks for reading it, I was ~OK with it as a first pass but your concerns are valid.

NJManganelli: I am gonna push back a bit on the axis=0 drop, these LLMs take statements like this too literally/authoritatively and dropping the context might have llms doing something ludicrous just to avoid a 10-iteration walk over axis!=0, so either it should become more explicit about why the general advice is "avoid for loops" so it doesn't do stupid nonsense for users, or go back to that specification

The most important use of ml_tools is not dak wrappers but the triton inference calling IMO (that was the origination of the R&D and Yi-Mu's improvement/upstream of it). Let's not throw out the baby with the bathwater. Maybe in this context that is not relevant, but yeah... 

The migration guide was originally focused on getting users smoothly into calver coffea, how much the dask to virtual migration gets documented is up to others, I see value in that it'll help get people onto virtual. Don't feel strongly on that one though

ikrommyd: what do you mean as a 10-iteration walk over axis!=0? Any other axis iteration gets repeated over all of the events. So it's not really a 10 iteration walk. An axis=1 iteration is iterating over particles. Any iteration should be forbidden in general (unless inside a jit compiled function). The only allowed iterations are over fields.

NJManganelli: If your call is vectorized over the events axis, there's nothing wrong with walking over e.g. 12 mu1 cross mu2 combination candidates. We want to avoid for loops over the events or events cross particles, we do not need to avoid short loops over other axes when vectorized over the large(st) axes, and we have to be explicit with these llms to prevent inflexible pattern matching and garbage downstream.

ikrommyd: Do you have a pattern in mind cause I've never seen such an iteration.

NJManganelli: I've done it numerous times, across combinatorics, across concatenated hlt boolean masks, across particle candidate pairings, etc. I just don't see a valid reason to shock the monkeys climbing the ladder, we have good reason to keep people from bad performance paradigms, a code-shortening loop over reasonable numbers of iterarions (10s to 100s) is not one of them.

lgray: Did this conversation converge to an actual judgement call?

lgray: This whole comment stack appears to be misplaced in the latest text. The text below only talks about the generalities of moving from dask-awkward back to virtual arrays. Performance optimization is a secondary concern.

lgray: As to the original comment from @ikrommyd at the beginning.  A lot of what you're trying to specify here is implication of moving from dask-awkward to virtual arrays, any review loop (especially including humans) will catch this and remove extraneous or non-working code and replace it with the something that works. Since forgetting to do this on the first pass produces a non-working state that fails horrendously, there's no need in my mind to specify these details. It's just wasted tokens to get it on the first try, which isn't necessarily a useful goal. The point is more to make sure the process is convergent, which it is by construction.

NJManganelli: Dead conversation now

lgray:
@ikrommyd @NJManganelli @nsmith- can one of y'all take this (and #1607) for a test drive to check for sick behaviors like validation tailchasing, truncated planning or implementation, etc.?

lgray:
Here's a weekly prod on this topic. Anyone taken #1605/#1607 for a spin?

ikrommyd:
If you mean to use them for development, not me. I'm not that agentic yet.

lgray:
At any level is good - the SKILL contains human-in-the-loop and completely autonomous portions.

PR #1607, opened 2026-07-26
docs: add planning and implementation review-loop skills

NJManganelli:
> 🤖 **AI text follows** 🤖 — this branch and pull request were authored by Claude Code (an AI coding agent) on behalf of @NJManganelli. Please review accordingly.

Implements the review loops @lgray asked for in #1605.

> **Stacked on #1605.** The base is `master`, so until #1605 merges the diff here
> also shows its commits. The commits belonging to *this* PR are the ones after
> `b998995a` — currently a single commit touching 7 files.

## What this adds

Four skills in `.claude/skills/`, plus a `README.md` holding the protocol once so
each skill can stay short:

| skill | reads | writes |
| --- | --- | --- |
| `coffea-planning` | task; latest plan review; `blocked.md` if present | `plan.md` |
| `coffea-planning-review` | `plan.md` | `plan-review-NN.md` |
| `coffea-implementation` | final `plan.md`; latest impl review | code, tests, `impl-notes.md` |
| `coffea-implementation-review` | `plan.md`, `impl-notes.md`, the diff | `impl-review-NN.md` |

Loop: planning ↔ planning-review until only LOW/NIT findings remain, one folding
round to absorb those, then implementation ↔ implementation-review until clean.
An implementation that cannot be completed as planned writes `blocked.md` and
returns to planning.

378 lines total, all read on demand. `AGENTS.md` grows by exactly one routing row.

## Against @lgray's spec

- **Individual skills for each step** — four, separate.
- **Tool-agnostic markdown** — no tool, model, or vendor named anywhere in the
  skills. Any agent that can read files, edit files and run tests can follow them.
- **No hard cap; reviews go until clean** — loops exit on their exit condition,
  never on a count.
- **Planning repeats until all LOW/NIT, then a folding round** — implemented as
  the `CLEAN-ENOUGH` verdict.
- **Unrecoverable implementation problem returns to planning** — via `blocked.md`,
  framed as a normal outcome rather than a failure.
- **Each step in a fresh sub-agent** — see below; this one needed a design
  decision to stay tool-agnostic.

## Design decisions worth arguing with

**Fresh context is enforced by file hand-off, not by instruction.** All loop state
lives in a gitignored `.agent-work/`; a step's only inputs are the repository and
the named files, never a transcript. This is what makes the freshness requirement
portable to tools with no sub-agent primitive, and checkable afterwards — an agent
either read `plan.md` or it did not. It also means a reviewer structurally cannot
inherit the author's justification for the decision it is meant to judge.

**Reviewers may not fix what they find.** Both review skills forbid editing.
Finding and fixing in one pass is how a reviewer talks itself into accepting its
own reasoning.

**Implementation may not improve on the plan.** If the plan is wrong it stops,
rather than quietly doing something better — otherwise "what was reviewed is what
was built" stops being true.

**Non-progress, not iteration count, is the escape hatch.** A round that returns
the same finding unchanged twice is reported to the human instead of retried.

**No human review gate.** The loop ends at a clean implementation review and does
not require the contributor to read the diff before opening a PR. Flagging this
explicitly because @nsmith- objected to exactly this ("I will never review PRs
that the author did not themselves review") — it seemed better to have that
argued here than to have the skill impose an answer.

## Open item: capability tiers

Steps name capability tiers — **C1** strongest reasoning, **C2** strong general,
**C3** fast/cheap — and never a concrete model, so the fastest-rotting part lives
with whichever tool is driving rather than in this repo. Defaults are C1 for
planning and both reviews, C2 for implementation, with a `critical`/`standard`/
`economy` fidelity dial scaling to the blast radius of the change.

**The C1/C2/C3 → model mapping is deliberately not in this PR and is not yet
agreed.** Open questions for @lgray @nsmith- @ikrommyd:

1. Who owns the model↔tier mapping as models churn? This is the version-targets
   problem again and will rot fastest.
2. Is fidelity declared per-PR by the contributor, or derived from the paths the
   diff touches (objective, same shape as `CODEOWNERS`), or both with path as a
   floor?
3. Should the loop report estimated spend? @lgray's "$20 subscription should
   suffice for a new NanoEvents schema" is only checkable if something measures it.
4. Is "escalate a tier after two rejections at the same tier" acceptable, given it
   silently raises cost? The alternative is to fail and hand back to the human.

## Caveats

- **The skills have not been run end-to-end.** They are a first draft of a
  process, not a tested one.
- The MEDIUM/LOW boundary is the kind of thing that only calibrates in use.
- A separate branch (`ci/agent-file-approvals`, not yet a PR) carries a workflow
  requiring two code-owner approvals on agent-governance files; it is unrelated to
  the loop and is kept out of this PR deliberately.


NJManganelli:
> No human review gate. The loop ends at a clean implementation review and does
not require the contributor to read the diff before opening a PR. Flagging this
explicitly because @nsmith- objected to exactly this ("I will never review PRs
that the author did not themselves review") — it seemed better to have that
argued here than to have the skill impose an answer.

Quite wrong phrasing: let's say, no decision is implemented here, and whether there's a hard gate to try to force a human review in the skills is now opened for discussion (I frankly didn't consider this "part of the skills" but now that Claude has proposed it, may as well dicuss). How does one force a human review: is a general understanding sufficient? Must every line be signed off, are block-signoffs sufficient? Does that happen before a PR can even be opened (question: what's a good place to view a potential PR prior to opening it? A raw diff in a text editor is almost certainly a non-starter for really reviewing anything, might be a trivial usage but one that's not coming to me off the top of my head). 

For model capabilities, what's our path forward for "hooking" anybody developing into this flowchart? Specific models will become outdated very rapidly, so we don't want "use Opus 5 for implementation, Fable 5.3 for planning, ..." nor do we want to see that kind of churn. Try to utilize generic benchmarks? As suggestions to the developer's main agent?

lgray:
Together with #1605.

@NJManganelli @ikrommyd @nsmith- I gave this a thorough agentic review on Fable 5.1 (asking it to make sure token overhead is minimized, and that it's good for human(-in-the-loop) contributions), and have looked over it. I think it's sized reasonably well for a wide variety of model tiers and coding methodologies (including human only). 

Let me know what you think.

PR #1609, opened 2026-07-28
fix: ci/triton resilience and eager ml tools tests

NJManganelli:
🧑‍🦱 Human text follows 👨‍🦲 

This tries to apply some fixes for the triton client failures seen in some recent CI runs: mostly intermittent failure to reach the image server, with the remedy being 1) 4 retries spaced exponentially apart (Claude originally tried 15s gaps x4, I prefer this exponentiating one) and 2) splitting tests to run both eagerly and in dask-awkward mode (Claude originally just removed these entirely from non-dask tagged tests... but that's not quite right, no brushing it under the rug, claude-bot)

Some long comments... but this one might be worthwhile keeping

@yimuchen if you want to throw a veto on anything here, you're welcome to. 

🤖  AI text follows 🤖 
Squash-merge collapses the revert pair, so the noise doesn't survive.

Two things worth carrying into the PR description:

- The module-level importorskip fix is the load-bearing one. Old module collected 0 tests without dask_awkward; new one collects 8. That's a coverage hole, not a cosmetic change.
- test_tensorflow_* skips in CI too, not just locally — coffea's create-args list only pytorch-cpu and py-xgboost. The tf wrapper is untested everywhere. Separate issue.


lgray:
@NJManganelli is this PR ready to go?

NJManganelli:
I think so. There was a failure in CI, so maybe the exponential retry is too long and triggers a timeout type death, but let's see if this improves things (or Claude can figure out a better schedule)

PR #1615, opened 2026-08-11
docs: draft docs for processor runner attributes

### LINKED ISSUE #1379: Documentation for `processor.Runner` attributes
Some of the [`processor.Runner`](https://coffea-hep.readthedocs.io/en/v2025.7.3/api/coffea.processor.Runner.html#runner) API is not currently documented with docstrings. It would be great to have short explanations of the various options.

I ran into this via a user asking about what `skipbadfiles` does exactly, looking at the other attributes I suspect also e.g. the behavior of `processor_compression` might not be obvious (though that one is easier to read off from the source code).

tomeichlersmith:
This resolves #1379 , starting small with an less-distributed update to get used to the developing flow before I start on larger rewrites/refactors of the documentation pages.

While I was adding in cross-links I slightly updated how the non-namespace modules are documented. Instead of just dumping them at the bottom of the API Reference page, I have them constructed into their own toctree which a spearate caption.

image

Other QoL updates:
- a few more comments about how to get started contributing to the documentation, mainly mentioning that optional dependencies should be included in developer environment
- include `sphinx-autobuild` so the pages are automatically rebuilt when source files change (its not very fast, but it avoids having to manually re-run `make` every time.)

### DIFF
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1170,6 +1170,18 @@ class Runner:
             determine chunking.  Defaults to a in-memory LRU cache that holds 100k entries
             (about 1MB depending on the length of filenames, etc.)  If you edit an input file
             (please don't) during a session, the session can be restarted to clear the cache.
+        skipbadfiles : bool or tuple of exceptions
+            If True, skip files which throw an exception when opened.
+            If a tuple of exceptions, only skip files that throw those exceptions.
+            If False, 
+        xrootdtimeout : int, optional
+        align_clusters: bool, optional
+        savemetrics: bool, optional
+        schema: schemas.BaseSchema, optional
+        processor_compression: int, optional
+        format: str, optional
+            Defines input file format, must be either 'root' or 'parquet'
+        cachestrategy: "dask-worker" or callable of a mapping, optional
         checkpointer : CheckpointerABC, optional
             A CheckpointerABC instance to manage checkpointing of each chunk output
         use_result_type : bool, optional
--- a/src/coffea/processor/executor.py
+++ b/src/coffea/processor/executor.py
@@ -1145,7 +1145,8 @@ class ParquetFileContext:
 class Runner:
     """A convenience wrapper to submit jobs for a file set
 
-    A file set in this context is a dictionary of dataset: [file list] entries.
+    A file set in this context is a dictionary where the key is the name
+    of a dataset and the value is a list of files.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


src/coffea/processor/executor.py
```diff
@@ -1170,6 +1166,48 @@ class Runner:
...
+            If True, skip files which throw an ``OSError`` exception when opened after reaching
+            the limit on retry attempts.
+            If a tuple of exceptions, only skip files that throw those exceptions.
+            If False, propagate the exception thrown by the file open upwards, probably causing
+            processing to cease.
+            Defaults to False.
+        xrootdtimeout : int, optional
+            Passed to `uproot.open <https://uproot.readthedocs.io/en/latest/uproot.reading.open.html>`_
+            as the ``timeout`` option where it is also undocumented, but I assume is the time in seconds
+            the function will wait before giving up on opening a file.
```
tomeichlersmith: Is this correct?
If so, I might open a PR on the uproot docs as well.

src/coffea/processor/executor.py
```diff
@@ -1170,6 +1167,51 @@ class Runner:
...
+            This is passed as ``compression_level`` to
+            `lz4.frame.compress <https://python-lz4.readthedocs.io/en/stable/lz4.frame.html#lz4.frame.compress>`_
+            for those curious.
+        format: str, optional
+            Defines input file format, must be either ``"root"`` or ``"parquet"``.
+            Defaults to ``"root"``.
+        cachestrategy: "dask-worker" or callable returning a mapping, optional
+            Define the cache used to hold columns of data in memory for re-processing.
+            The default is ``None`` where no caching takes place.
+            If the literal "dask-worker", use the :py:class:`ColumnCache <coffea.processor.dask.ColumnCache>`
```
tomeichlersmith, line 1209: This link is not being generated in the produced documentation. I'm not sure why, but I think its because its not included in the `__all__` of the `processor` module.

tomeichlersmith (commented):
Still some tweaks to do, but happy for feedback at this point as well.

src/coffea/processor/executor.py
tomeichlersmith: I checked the uproot source and the number is indeed in seconds https://github.com/scikit-hep/uproot5/blob/79369aface6543c180fca611a317516fbef7345a/src/uproot/source/xrootd.py#L134

lgray:
@tomeichlersmith this all is starting to look more clean - thanks for the additions and the alignment of the docs with code comments. It's super helpful!

lgray:
@tomeichlersmith open another PR to continue this one and builds should be automatically spawned now, so you can work much more smoothly.

Issue #1281, opened 2025-02-20
Update the `vector` deprecation warning


ikrommyd:
Currently importing coffea things gives
```
/home/iason/Dropbox/work/coffea_dev/coffea/src/coffea/nanoevents/schemas/fcc.py:5: FutureWarning: In version 2025.1.0 (target date: 2024-12-31 11:59:59-06:00), this will be an error.
To raise these warnings as errors (and get stack traces to find out where they're called), run
    import warnings
    warnings.filterwarnings("error", module="coffea.*")
after the first `import coffea` or use `@pytest.mark.filterwarnings("error:::coffea.*")` in pytest.
Issue: coffea.nanoevents.methods.vector will be removed and replaced with scikit-hep vector. Nanoevents schemas internal to coffea will be migrated. Otherwise please consider using that package!.
  from coffea.nanoevents.methods import vector
```
`2025.1.0` has passed so this deprecation warning should be updated to future date (or removed completely).

PR #1622, opened 2026-08-19
feat: add overloads for some awkward reducers for coffea's vectors

### LINKED ISSUE #1620: Certain awkward functions fail on derived classes
**Describe the bug**

Adding objects derived from four-vector classes using `+` works, but using `ak.sum` does not.

**To Reproduce**

Using the [Delphes test file](https://github.com/scikit-hep/coffea/blob/master/tests/samples/delphes.root):
```python
import os, sys
import awkward as ak
import vector as vec
from coffea.nanoevents import DelphesSchema, NanoEventsFactory
from coffea.nanoevents.methods.delphes import behavior

class DelphesSchema2(DelphesSchema):
    def __init__(self, base_form):
        behavior[ak.sum, "Jet"] = vec.backends.awkward.behavior[ak.sum, "Momentum4D"]
        super().__init__(base_form)

myclass = DelphesSchema
if len(sys.argv)>1 and sys.argv[1]=="fix":
    myclass = DelphesSchema2

path = os.path.abspath("delphes.root")
    {path: "Delphes"}, schemaclass=myclass
)
events = factory.events()

test1 = events["Jet"][:,0] + events["Jet"][:,1]
print(test1)

test2 = ak.sum(events["Jet"], axis=-1)
print(test2)
```

**Expected behavior**

Running the example as `python3 test_sum.py fix` succeeds.

**Output**

Running the example without the behavior fix gives:
```
TypeError: no ak.sum overloads for custom types: fUniqueID, fBits, PT, Eta, Phi, T, Mass, DeltaEta, DeltaPhi, Flavor, FlavorAlgo, FlavorPhys, BTag, BTagAlgo, BTagPhys, TauTag, TauWeight, Charge, EhadOverEem, NCharged, NNeutrals, NeutralEnergyFraction, ChargedEnergyFraction, Beta, BetaStar, MeanSqDeltaR, PTD, FracPt_5, Tau_5, SoftDroppedJet, SoftDroppedSubJet1, SoftDroppedSubJet2, TrimmedP4_5, PrunedP4_5, SoftDroppedP4_5, NSubJetsTrimmed, NSubJetsPruned, NSubJetsSoftDropped, ExclYmerge23, ExclYmerge34, ExclYmerge45, ExclYmerge56, Constituents, Particles, Area, pt, eta, phi, mass

    ak.sum(
        <JetArray [[Jet, ...], ...] type='25 * var * jet'>
        axis = -1
    )
```

 - OS: Alma9
 - Version: 2025.12.0

**Additional context**

My main concern is that propagating the fix in the minimal example above to every reducer for every class in every schema is incredibly tedious. Is there a better way to automate this propagation of behavior? (There are many other reducers and classes that will have this problem.)

ikrommyd:
Closes https://github.com/scikit-hep/coffea/issues/1620

Adds overloads for `ak.sum`, `ak.count` and `ak.count_nonzero` on vectors coffea defines. The overloads for count and count nonzero are taken directly from scikit-hep/vector while for sum we use coffea's own `sum` method that its vectors define since we also need to deal with charge which scikit-hep/vector does not deal with.

### DIFF
--- a/src/coffea/nanoevents/methods/vector.py
+++ b/src/coffea/nanoevents/methods/vector.py
@@ -142,6 +142,16 @@ def _coordinate_validation(fields):
     )
 
 
+# awkward looks up reducer overloads by record name only, so the ones vector
+# registers for Momentum{2,3,4}D never reach our collections. Forward those that
+# don't rebuild a record. Sum has to go through our own, which keeps the charge
+# field and names the result after what it actually built.
+_count_reducer = vector.backends.awkward.behavior[awkward.count, "Momentum4D"]
+_count_nonzero_reducer = vector.backends.awkward.behavior[
+    awkward.count_nonzero, "Momentum4D"
+]
+
+
 @awkward.mixin_class(behavior)
 class TwoVector(MomentumAwkward2D):
     """A cartesian 2-dimensional vector
@@ -188,6 +198,18 @@ class TwoVector(MomentumAwkward2D):
             with_name="TwoVector",
             behavior=self.behavior,
         )
+
+    @awkward.mixin_class_method(awkward.sum)
+    def _reduce_sum(self, mask_identity):
+        return self.sum(axis=1)
+
+    @awkward.mixin_class_method(awkward.count)
+    def _reduce_count(self, mask_identity):
+        return _count_reducer(self, mask_identity)
+
+    @awkward.mixin_class_method(awkward.count_nonzero)
+    def _reduce_count_nonzero(self, mask_identity):
+        return _count_nonzero_reducer(self, mask_identity)
 
     @awkward.mixin_class_method(numpy.multiply, {numbers.Number})
     def multiply(self, other):
@@ -303,6 +325,18 @@ class ThreeVector(MomentumAwkward3D):
             with_name="ThreeVector",
             behavior=self.behavior,
         )
+
+    @awkward.mixin_class_method(awkward.sum)
+    def _reduce_sum(self, mask_identity):
+        return self.sum(axis=1)
+
+    @awkward.mixin_class_method(awkward.count)
+    def _reduce_count(self, mask_identity):
+        return _count_reducer(self, mask_identity)
+
+    @awkward.mixin_class_method(awkward.count_nonzero)
+    def _reduce_count_nonzero(self, mask_identity):
+        return _count_nonzero_reducer(self, mask_identity)
 
     @awkward.mixin_class_method(numpy.multiply, {numbers.Number})
     def multiply(self, other):
@@ -423,6 +457,18 @@ class LorentzVector(MomentumAwkward4D):
             with_name="LorentzVector",
             behavior=self.behavior,
         )
+
+    @awkward.mixin_class_method(awkward.sum)
+    def _reduce_sum(self, mask_identity):
+        return self.sum(axis=1)
+
+    @awkward.mixin_class_method(awkward.count)
+    def _reduce_count(self, mask_identity):
+        return _count_reducer(self, mask_identity)
+
+    @awkward.mixin_class_method(awkward.count_nonzero)
+    def _reduce_count_nonzero(self, mask_identity):
+        return _count_nonzero_reducer(self, mask_identity)
 
     @awkward.mixin_class_method(numpy.multiply, {numbers.Number})
     def multiply(self, other):
--- a/tests/test_nanoevents_vector.py
+++ b/tests/test_nanoevents_vector.py
@@ -1225,3 +1225,27 @@ def test_genvistau_addition_propagates_charge():
     common = (ak.num(gvt) >= 1) & (ak.num(mu) >= 1)
     gm = gvt[common][:, 0] + mu[common][:, 0]
     assert "charge" in gm.fields
+
+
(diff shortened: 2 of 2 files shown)


ikrommyd:
cc @kpedro88 you can try this one out.

kpedro88:
This works, thanks!

src/coffea/nanoevents/methods/vector.py
```diff
@@ -189,6 +199,18 @@ def sum(self, axis=-1):
             behavior=self.behavior,
         )
 
+    @awkward.mixin_class_method(awkward.sum)
+    def _reduce_sum(self, mask_identity):
+        return self.sum(axis=1)
```
nsmith-, line 204: We overload this because we want to sum the `charge` field. Should we implement `mask_identity` as an optional argument to `self.sum` (default False) and forward this argument?
> mask_identity ([bool](https://docs.python.org/3/library/functions.html#bool)) – If True, reducing over empty lists results in None (an option type); otherwise, reducing over empty lists results in the operation’s identity.

If we don't want to do so, we should raise an error here if `mask_identity` is true.

nsmith- (changes requested):

src/coffea/nanoevents/methods/vector.py
nsmith-: Actually, I have this wrong: here in `methods/vector.py` we should just forward to the vector impl and delete our own `sum` method. Same for ThreeVector etc.
In https://github.com/scikit-hep/coffea/blob/0c2f9f2a4d84e45277903ed530eb488ae58dd4b3/src/coffea/nanoevents/methods/candidate.py#L17 we should however implement the same forwards for the others and redirect to our own sum for `awkward.sum`

nsmith-: Which reminds me of https://github.com/scikit-hep/coffea/issues/1281 I will make an issue to remind us to figure out the deprecation plan for the `methods/vector.py` wrappers (done: #1624 )

ikrommyd: I do not fully follow. So what do you want to do about the charge fields that coffea candidates carry?

nsmith-: In `Candidate` implement the same forwards as you have here already, and add `mask_identity` handling to our custom `sum()`.
In `*Vector` delete the `sum()` method and forward to vector's implementation.

nsmith-: >  In *Vector delete the sum() method and forward to vector's implementation.

As discussed in the meeting, this does not work because the output will have the input type always, rather than decaying to the appropriate type (e.g. summing `Jet`s should not give back `Jet`, rather `Candidate`)

tests/test_nanoevents_vector.py
```diff
@@ -1225,3 +1225,27 @@ def test_genvistau_addition_propagates_charge():
...
+
+    a = ak.zip(
+        {f: [[1.0, 0.0], [], [2.0]] for f in fields},
+        with_name=name,
+        behavior={"vector": vector, "candidate": candidate}[behavior].behavior,
+    )
+    assert_record_arrays_equal(ak.sum(a, axis=1), a.sum(axis=1))
+    assert ak.to_list(ak.sum(a, axis=1, mask_identity=True))[1] is None
+    assert ak.to_list(ak.count(a, axis=1)) == [2, 0, 1]
+    assert ak.to_list(ak.count_nonzero(a, axis=1)) == [1, 0, 1]
```
nsmith-: Somehow I guess the mask identity argument is already being handled. It's not obvious to me how (maybe because upstream of the record array is the list array and the mask is handled there)
```suggestion
        {f: [[1.0, 0.0], [], [2.0], [1.0, 2.0]] for f in fields},
        with_name=name,
        behavior={"vector": vector, "candidate": candidate}[behavior].behavior,
    )
    expected_sum = ak.zip(
        {f: [1.0, 0.0, 2.0, 3.0] for f in fields},
        with_name=name,
        behavior={"vector": vector, "candidate": candidate}[behavior].behavior,
    )
    assert_record_arrays_equal(ak.sum(a, axis=1), a.sum(axis=1))
    assert_record_arrays_equal(ak.sum(a, axis=1), expected_sum)
    assert ak.to_list(ak.sum(a, axis=1, mask_identity=True))[1] is None
    assert ak.to_list(ak.count(a, axis=1)) == [2, 0, 1, 2]
    assert ak.to_list(ak.count_nonzero(a, axis=1)) == [1, 0, 1, 2]
```

ikrommyd: I think its external code handled in awkward. For jagged arrays, the offsets already tell you if there is an element or not. So I assume awkward sees this and handles this externally, outside of the `_reduce_sum` function, and just masks out the elements. It's here https://github.com/scikit-hep/awkward/blob/6062e76d21b50e2ae83456bb793e00815e76af1e/src/awkward/contents/recordarray.py#L946

nsmith-: That makes sense. Do you want to apply my suggested change? Then I think this is all set to merge

nsmith- (approved):

PR #1625, opened 2026-08-24
feat: allow uv sync to work as expected

nsmith-:
Creates a dev dependency group that references the existing dev extra. A long-term migration to dependency groups for development may be in order but for now this is simple and backwards-compatible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,6 @@ dask-worker-space/
 src/coffea/_version.py
 
 .DS_Store
+
+# Don't track lock files
+uv.lock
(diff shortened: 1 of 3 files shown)


nsmith-:
Related to #1503

ikrommyd:
We should actually drop the whole dev/test optional dependencies and have those only as dependency groups. Optional dependencies are for things you want users to install, not devs.

Overall the coffea project setup is outdated and the CI is not modern enough. I wanted to use some skills Henry has with claude to get it much closer to the sp-repo-review standards and get a more secure CI.

nsmith-:
I was going to say this is superseded by #1628

lgray:
Can fix it over there.

PR #1628, opened 2026-09-02
ci: secure and modernize CI and packaging

ikrommyd:

Applies the `sp-repo-review` / secure-CI pattern recently used in `vector` and `hist`.

**CI security** — all 23 action refs pinned to SHAs with version comments; `permissions: {}` at workflow level with per-job grants documented inline; `persist-credentials: false` on every checkout; matrix expressions moved out of `run:` blocks into env vars. `zizmor --persona=pedantic` goes from 42 findings to none, and is now a pre-commit hook.

**Publishing** — the release job moves out of `ci.yml` into `cd.yml`, triggered on `release: published` instead of a tag push. Build and deploy are separate jobs so nothing that builds ever holds deploy permissions, and `PYPI_TOKEN` is replaced by Trusted Publishing through a `pypi` environment.

> [!IMPORTANT]
> Before the next release: create the PyPI trusted publisher (repo `scikit-hep/coffea`, workflow `cd.yml`, environment `pypi`), create the `pypi` environment in the repo settings, then delete the `PYPI_TOKEN` secret. Releases are now cut by publishing a GitHub Release rather than by pushing a `v*.*.*` tag.

**Cooldowns** — 7 days in all three places that pull dependencies: dependabot (now grouped into one PR), `[tool.uv] exclude-newer`, and pre-commit revs frozen to SHAs.

**Packaging** — closes sp-repo-review GH105, GH212, SEC001, PP006, PP305, PP306, PP307. The `dev` extra becomes `[dependency-groups]` split into `test`/`docs`/`dev`, so `pip install coffea[dev]` is now `--group dev` (CI, RTD and the install docs are updated). pytest gains `--strict-config`, `--strict-markers` and `xfail_strict`. A new `dist` job surfaces packaging breakage on PRs, and the stub `pass` job becomes `re-actors/alls-green` — the job id is unchanged, so the existing branch-protection context still matches.

Bugbear (RF101), a type checker, a task runner and the doc formatters are all left for their own PRs.


ikrommyd:
@henryiii is there any chance you could take a look here? I used `sp-repo-review` and your skills to do this. coffea packaging and ci was very old-fashioned.

lgray:
Hey - handmade CI is a great thing :-p

henryiii:
Mostly looks good, a few comments:

* Note that the branch protection will need to be renamed to `Pass` from `pass`.
* Ignore the PyPI environment comment below, it gets created automatically on publish.
* Ignore the publish comment too, you only trigger a release after CI passes anyway.
* Maybe contributing examples should use uv?


`.github/workflows/ci.yml`
  ● 279 `[correctness]`    Adding `name: All CI jobs passed` to the `pass` job changes its check-run context, but branch protection on master requires the status context `pass`, so the required check is never reported.
  ●  99 `[efficiency]`     The diff creates separate `test` and `docs` groups but every install in ci.yml and .readthedocs.yaml uses `--group dev`, so all matrix cells install the sphinx/myst-nb/jupyter stack and pre-commit although only the ubuntu/3.14 cell builds docs.

`.github/workflows/cd.yml`
  ●  33 `[correctness]`    `publish` needs only `[dist]`; the deleted `release` job needed the test jobs, and ci.yml no longer triggers on `tags: v*.*.*`, so nothing runs tests for the tagged commit before it goes to PyPI and a plain tag push runs no workflow.
  ●  35 `[correctness]`    Publishing switches from `secrets.PYPI_TOKEN` to trusted publishing via `environment: pypi`, but the repo has no `pypi` environment (only `github-pages`) and nothing records that a PyPI trusted publisher for cd.yml / environment pypi must be registered.

`pyproject.toml`
  ● 166 `[correctness]`    `[tool.uv] exclude-newer = "7 days"` is honoured by every `uv pip install` in ci.yml, so CI cannot resolve any release younger than a week and the nightly cron no longer tests fresh upstream releases.
  ● 130 `[correctness]`    `pyinstrument` is in the `dev` group but tests/test_jetmet_tools.py:5 has an unconditional `import pyinstrument`, so the new `test` group cannot collect the suite on its own.
  ● 117 `[simplification]` The new `docs` group carries `nbsphinx` (commented out in docs/source/conf.py:47) and `sphinx-rtd-theme` (html_theme is pydata_sphinx_theme), plus a `sphinx<8` pin from 2024-08 with no recorded reason.

`docs/source/getting_started/installation.md`
  ● 277 `[correctness]`    Documented dev install is now `pip install --editable . --group dev` (also contributing.md:74); `--group` exists only in pip>=25.1 and the `dev` extra was removed, so there is no fallback and the docs do not state the pip requirement.


The review is complete with eight findings reported. The three that block a release or merge: the `pass` job rename breaks the required branch-protection check, `publish` in cd.yml is no longer gated on tests, and no `pypi` environment or trusted publisher exists yet. The `exclude-newer` cooldown also applies to all CI installs and the nightly cron. The rest are dependency-group cleanup: move `pyinstrument` to `test`, use `test`/`docs` groups in CI, document the pip>=25.1 requirement, and drop dead docs entries.

lgray:
just ping me when this one is ready

ikrommyd:
@henryiii - please check again whenever you can.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,72 @@
...
+    needs: [dist]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+    environment:
+      name: pypi
+      url: https://pypi.org/p/coffea
+    permissions:
+      id-token: write # trusted publishing to PyPI
+      attestations: write # record build provenance for the sdist and the wheel
+      contents: read # gh attestation verify reads this repository
```
henryiii: Up to you, but this line is only needed for private repos.

henryiii:
Looks pretty good. Is there a reason you are verifying the attestations right after making them? You just made them so they should verify.

ikrommyd:
> Looks pretty good. Is there a reason you are verifying the attestations right after making them? You just made them so they should verify.

Nope just just what the model generated. I think I've addressed your last two comments as well.

lgray:
It's very likely you can take this out of draft.

ikrommyd:
Yeah I will as soon as I solve the problem with the new edm4hep schemas ci that was added to master. zizmor does not like it at all and is failing right now.

lgray:
I'm sure a way can be figured out around it. It's pretty simple automation that also makes sense as automation.

ikrommyd:
Btw @lgray or someone else, I think you need to do this

> Before the next release: create the PyPI trusted publisher (repo scikit-hep/coffea, workflow cd.yml, environment pypi), create the pypi environment in the repo settings, then delete the PYPI_TOKEN secret. Releases are now cut by publishing a GitHub Release rather than by pushing a v*.*.* tag.

I don't have pypi permissions.

ikrommyd:
@henryiii @lgray should be good now from my side. This needs to be done:

> Note that the branch protection will need to be renamed to Pass from pass.

cause the existing `pass` requirement is just hanging in CI and the pypi update that I said in my message above.

lgray:
OK to go ahead and merge this?

ikrommyd:
From my side yes, I only wonder if Henry has any more comments.

lgray:
@henryiii merging at the end of today (5pm) central time if you make no further comments.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,59 @@
...
+  cancel-in-progress: false
+
+env:
+  FORCE_COLOR: 3
+
+jobs:
+  dist:
+    name: Build sdist and wheel
+    runs-on: ubuntu-latest
+    permissions: {}
```
henryiii: I think this inherits from the above? Does zizmor complain if you remove this line?

henryiii:
BTW, pre-commit.ci will not run the online-only zizmor checks. You have to run it on GitHub (like with the prek action) and GITHUB_TOKEN set to get the online-only checks enabled. Just something I've noticed recently.

.github/workflows/ci.yml
```diff
@@ -34,13 +35,14 @@ jobs:
       with:
         python-version: "3.14"
     - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
```
henryiii: There's a prek action, by the way.

PR #1629, opened 2026-09-03
chore: bokeh 3.10 does not work anymore with the pinned dask/distributed that dask-awkward imposes

ikrommyd:
Hit in the integration tests, when we install dask-awkward, get get an old version of dask/distributed. We should pin bokeh to < 3.10 in that case as well. Also the dask-awkward optional dependencies should be an extension of the dask dependencies. So start with coffea[dask] at the top.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,8 +83,10 @@ dask = [
   "bokeh!=3.0.*,>=2.4.2",
 ]
 dask-awkward = [
+  "coffea[dask]",
   "dask-awkward>=2025.9.0",
   "dask-histogram>=2025.2.0",
+  "bokeh<3.10",
 ]
 parsl = [
   "parsl>=2024.12.09"


nsmith-:
I think we have bokeh as a transitive dependency because its an optional extra for distributed (the dashboards)
Now they don't seem to offer any `distributed[bokeh]` extra, just this https://github.com/dask/distributed/blob/dc182bda54af488d029f2c40db0893d5f48e69d5/distributed/versions.py#L17 and a warning https://github.com/dask/distributed/blob/main/distributed/http/scheduler/missing_bokeh.py#L12

Seems they bumped in https://github.com/dask/distributed/pull/8861

ikrommyd:
Yeah the problem here was that latest bokeh 3.10 is fine with latest distributed but client shutdown crashes with distributed 2025.4.0 which is what we pin at when dask-awkward is in the env. So I had to do this hack.

PR #1631, opened 2026-09-07
fix(jetmet): align Type-1 MET corrections with Analysis Recipes

runtingt:
Update `CorrectedMETFactory` Type-1 mode to reproduce the CMS Analysis Recipes/CMSJMECalculators, including under JERC variations.

Existing call signatures remain compatible but Type-1 numerical results change (intentionally).

## Changes

- Preserve the required float32 correction factors and multiplication order.
- Match datatypes to those used in CMSJMECalculators to reproduce the Analysis Recipes output
- Apply muon subtraction before the Type-1 jet selection.

## Tests

Replace the existing tests with tests targeting:

- Jet-only, `CorrT1METJet`-only, and combined contributions
- exact pT, eta, and EM-fraction selection boundaries for both collections
- muon subtraction and `muonSubtrDeltaPhi`
- nominal JER smearing and JER up/down variations
- Total-JES propagation to both collections
- the nominal-JER offset in JES variations
- unclustered-energy shifts
- agreement between text-based and correctionlib JECs in smeared and unsmeared modes
- a real Summer22 correctionlib Total-JES


runtingt:
In case it's handy, for validation I ran the [`met_type1_syst_case_2`](https://gitlab.cern.ch/cms-analysis/general/analysis_recipes#met_type1_syst_case_2) recipe over `2023preBPix`. This tests the propagation of the Type-1 corrections to `MET_pt` and `MET_phi` under the full JEC split (including the nominal), as well as under the MET `uncusteredEnergy` variations. The tests are against a "truth" sample of 10k ttbar events.

The truth files are under `/eos/cms/store/group/cat/datasets/recipes/Run3-23CSep23-Summer23-NanoAODv12/`, and were generated using the JERC at `/cvmfs/cms-griddata.cern.ch/cat/metadata/JME/Run3-23CSep23-Summer23-NanoAODv12/2026-06-05/jet_jerc.json.gz`. 

I ran the comparison before/after this PR using the attached python file [check.py](https://github.com/user-attachments/files/32020716/check.py). Comparing to the reference file, the nominal + MET variations match, but fail under the jet variations:

| Source revision | Nominal + unclustered | Regrouped JES |
| --- | --- | --- |
| Before: `db25290ba4d7ba44b3d96aa9291575dbc102afb6` | 3/3 pass | 0/22 pass |
| After: `6154ccc67034f4749b08ee1c70d484de38e3c64c` | 3/3 pass | 22/22 pass |

Before the PR, the largest pT difference is 9.8521 GeV  and the largest phi difference is 1.3336 rad. After the PR, the largest pT difference across all variations is 4.2113e-5 GeV and the largest phi difference is 1.6035e-6 rad, which is sufficiently close to pass the analysis recipe check. For completeness I've also attached the complete set of values for all events before/after the PR.

[before.csv](https://github.com/user-attachments/files/32021042/before.csv)
[after.csv](https://github.com/user-attachments/files/32021044/after.csv)

lgray:
@runtingt Thanks for the validation - that basically closes all questions here.

lgray:
@runtingt as a followup here: the tests are essentially the same as check.py? If not could you integrate check.py as a test in coffea so our surface is hardened?

PR #1635, opened 2026-09-10
docs: cleanup readthedocs config

### LINKED ISSUE #1634: docs: wrong path to logo in readthedocs
**Describe the bug**

Image

**To Reproduce**
Observed on https://coffea-hep.readthedocs.io/en/latest/

**Expected behavior**
I'd expect the docs rendered on readthedocs to be identical to the docs rendered with `make -C docs html`.

For funsies (I guess), we have a separate configuration file and build pipeline for readthedocs. 

https://github.com/scikit-hep/coffea/blob/master/.readthedocs.yaml

The quick solution is to just remove the `post_checkout` step which (erroneously) changes the path to the logo, but I want to look in to preventing this in the future by syncing what readthedocs does and what the `docs/Makefile` does.

tomeichlersmith:
This resolves #1634 

- remove comments leftover from template
- bump python version to 3.14
- rework `install` and `build` steps to more closely align with local development (i.e. update the `docs` group and using `make` to run the `sphinx-` commands)
- update `:class:` for the logo so that a white background is not inserted by the CSS


tomeichlersmith:
alright, easy but not-future-proof version works

image

so we can fallback to that commit if I can't find a more elegant solution

tomeichlersmith:
Alright, using `pip` directly instead of `python -m pip` seems to have put the `sphinx-` executables into the PATH so that `make` can be run. Not sure why this is but is probably because RTD is using asdf to shim the specific python version requested inside the requested OS.

There are other dependencies that should be installed in order for the documentation to be complete, see for example
```
* AttributeError: module 'coffea.processor' has no attribute 'dask'
* ModuleNotFoundError: No module named 'zict'
```

I will experiment a bit, but it looks like the change from `'.[dev]'` to `--group docs .` in #1628 removed these. The slop giveth and the slop taketh away.

tomeichlersmith:
My naive attempt using `--extra` did not work, I think I'll look and see if I can reference packages of optional dependencies within dependency-groups or vice-versa so that the `docs` can pull in the optional dependencies without pulling in the test infrastructure. If there isn't an easy way to do that, I'll just update the CI to use `--group dev`.

lgray:
This is starting to look really good and professional, thanks a lot @tomeichlersmith! A minor style gripe, could we get a transparent background on the coffea logo image so that it overlays with the page style and doesn't clash?

tomeichlersmith:
Do you have the original source of the coffea logo? I've done a Typst-source logo before which works well and then can generate multiple formats: https://github.com/LDMX-Software/logo

tomeichlersmith:
Actually, nevermind. I'm pretty sure the coffea logo has a transparent background as well.

image


(The favicon looks the same.) What are you seeing that makes you think the coffea logo does not have a transparent background?

lgray:
image

lgray:
This is how the website looks build by RTD and rendered in chrome.

tomeichlersmith:
Alright, I'm seeing the same thing in firefox with dark mode. I'll look into how sphinx is embedding the image.

tomeichlersmith:
yep, just need to add a class https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/light-dark.html#images-and-content-that-work-in-both-themes

lgray:
Yep - that fixed it - great!

tomeichlersmith (commented):
Alright, I'm ready for this to be merged. The warnings in the readthedocs CI build look the same as the ones I'm seeing locally now.

Issue #1623, opened 2026-08-20
docs: fixup link errors in documentation site


tomeichlersmith:
#1619 uses `nitpicky` to warn about all links that are not successfully finding a destination and then `--warning-file` to store these warnings into a text file. The full file is [warnings.txt](https://github.com/user-attachments/files/31274839/warnings.txt) to download if you're curious.

There are many categories of these warnings that I will enumerate here.

### `stub file not found`
e.g. `autosummary: stub file not found 'coffea.processor.dask.get_worker'.`

I would like to see if we can update the Sphinx configuration to recognize that this object is just our local copy of `distributed.get_worker` and to insert a link to [that object's documentation](https://distributed.dask.org/en/stable/api.html#distributed.get_worker).

### `py:class reference target not found`
e.g. `py:class reference target not found: iterable`

As far as I can tell, these are all present because of a mis-spelling or pluralization. The most complicated is when it references a `namedtuple` that is defined within the function. I think it makes sense to move these `namedtuple`s outside of the function for user typing access anyways.

Some are referencing classes that appear to be spelled correctly but are not appropriately cross linked (e.g. `py:class reference target not found: hist.Hist`), so I can try to add in some more intersphinx_mapping links.

### `'any' reference target not found`
with `ConfigDict`, `kind`, `what`, and `varying_function`.

Edit: The 'any' reference target is used as a place holder when no `:target:` is given before the opening tick. i.e.
```rst
This `module` gets resolved as 'any' reference target.
We should use ``double ticks`` to present in-line code
without asking Sphinx to try to render a link.
```

tomeichlersmith:
Many of the `vector` classes that are in the ancestry tree of `coffea`'s vector methods are not included within the Sphinx object inventory for the `vector` package. I'm trying to decide how to handle this. Especially considering #1624 , having the cross linking be automatic would be very helpful to help guide folks through the deprecation.

PR #1636, opened 2026-09-15
docs: patch some links with updated sphinx config

tomeichlersmith:
I've been testing different approaches to resolving #1623 and I've gotten lost in the piles of warnings. This initial PR is an attempt at the _infrastructural_ changes to Sphinx (i.e. the stuff in `docs/`) and I will come back around to fixing up the docstring syntactically errors in `src/` later.

I'm also mainly opening this PR to get RTD's diff between this PR and the docs from master.


docs/source/api_reference.md
```diff
@@ -22,6 +22,7 @@ imported explicitly.
     :toctree: modules
     :caption: Available by Default
     :template: automodapi_templ.rst
+    :recursive:
```
tomeichlersmith, line 25: Not confident that these `:recursive:` lines are necessary.

docs/source/conf.py
```diff
@@ -101,14 +101,21 @@ def linkcode_resolve(domain, info: dict):
...
+    "rich": ("https://rich.readthedocs.io/en/stable/", None),
+    "pandas": ("https://pandas.pydata.org/docs/", None),
+    "pydantic": ("https://pydantic.dev/docs/validation/latest/", None),
 }
 
 napoleon_preprocess_types = True
 
 autodoc_type_aliases = {
     "awkward.Array": "ak.Array",
+    "Console": "rich.console.Console",
```
tomeichlersmith, line 118: I don't believe this is used anywhere but I should check.

tomeichlersmith (commented):
The part that is frustrating me here is that `automodapi.smart_resolver` is supposed to handle the fact that `coffea_console` is the same between `coffea.util` and `coffea.dataset_tools.dataset_query` but the current configuration just generates two pages, one for each module. It might be because `coffea_console` is a global variable rather than a `class` so the de-duplication is not performed.

PR #1637, opened 2026-09-19
feat(nanoevents): NanoEventsFactory.from_root(mode="graphed")

lgray:

`NanoEventsFactory.from_root(..., mode="graphed")` returns a deferred [graphed](https://github.com/graphed-org/graphed) array, in parallel to `mode="dask"`. graphed is a deferred-execution engine for HEP array analyses; building the analysis reads no event data — field access, behaviors and cross-references are recorded as a graph, and the branches each partition needs are derived from that graph when it runs. Sibling PRs: scikit-hep/uproot5#1720 (the `uproot.graphed` source this opens files through), scikit-hep/fastjet#401, and the merged scikit-hep/hist#721.

**Opened now to start review; not mergeable until uproot releases #1720.** Before merge: (1) delete the CI step "Install uproot from scikit-hep/uproot5#1720" and raise the `uproot` floor to the release that carries it; (2) drop `exclude-newer-package = {graphed = false}` once graphed 0.0.3 (2026-09-19) is older than the seven-day uv cooldown. The CI step installs the PR's head ref from the upstream repository, whose tags give the build a real version (`5.7.7.devN`, so `uproot>=5.7.0` holds).

**How it is wired.**
- `from_root` hands `uproot.graphed` the same `_map_schema_uproot` form mapping the dask arm gives `uproot.dask`, so the schema'd form, the behaviors and the branch-level read list are one implementation for both. `from_parquet` / `from_preloaded` raise `NotImplementedError` for this mode.
- `coffea/nanoevents/_graphed.py` is the only module that imports graphed, and only `mode="graphed"` imports it; importing coffea never does. Without graphed installed the mode names the install (`pip install coffea[graphed]`).
- Every name asked of a NanoEvents collection takes exactly one route: a record field → a recorded `field` op; a `no_dispatch` descriptor → run eagerly on the record-time typetracer; a `dask_property` → its existing `.dask` body, with the graphed array in the deferred array's place; anything else → graphed's own attribute dispatch. **Existing behaviors therefore need no third arm**: there is no `.graphed` slot on `dask_property`, and cross-references come out as ordinary graph edges with both collections as operands, so a worker replays them from the graph alone.
- A schema opts in with `__graphed_capable__ = True` declared in its own class (not inherited, so a subclass that changes the form has to say so itself): `NanoAODSchema` and `PFNanoAODSchema` here.
- Refusals rather than silent wrong answers: `add_systematic` points at `graphed.vary`; `maybe_map_partitions` raises on a graphed array (unguarded it returned `None`); `steps_per_file` / options the source cannot honour are refused at construction.
- A recorded op's provenance — and a run-time error — points at the analyst's line, not a coffea frame.

**Packaging / CI.** A `graphed` extra (`graphed>=0.0.3; python_version>='3.11'`), installed in both test jobs on every OS; its marker keeps it off 3.10, where the three new test files `importorskip`, as they do in the no-dask job. graphed 0.0.3 ships abi3 (cp311+) and cp314t wheels for manylinux/musllinux (x86_64, aarch64), macOS (x86_64, arm64) and win_amd64 — the whole test matrix.

**Tests.** `tests/test_nanoevents_graphed.py` (results against the eager arm on `nano_dy.root`, projection, a pickled plan replayed in a fresh process), `…_guards.py` (both sides of every refusal) and `…_routes.py` (one test per attribute route): 74 tests. The CI install sequence was reproduced in a fresh venv — PyPI graphed 0.0.3, uproot from the PR ref, this branch — and all 74 pass. The user-guide section's example was executed as written.


tests/test_nanoevents_graphed.py
```diff
@@ -0,0 +1,470 @@
...
+        combine=combine_lists,
+        empty=no_lists,
+        backend="coffea.nanoevents._graphed:graphed_backend",
+    )
+    # the count-only output reads no Jet leaf, only the counter that gives it its length
+    assert tuple(sorted(plan.process.columns)) == ("Muon_pt", "nJet", "nMuon")
+    # and the plan ships exactly what the public projection reports for the same outputs
+    projected = set()
+    for output in outputs:
+        projected |= uproot.necessary_columns(output)["Events"]
```
lgray: :robot: _AI text below_ :robot:

`uproot.necessary_columns` is exported neither by uproot main (the #1720 squash 9aa6477) nor by the ref CI installs (`refs/pull/1720/head`, now 11185f5), so this test fails on any fresh CI run.
```
$ python -m pytest tests/test_nanoevents_graphed.py::test_plan_read_list_names_the_counter_branch
E   AttributeError: module 'uproot' has no attribute 'necessary_columns'
tests/test_nanoevents_graphed.py:316: AttributeError
$ git ls-remote https://github.com/scikit-hep/uproot5 refs/pull/1720/head
11185f5e2386c58a12b8fa1a6f18d5c8e71c152e
$ gh api 'repos/scikit-hep/uproot5/contents/src/uproot/__init__.py?ref=11185f5' | base64 -d | grep graphed
from uproot._graphed import graphed, graphed_head
```
Drop the loop, since the literal `("Muon_pt", "nJet", "nMuon")` assertion above already pins the read list. If a cross-check is still wanted, compare against the public driver hook `graphed.write.declared_columns(source, outputs)`, which returns `['Muon_pt', 'nJet', 'nMuon']` here.

src/coffea/nanoevents/_graphed.py
```diff
@@ -0,0 +1,118 @@
...
+        if name in _REFUSED:
+            raise NotImplementedError(_VARY_POINTER)
+        if isinstance(static, _DaskProperty):
+            return static._dask_get(tracer, type(tracer), self)
+        # everything left is graphed's own dispatch, reusing its method/property split; unlike
+        # graphed.Array.__getattr__ it is reached for underscore names too, which the behavior
+        # classes use for their internals
+        if session.backend.attribute_kind(form, name) == "method":
+            return _GraphedBoundMethod(self, name)
+        return session.record_op("field", [self], {"field": name})
```
lgray: :robot: _AI text below_ :robot:

ak.Array's own metadata properties (`fields`, `type`, `ndim`, `layout`, `behavior`, `attrs`, `nbytes`, `typestr`, `mask`, `is_tuple`, `named_axis`, `positional_axis`) fall through to this line and are recorded as `field` ops whose form is a Python list or int. As a result `"pt" in events.Muon.fields` raises, and `events.Muon.ndim == 1` is truthy.
```
$ python p_introspect.py / p_f2_bool.py / p_f2_bare.py
graphed events.Muon.fields          OK   GraphedNanoArray: 'Array(node_id=3)'
graphed 'pt' in events.Muon.fields  FAIL TypeError: argument of type 'GraphedNanoArray' is not iterable
graphed bool(Muon.ndim == 1) -> True
dask    events.Muon.fields          OK   list: "['dxy', 'dxyErr', ..."   ndim -> 2
bare AwkwardBackend array: bare.fields -> Array | form.tt: list | gak.fields(bare): ['x']
```
The root cause is in graphed: `AwkwardBackend.attribute_kind` (python/graphed/awkward/backend.py) treats ak.Array's own descriptors as recordable properties, and `graphed.Array.__getattr__` (python/graphed/array.py) records them. Lines 77-79 re-implement that fallback, so they also need to answer these names from the record-time typetracer. The check is `getattr_static(ak.Array, name, None) is static`; with it, `'pt' in Muon.fields` is True, `ndim` is 2, and the PR suite still has only the F1 failure.

src/coffea/nanoevents/_graphed.py
```diff
@@ -0,0 +1,118 @@
...
+            return static._dask_get(tracer, type(tracer), self)
+        if name in _REFUSED:
+            raise NotImplementedError(_VARY_POINTER)
+        if isinstance(static, _DaskProperty):
+            return static._dask_get(tracer, type(tracer), self)
+        # everything left is graphed's own dispatch, reusing its method/property split; unlike
+        # graphed.Array.__getattr__ it is reached for underscore names too, which the behavior
+        # classes use for their internals
+        if session.backend.attribute_kind(form, name) == "method":
+            return _GraphedBoundMethod(self, name)
```
lgray: :robot: _AI text below_ :robot:

This dispatch accepts single-underscore names so that coffea's `_events` and `_apply_global_index` resolve, but it also accepts ak.Array's own hooks (`_repr_mimebundle_`, `_ipython_key_completions_`, `_repr`, `_histogram_module_`). `_repr_mimebundle_` becomes a `method` op, so every Jupyter display of a collection prints a GraphedTypeError traceback.
```
$ python p_fields.py   # IPython DisplayFormatter().format(events.Muon)
graphed.errors.GraphedTypeError: ill-typed op 'method' at .../IPython/core/formatters.py:1035:
  _repr_mimebundle_(): _repr_mimebundle_() returned dict, which is not an awkward array ...
IPython display data keys: ['text/plain'] | repr: Array(node_id=1)
```
Raise `AttributeError` for underscore names whose static is ak.Array's own attribute (`getattr_static(ak.Array, name, None) is static`); with that gate the display prints no traceback and the routes test stays green. The narrower gate that keeps only `_DaskMethod`/`_DaskProperty` breaks `test_every_name_takes_exactly_one_route`, because coffea's plain `_reduce_count` and `_build_variations` also route through here.

docs/source/user_guide/nanoevents.md
```diff
@@ -93,6 +93,87 @@ df = ak.to_dataframe({"mass": mass, "pt": pt})
...
+    if left is None or right is None:
+        return left if right is None else right
+    return tuple(a + b for a, b in zip(left, right))
+
+
+def empty():
+    return None
+
+
+plan = graphed.aggregate_plan(
```
lgray: :robot: _AI text below_ :robot:

`aggregate_plan` costs about 27 ms per output on NanoAOD, because each output triggers its own `typetracer_with_report` replay over the whole ~1500-leaf mapped form. A processor with a few hundred outputs spends seconds building the plan before it reads any data.
```
$ python p_planprof.py   # nano_dy.root, all outputs distinct, steps_per_file=2
outputs=41  nodes=101 build=1.09s (27 ms/output)
outputs=164 nodes=221 build=4.84s (30 ms/output)
outputs=656 nodes=701 build=18.12s (28 ms/output)
cProfile @164 outputs: 164 x awkward typetracer_with_report = 14.15s of 17.2s
$ python p_perf.py
[dask] reps=1 compute=0.526s   # the whole 41-output analysis, read + compute
```
The root cause is upstream. uproot's `_GraphedTTreeSource.projected_columns` (src/uproot/_graphed.py) calls graphed's single-array `project_buffers` once per output. `_replay` in python/graphed/awkward/projection.py builds a fresh report typetracer for each call. graphed needs a multi-output projection with one replay per source, and uproot should call it once.

pyproject.toml
```diff
@@ -89,6 +89,9 @@ dask-awkward = [
   "dask-histogram>=2025.2.0",
   "bokeh<3.10",
 ]
+graphed = [
+  "graphed>=0.0.3;python_version>='3.11'",
```
lgray: :robot: _AI text below_ :robot:

The floor admits graphed 0.0.3, but with 0.0.3 `from_root(mode="graphed")` fails outright. The uproot this PR opens files through (#1720 at 11185f5, main 9aa6477) gates `uproot.extras.graphed()` on `_GRAPHED_MIN = "0.0.4"`. The tool.uv comment at line 171 also names 0.0.3 as the in-window release, while 0.0.4 (2026-09-22) is the one inside the 7-day cooldown.
```
$ python p_f5.py   # graphed.__version__ set to "0.0.3" before the call
from_root(mode='graphed') with graphed 0.0.3: ModuleNotFoundError uproot.graphed needs graphed 0.0.4 or newer; you have graphed 0.0.3
control, graphed 0.0.4 : OK GraphedNanoArray
$ gh api 'repos/scikit-hep/uproot5/contents/src/uproot/extras.py?ref=11185f5' | base64 -d | grep _GRAPHED_MIN
_GRAPHED_MIN = "0.0.4"
```
Raise the extra to `graphed>=0.0.4;python_version>='3.11'`, and make the comment name 0.0.4 (2026-09-22) as the release the bypass exists for.

src/coffea/nanoevents/_graphed.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def check_from_root(schemaclass, steps_per_file, uproot_options):
+    """Refuse the ``from_root`` arguments and schemas graphed mode cannot honour."""
+    if steps_per_file is not uproot._util.unset:
+        raise NotImplementedError(
+            "graphed mode does not partition files here; pass steps_per_file where the plan is "
+            "built, graphed.aggregate_plan(..., steps_per_file=...)"
+        )
+    if "allow_read_errors_with_report" in uproot_options:
```
lgray: :robot: _AI text below_ :robot:

The key test refuses the no-op `allow_read_errors_with_report=False`, which the dask arm accepts, and the error message talks about a report tuple the caller did not ask for.
```
$ python p_f7.py
dask False: OK Array
graphed False: NotImplementedError graphed mode has no report tuple; failed partitions are dead-lettered ...
# with only the value tested in check_from_root:
TypeError uproot.graphed() got allow_read_errors_with_report=, which belongs to uproot.dask; ...
```
Refuse only a truthy value, and drop a falsy key before forwarding `**uproot_options`, because uproot's `_DASK_ONLY` guard (src/uproot/_graphed.py) also refuses the key by presence.

src/coffea/nanoevents/factory.py
```diff
@@ -253,7 +271,18 @@ def __call__(self, form):
...
+
+def _tree_to_open(file, treepath):
+    """The tree the deferred arms hand to uproot, resolving a directory through ``treepath``."""
+    if not isinstance(file, uproot.reading.ReadOnlyDirectory):
+        return file
+    if treepath is uproot._util.unset:
+        raise ValueError(
+            "The treepath argument must be specified when the file argument is an uproot.reading.ReadOnlyDirectory"
+        )
+    return file[treepath]
```
lgray: :robot: _AI text below_ :robot:

The eager arm further down still has its own copy of this block (same `isinstance`, same message, `tree = file[treepath]`), so the two error paths can drift apart.
```
$ sed -n 481,487p src/coffea/nanoevents/factory.py
        if isinstance(file, uproot.reading.ReadOnlyDirectory):
            if treepath is uproot._util.unset:
                raise ValueError(
                    "The treepath argument must be specified when the file argument is an uproot.reading.ReadOnlyDirectory"
                )
            tree = file[treepath]
            file_handle = file
```
Replace it with `tree = _tree_to_open(file, treepath); file_handle = file` inside the same `isinstance` branch.

lgray (commented):

Deep adversarial review of #1637 at head `9c89d7b1` (base `master` `c917005`), run against the upstream mains now that uproot5#1720 is merged. Two passes (finder, adjudicator, completeness pass, adjudicator); only findings the adjudicator reproduced are posted, each inline with its probe.

**Environment.** CPython 3.13, every package editable from source: awkward main `f9e63be` (2.14.0, awkward-cpp 57), uproot5 main `a4b58cb` (unreleased: 10 commits past v5.7.6, including the #1720 squash `9aa6477`), hist `52303af`, vector `e036823`, correctionlib `1c0d734`, graphed main `234ceab` (0.0.4), graphed-executors `84f5920`, graphed-histogram `4beee73`; dask 2025.3.1, dask-awkward 2026.2.1. Samples: `tests/samples/nano_dy.root`, `pfnano.root`.

**Baseline.** `pytest tests/test_nanoevents_graphed.py tests/test_nanoevents_graphed_guards.py tests/test_nanoevents_graphed_routes.py`: 73 passed, 1 failed. The failure is `test_plan_read_list_names_the_counter_branch`: `uproot.necessary_columns` is no longer exported by merged uproot (it moved to the private `uproot._graphed`). CI is still green only because it installs an older `refs/pull/1720/head`; a fresh run turns red.

**Confirmed: 1 high, 3 medium, 9 low** (inline).
- high: the test above calls an uproot name that no longer exists.
- medium: ak.Array metadata properties (`fields`, `ndim`, `type`, …) are recorded as ops, so `"pt" in events.Muon.fields` raises and `events.Muon.ndim == 1` is truthy. Rooted in graphed; fix open as graphed-org/graphed#42 (the coffea `__getattr__` fallback needs the matching arm).
- medium: `_repr_mimebundle_` is dispatched as a method, so every Jupyter display of a collection prints a GraphedTypeError traceback (coffea-only: bare graphed arrays refuse `_` names).
- medium: plan build costs ~27 ms per output (one `typetracer_with_report` replay per output). Rooted in graphed + uproot's call site; graphed side open as graphed-org/graphed#43 (`project_buffers_many`), uproot's `projected_columns` to switch to it once that is released.
- low: graphed floor must be 0.0.4 (uproot's `_GRAPHED_MIN` rejects 0.0.3); `allow_read_errors_with_report=False` is refused; the eager arm duplicates `_tree_to_open`; the routes test duplicates the pair fixtures; `graphed_backend` repeats `aggregate_plan`'s default; the guide never shows how to get the analyst-line `StageError` (plain plan runs re-raise the bare awkward error by design in graphed; `graphed.debug.run` attributes); a `__main__` schema subclass and the guide's own `reduce`/`combine`/`empty` cannot be loaded on a worker; the missing-graphed message suggests an install that does nothing on Python 3.10; a graphed-mode factory cannot be pickled.

**Refuted (not posted).** `entry_start`/`entry_stop` being ignored in graphed mode is pre-existing and documented as eager/virtual only, and dask mode ignores it the same way. Plan pickle bytes varying with the hash seed come from pre-existing 2023 schema code, identical in eager mode; the compiled IR bytes and the read list are seed-stable.

**Checked with no findings.**
- The docs example, run as written, prints the documented `['Jet_pt', 'Muon_jetIdx', 'nJet', 'nMuon']` and `[5, 8, 5, 3, 5]`.
- Graphed vs eager parity over 40 analyses on nano_dy.root and pfnano.root: 37 identical, 3 fail identically in both modes.
- Plan pickles are byte-identical across two factories; a plan run twice gives identical results; `materialize` matches the plan result; a pickled plan (with the `matched_jet` cross-reference) replays identically in a fresh interpreter.
- Input shapes: a `path:tree` string, a list of two files, and a dict with explicit `steps` give 1, 2 and 2 tasks with the correct row counts.
- Refusals: steps_per_file, the report option, systematics, callables, from_parquet and maybe_map_partitions all raise with a pointer to the alternative.
- Repeated `events()` calls leave no `GraphedNanoArray` or session alive: 0.0 MB net growth over 10 calls (dask mode: 8.5 MB).
- Recording cost is linear in the number of field accesses (graphed 0.69 ms per access vs dask 0.40).

tests/test_nanoevents_graphed_routes.py
```diff
@@ -0,0 +1,216 @@
...
+
+@pytest.fixture(scope="module")
+def nanoaod(tests_directory):
+    return _arms(tests_directory, "nano_dy.root", NanoAODSchema)
+
+
+@pytest.fixture(scope="module")
+def pfnano(tests_directory):
+    return _arms(tests_directory, "pfnano.root", PFNanoAODSchema)
+
```
lgray: :robot: _AI text below_ :robot:

`_arms` and the `nanoaod`/`pfnano` fixtures here re-implement `_both_arms`, `nanoaod` and `pfnano` from `test_nanoevents_graphed.py`, which this module already imports `assert_same` from.
```
$ sed -n 172,188p tests/test_nanoevents_graphed.py   # _both_arms -> (graphed events, eager events); nanoaod/pfnano fixtures
$ sed -n 90,110p tests/test_nanoevents_graphed_routes.py  # _arms: same loop over ("graphed", "eager"); same two fixtures
$ sed -n 18p tests/test_nanoevents_graphed_routes.py
from test_nanoevents_graphed import assert_same  # noqa: E402
```
Import `nanoaod` and `pfnano` (or `_both_arms`) from `test_nanoevents_graphed` next to `assert_same` and delete the copy.

src/coffea/nanoevents/_graphed.py
```diff
@@ -0,0 +1,118 @@
...
+
+def graphed_backend():
+    """The worker-side recording backend, addressed as an import reference::
+
+        graphed.aggregate_plan(..., backend="coffea.nanoevents._graphed:graphed_backend")
+
+    It carries no behavior dict: a mapped chunk is assembled by ``awkward.from_buffers`` with the
+    schema's behavior already applied, and that dict holds closures that would not pickle.
+    """
+    return GraphedNanoBackend()
```
lgray: :robot: _AI text below_ :robot:

`graphed_backend` repeats what `graphed.aggregate_plan` already does when `backend=` is omitted. The default puts `type(session.backend)`, i.e. `GraphedNanoBackend`, in the plan, and a worker constructs it with no behaviors, which is the same object this function returns.
```
$ python r2_backend_default.py planA.pkl && python r2_fresh_run.py planA.pkl
A. backend_factory without backend=: <class 'coffea.nanoevents._graphed.GraphedNanoBackend'>
A. backend_factory with docs string: coffea.nanoevents._graphed:graphed_backend
A. resolved backend types: GraphedNanoBackend GraphedNanoBackend | behavior: None None
A. in-process run without backend=: [5, 8, 5, 3, 5]
fresh-interpreter result[1][:5]: [5, 8, 5, 3, 5]
# graphed/aggregate.py aggregate_plan:
#   backend_factory=backend if backend is not None else type(session.backend)
```
Delete `graphed_backend`, `test_worker_backend_is_importable_by_reference` and the `backend=` arguments in the tests and the guide example, and reword the guide's "Workers build coffea's recording backend from an import reference" sentence to match.

docs/source/user_guide/nanoevents.md
```diff
@@ -93,6 +93,87 @@ df = ak.to_dataframe({"mass": mass, "pt": pt})
...
+    reduce=reduce,
+    combine=combine,
+    empty=empty,
+    steps_per_file=2,
+    backend="coffea.nanoevents._graphed:graphed_backend",
+)
+print(sorted(plan.process.columns))
+# ['Jet_pt', 'Muon_jetIdx', 'nJet', 'nMuon']
+
+matched, n_jet_per_event = SequentialRunner().run(plan).value
```
lgray: :robot: _AI text below_ :robot:

The PR description says a run-time error points at the analyst's line, but the runner this guide teaches re-raises awkward's bare exception. For a plan with no variations, graphed's `_PartitionReduce._attribute` installs no handler, so only `graphed.debug.run` produces the attributed `StageError`. The guide never mentions it, and `import graphed` does not import `graphed.debug`.
```
# ev.Muon.pt[:, 0] recorded at line 8; SequentialRunner().run(plan) at line 26
$ python r2_runtime_error.py
IndexError: cannot slice ListArray (of length 40) with array(0): index out of range ...
traceback frames in the analysis file: ['... r2_runtime_error.py", line 26, in <module>']
# the same op run with `import graphed.debug; graphed.debug.run(ev.session, first)`
$ python r2_debug_run2.py
StageError in op 'subscript' at .../r2_debug_run2.py:4 (partition sample, opt_level=1): IndexError: ...
picklable: StageError
```
After the example, add a sentence pointing at `graphed.debug.run(events.session, out)` for errors mapped to the analysis line, or drop the run-time-error claim from the PR description.

docs/source/user_guide/nanoevents.md
```diff
@@ -93,6 +93,87 @@ df = ak.to_dataframe({"mass": mass, "pt": pt})
...
+| --- | --- |
+| `from_root(..., steps_per_file=...)` | `graphed.aggregate_plan(..., steps_per_file=...)` |
+| `uproot_options={"allow_read_errors_with_report": True}` | `graphed.checkpoint.run_resumable` |
+| `add_systematic`, `_ensure_systematics` | `graphed.vary(...)` |
+| a callable argument, e.g. `metric_table(..., metric=fn)` | `graphed.vary(...)` for a variation; otherwise record the metric with array operations |
+| `from_parquet(..., mode="graphed")` | `from_root(..., mode="graphed")` |
+| `coffea.util.maybe_map_partitions` | record the operation on the graph |
+
+A schema is admitted only if it declares `__graphed_capable__ = True` in its own class body, which
+`NanoAODSchema` and `PFNanoAODSchema` do. Subclasses do not inherit the flag: declare it once the
```
lgray: :robot: _AI text below_ :robot:

This sentence invites users to declare a schema subclass, but one declared in a notebook or script runs in-process and then cannot be loaded by any worker process. `_map_schema_uproot.__reduce__` ships the class by reference, and graphed plans travel by plain `pickle`. The dask arm survives the same subclass because cloudpickle ships `__main__` classes by value. The example's own `reduce`/`combine`/`empty`, defined in `__main__`, fail the same way.
```
$ python r2_main_schema.py planH.pkl mapH.pkl   # class MyNanoAODSchema(NanoAODSchema) in __main__
H. __main__ schema subclass: plan pickles in-process, bytes 277943 | in-process run: [[], []]
$ python r2_fresh_load.py planH.pkl pickle mapH.pkl cloudpickle
fresh load pickle: AttributeError: Can't get attribute 'MyNanoAODSchema' on <module '__main__' ...>
fresh load cloudpickle: OK _map_schema_uproot <class '__main__.MyNanoAODSchema'>
$ python r3_docs_main.py planD3.pkl && python r2_fresh_load.py planD3.pkl pickle   # guide's reduce/combine/empty
fresh load pickle: AttributeError: Can't get attribute 'reduce' on <module '__main__' ...>
```
Say once in this section that a plan's callables and any custom schema class must live in a module the workers can import.

src/coffea/util.py
```diff
@@ -133,6 +135,21 @@ def _import_dask_awkward():
     return dask_awkward
 
 
+def _import_graphed():
+    try:
+        import graphed
+    except ModuleNotFoundError as err:
+        raise ModuleNotFoundError("""to use this feature, you must install graphed:
+
+pip install coffea[graphed]
```
lgray: :robot: _AI text below_ :robot:

On Python 3.10, which coffea supports and CI runs, the extra's marker makes `pip install coffea[graphed]` install nothing. This message therefore sends the user to a no-op, and `mode="graphed"` keeps raising the same error.
```
$ for v in 3.10 3.11; do uv pip compile --python-version $v --python-platform linux \
    --extra graphed pyproject.toml | grep '^graphed='; done
py3.10: (no graphed line)
py3.11: graphed==0.0.4
$ grep requires-python pyproject.toml
requires-python = ">=3.10"
```
Name the Python 3.11 floor in the message, as the guide does, or raise a version-specific error before the import on 3.10.

src/coffea/nanoevents/factory.py
```diff
@@ -377,18 +408,40 @@ def from_root(
...
+            opener = partial(
+                uproot.graphed,
+                _tree_to_open(file, treepath),
+                full_paths=True,
+                ak_add_doc={"__doc__": "title", "typename": "typename"},
+                filter_branch=_is_interpretable,
+                known_base_form=known_base_form,
+                # the recording backend carries the behavior because the typetracer the deferred
+                # routes dispatch on is built from it
+                backend=_graphed.GraphedNanoBackend(behavior=behavior),
```
lgray: :robot: _AI text below_ :robot:

Binding the behavior-carrying `GraphedNanoBackend` into the opener partial makes a graphed-mode `NanoEventsFactory` unpicklable, while eager and virtual factories round-trip through `__getstate__` (`test_factory_pickle_preserves_mode`).
```
$ python r3_factory_pickle.py
M. eager: pickle round-trip OK, events() -> NanoEventsArray
M. virtual: pickle round-trip OK, events() -> NanoEventsArray
M. graphed: PicklingError: Can't pickle <function <lambda> ...>: attribute lookup <lambda> on vector.backends.awkward failed
# same factory with backend= dropped from the partial, passed at events() time from self._schema.behavior:
M. backend moved to events(): round-trip OK; GraphedNanoArray True
```
Build the backend in `events()` from `self._schema.behavior` (the mapping already rebuilds its behavior on unpickle) instead of binding it in the partial, and add "graphed" to `test_factory_pickle_preserves_mode`.

tests/test_nanoevents_graphed.py
lgray: :robot: _AI text below_ :robot:

Dropped the loop in 44e322ba; the literal `("Muon_pt", "nJet", "nMuon")` assertion still pins the read list.

src/coffea/nanoevents/_graphed.py
lgray: :robot: _AI text below_ :robot:

01b7d4b0 adds the matching arm to `GraphedNanoArray.__getattr__`: `fields`, `type`, `typestr`, `ndim`, `is_tuple` and `positional_axis` are answered from the record-time typetracer and the other ak.Array properties raise `AttributeError`, the same split as graphed-org/graphed#42, which fixes the root in graphed.

src/coffea/nanoevents/_graphed.py
lgray: :robot: _AI text below_ :robot:

01b7d4b0 raises `AttributeError` for underscore names whose static is ak.Array's own attribute; a new test drives IPython's `DisplayFormatter`, and `test_every_name_takes_exactly_one_route` stays green.

docs/source/user_guide/nanoevents.md
lgray: :robot: _AI text below_ :robot:

No coffea change can fix this: graphed-org/graphed#43 (`project_buffers_many`) does one replay per source, and uproot's `projected_columns` switches to it once a graphed release carries it.

pyproject.toml
lgray: :robot: _AI text below_ :robot:

949fc67f raises the extra to `graphed>=0.0.4;python_version>='3.11'` and the tool.uv comment now names 0.0.4 (2026-09-22); no lock file is tracked.

src/coffea/nanoevents/_graphed.py
lgray: :robot: _AI text below_ :robot:

3e44888a refuses only a truthy `allow_read_errors_with_report` and drops the key before forwarding to `uproot.graphed`; a new test covers `False`.

src/coffea/nanoevents/factory.py
lgray: :robot: _AI text below_ :robot:

ab82ca88 replaces the eager arm's copy with `tree = _tree_to_open(file, treepath)`.

tests/test_nanoevents_graphed_routes.py
lgray: :robot: _AI text below_ :robot:

be770bc5 imports `_both_arms` and deletes `_arms`; importing the fixtures themselves trips ruff F401/F811, so the two one-line fixtures stay.

src/coffea/nanoevents/_graphed.py
lgray: :robot: _AI text below_ :robot:

d39e3649 deletes `graphed_backend`, its test and every `backend=` argument; the guide now says the plan names coffea's recording backend by its class.

docs/source/user_guide/nanoevents.md
lgray: :robot: _AI text below_ :robot:

0ea7346b adds a sentence after the example pointing at `graphed.debug.run(events.session, output)` for a `StageError` naming the analysis line.

docs/source/user_guide/nanoevents.md
lgray: :robot: _AI text below_ :robot:

0ea7346b adds one sentence to this section: a plan's `reduce`/`combine`/`empty` and any custom schema class must live in a module the workers can import.

src/coffea/util.py
lgray: :robot: _AI text below_ :robot:

68600755 names the Python 3.11 floor in the message.

src/coffea/nanoevents/factory.py
lgray: :robot: _AI text below_ :robot:

16818df7 builds `GraphedNanoBackend` in `events()` from `self._schema.behavior` instead of binding it into the opener, and `test_factory_pickle_preserves_mode` now covers "graphed".

docs/source/user_guide/nanoevents.md
lgray: :robot: _AI text below_ :robot:

Closed upstream: scikit-hep/uproot5#1726 (merged 2026-09-24, `ca3a8a2`) projects every output's buffers in one form replay via graphed's `project_buffers_many`, and CI now installs uproot main at that commit (b2612ab). Same probe, same file, graphed 0.0.5:

```
$ python p_planprof.py   # nano_dy.root, all outputs distinct, steps_per_file=2
outputs=41  nodes=101 build=0.06s (1 ms/output)     # was 1.09s (27 ms/output)
outputs=164 nodes=221 build=0.16s (1 ms/output)     # was 4.84s (30 ms/output)
outputs=656 nodes=701 build=0.53s (1 ms/output)     # was 18.12s (28 ms/output)
```

lgray:

scikit-hep/uproot5#1720, #1726 and #1727 are all merged on uproot main, so CI now installs uproot from main at `ca3a8a2` (the #1726 merge commit) instead of the #1720 PR ref (b2612ab). Nothing else in this PR changes:

- #1726 (one form replay for all outputs) removes the plan-build cost noted in the docs thread: 27–30 ms/output → 1 ms/output on the same probe (0.53 s for 656 outputs, was 18.1 s).
- #1727 (basket-aligned partitions) is opt-in through `partitions=uproot._graphed.graphed_partitions(..., align_baskets=True)` where the plan is built; `from_root(mode="graphed")` does not partition, so no coffea change.
- The CI install sequence reproduced in a fresh venv (PyPI graphed 0.0.5, uproot main, this branch): the 77 graphed tests pass.

Still pending before merge: an uproot release carrying the three, then the CI step goes and the `uproot` floor rises; the uv `exclude-newer` exception for graphed can go on 2026-09-29.

